PR #48, opened 2015-07-31
Add support for forbidden C calls

tomato42:
add a generic mechanism to blacklist certain C calls with ability to
automatically waive the warning if a known good value is defined
in executable.

### DIFF
--- a/BinariesCheck.py
+++ b/BinariesCheck.py
@@ -31,6 +31,12 @@ def create_regexp_call(call):
     r = "\s+FUNC\s+.*?\s+(%s(?:@GLIBC\S+)?)(?:\s|$)" % call
     return re.compile(r)
 
+def create_nonlibc_regexp_call(call):
+    if type(call) == type([]):
+        call = '(?:' + '|'.join(call) + ')'
+    r = "\s+FUNC\s+.*?\s+UND\s+(%s)\s?.*$" % call
+    return re.compile(r)
+
 
 class BinaryInfo:
 
@@ -51,6 +57,18 @@ class BinaryInfo:
     setuid_call_regex = create_regexp_call(['setresuid', 'seteuid', 'setuid'])
     setgroups_call_regex = create_regexp_call(['initgroups', 'setgroups'])
     chroot_call_regex = create_regexp_call('chroot')
+
+    forbidden_functions = Config.getOption("WarnOnFunction")
+    if forbidden_functions:
+        for name, func in forbidden_functions.items():
+            # precompile regexps
+            f_name = func['f_name']
+            func['f_regex'] = create_nonlibc_regexp_call(f_name)
+            if 'good_param' in func:
+                func['waiver_regex'] = re.compile(func['good_param'])
+            # register descriptions
+            addDetails(name, func['description'])
+
     chdir_call_regex = create_regexp_call('chdir')
     mktemp_call_regex = create_regexp_call('mktemp')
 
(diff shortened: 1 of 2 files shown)


BinariesCheck.py
```diff
@@ -51,6 +57,18 @@ class BinaryInfo:
     setuid_call_regex = create_regexp_call(['setresuid', 'seteuid', 'setuid'])
     setgroups_call_regex = create_regexp_call(['initgroups', 'setgroups'])
     chroot_call_regex = create_regexp_call('chroot')
+
+    forbidden_functions = Config.getOption("WarnOnFunction")
+    if forbidden_functions:
+        for name, func in forbidden_functions.items():
```
tomato42: I've used `items()` instead of `iteritems()` as the former is Python 3 compatible

BinariesCheck.py
```diff
@@ -145,6 +169,26 @@ def __init__(self, pkg, path, file, is_ar, is_shlib):
...
+                if not res[0]:
+                    for l in res[1].splitlines():
+                        # as we need to remove elements, iterate backwards
+                        for i in range(len(self.forbidden_calls)-1, -1, -1):
+                            func = self.forbidden_calls[i]
+                            try:
+                                f = BinaryInfo.forbidden_functions[func]
+                                r = f['waiver_regex'].search(l)
+                                if r:
+                                    del self.forbidden_calls[i]
```
tomato42: should I put the above two lines in an `else` statement or leave them as-is?

scop: Please check for existence of waiver_regex before invoking it instead of using try...except here.

scop:
To try out the patch, I need some packages to test with: one or more that causes messages to be emitted and one or more that matches the forbidden regex but which is silenced by good_param, and the configuration to test those.

tomato42:
will do. One technical question: you prefer for me to overwrite this pull request with new patches, or creating a new pull request with [v2] in title and closing this one as obsolete. The former will drop comments made under specific commits while the latter can be said to clutter the pull request history. (asking since there's no CONTRIBUTING file...)

scop:
No strong opinions, but maybe leaning slightly towards the former option.

tomato42:
code updated

the `config` file to test this needs to contain:
```
bad_crypto_warning = \
'''This application package calls a function to explicitly set crypto ciphers
for SSL/TLS. That may cause the application not to use the system-wide set
cryptographic policy and should be modified in accordance to:
https://fedoraproject.org/wiki/Packaging:CryptoPolicies'''

call_blacklist = {'crypto-policy-non-compliance-openssl' :
                                     {'f_name' : 'SSL_CTX_set_cipher_list',
                                      'description' : bad_crypto_warning},
                  'crypto-policy-non-compliance-gnutls-1' :
                                      {'f_name' : 'gnutls_priority_set_direct',
                                       'description' : bad_crypto_warning},
                  'crypto-policy-non-compliance-gnutls-2' :
                                      {'f_name' : 'gnutls_priority_init',
                                       'good_param' : 'SYSLOG',
                                       'description' : bad_crypto_warning}
                 }
setOption("WarnOnFunction", call_blacklist)
```

testable packages that fail are:
 * https://kojipkgs.fedoraproject.org//packages/dovecot/2.2.9/1.fc20/x86_64/dovecot-2.2.9-1.fc20.x86_64.rpm
 * https://kojipkgs.fedoraproject.org//packages/cyrus-imapd/2.4.17/9.fc22/x86_64/cyrus-imapd-2.4.17-9.fc22.x86_64.rpm

testable packages that pass because of waiver are:
 * https://kojipkgs.fedoraproject.org//packages/ngircd/22/2.fc22/x86_64/ngircd-22-2.fc22.x86_64.rpm

scop:
Thanks! 48940f69552737b649ac0edfc8579763165abb11 adds your cases to the test suite.

PR #50, opened 2015-10-22
Workaround slowness due to python-magic bug

StefanBruens:
The file method evaluates the file twice, the descriptor method does not
show this problem. For large packages (e.g. kernel), this saves several
minutes of run time.


StefanBruens:
Superseded by https://github.com/rpm-software-management/rpmlint/pull/51

PR #53, opened 2015-10-23
Report which checks have been run and how long each took to finish

StefanBruens:
This is especially useful for long running checks to give some feedback
and to determine which checks are slow to run.

### DIFF
--- a/rpmlint
+++ b/rpmlint
@@ -18,6 +18,7 @@ import re
 import stat
 import sys
 import tempfile
+import time
 
 # 1 instead of 0 here because we want the script dir to be looked up first,
 # e.g. for in-place from tarball or SCM checkout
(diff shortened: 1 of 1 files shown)


scop:
Looks fine on surface in principle, but I think some improvements are in order:
- Instead of `print`, `printInfo` should be used (or an explanation provided why not).
- If a config option is added, it should be documented in the `config` file like the rest.
- I don't think this warrants a config option in the first place. I'm leaning towards not having one, and also removing the added `progress` flag, and enabling this stuff output based on the existing `verbose` flag.

StefanBruens:
printInfo is buffered, thus not useful for continuous progress reporting.

The verbose flag and progress reporting are orthogonal. The verbose flag enables additional output from a check, whereas the progress flag enables output regarding the whole check sequence.

scop:
Is printInfo really buffered (beyond? At least with python 3 it's just print. And FWIW the rpmlint executable runs python with the -u flag. IIRC the intent of fiddling with stdout.buffer in python < 3 is related to char encoding issues, not buffering.

--verbose is documented to just "Operate in verbose mode", and it already enables some progress-like output, search for "checking" in the rpmlint executable.

scop:
There's also the tty vs non-tty can of worms (related to encodings) which will bite if using print directly.

StefanBruens:
By buffered I meant "swallowed by the big rpmlint internal buffer and only printed after all tests have run". That is useless for a progress output.
If you want to see the checking messages during the checks, you have to use --rawout

StefanBruens:
Yes, this is actually the diagnostics buffer added by setting the threshold - although I am wondering why setting Badness/Threshold changes the output behaviour ...

setBadness is used by Tizen and openSUSE:

https://build.tizen.org/package/view_file?expand=1&file=config&package=rpmlint-tizen&project=Tizen%3AIVI%3ARelease%3Abuild

https://build.opensuse.org/package/view_file/openSUSE:Factory/rpmlint/config?expand=1

Conan-Kudo:
@StefanBruens Mageia (a Mandriva successor) uses setBadness, see below:

http://svnweb.mageia.org/packages/cauldron/rpmlint/current/SOURCES/rpmlint.config?view=markup

http://svnweb.mageia.org/packages/cauldron/rpmlint-mageia-policy/current/SOURCES/

scarabeusiv:
This is sadly no longer applicable due to the changes in the script but feel free to review it and I would point your eyes to ticket #216 which should get what you are after.

PR #51, opened 2015-10-23
File magic enhancements

StefanBruens:
Supersede https://github.com/rpm-software-management/rpmlint/pull/50

PR #54, opened 2015-11-01
File magic enhancements

StefanBruens:
Supercedes #51


scop:
Thanks. The symlink string wasn't quite the same as the one from file yet, and the file string from descriptor is apparently bytes, not a string so it broken on Python 3. Both fixed in followup commits.

scop:
Looks like the file version stuff testing/documentation was forgotten from this PR, also fixed in a followup commit.

Issue #46, opened 2015-05-19
remove files-attr-not-set check


scop:
%defattr(-,root,root) is default since rpm 4.4, released > 10
years go so it's about time to remove that check

Original comment by: l-n

scop:
0001-remove-files-attr-not-set-check.diff from original ticket: https://gist.github.com/scop/93804614817c7ee6a1e8

scarabeusiv:
this was merged in.

PR #56, opened 2015-11-16
remove files-attr-not-set check

lnussel:
%defattr(-,root,root) is default since rpm 4.4, released > 10
years go so it's about time to remove that check


lnussel:
closes #46

dirkmueller:
Ping? this is policy in Fedora even meanwhile I was told (to not set %defattr)

scop:
Just wondering if existence of this check causes an actual problem, i.e. why remove it (besides that most configs end up filtering it and thus need one line in config)? Are there any supported distro versions out there that have only rpm < 4.4? What are the rpm versions corresponding to the last few LSB versions?

dirkmueller:
There is no actual rpm version dependency in the LSB spec as far as I can see, but debian ships rpm 4.8. 

The check wastes valueable cpu cycles and in case the filtering is missing it is spreading outdated advice, which we should prevent imho.

scop:
The oldest rpm versions in supported distros I'm aware of are 4.4.2.3 on RHEL 5.x and friends. If someone knows of a relevant still supported distro that has rpm < 4.4 only, let me know.

PR #57, opened 2015-11-16
add check for tmpfiles created at runtime

lnussel:
this check parses files in /usr/lib/tmpfiles.d and verifies that
entries that create files or directories are actually listed in
%files.

The check also handles the ghost file check as rpmlint shouldn't
complain about ghost files handled by the tmpfiles mechanism.

### DIFF
--- a/PostCheck.py
+++ b/PostCheck.py
@@ -11,8 +11,6 @@
 import os
 import re
 
-import rpm
-
 from Filter import addDetails, printError, printWarning
 import AbstractCheck
 import Config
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@lnussel Could you rebase this against current master?

lnussel:
I wont have time for that any time soon unfortunately

scarabeusiv:
This is a part of the stockpile we have now in #227 as it was added to the suse-only specific part.
As such this will be hopefully included when we convert those tests to rpmlint 2.0.

PR #58, opened 2015-11-29
Do not treat all packages with lib in name as libs

keszybz:
python-matplotlib, python-joblib, and many other packages have names
which end in "lib". When they are used as a dependency, this causes
a confusing false-positive "E: explicit-lib-dependency".

Modify the pattern to match 'something-lib' but not 'somethinglib'.

### DIFF
--- a/TagsCheck.py
+++ b/TagsCheck.py
@@ -415,6 +415,7 @@ devel_number_regex = re.compile('(.*?)([0-9.]+)(_[0-9.]+)?-devel')
 lib_devel_number_regex = re.compile('^lib(.*?)([0-9.]+)(_[0-9.]+)?-devel')
 invalid_url_regex = re.compile(Config.getOption('InvalidURL'), re.IGNORECASE)
 lib_package_regex = re.compile('(?:^(?:compat-)?lib.*?(\.so.*)?|libs?[\d-]*)$', re.IGNORECASE)
+python_package_regex = re.compile('^python[23]?-(?!libs)|.*-python[23]?$')
 leading_space_regex = re.compile('^\s+')
 license_regex = re.compile('\(([^)]+)\)|\s(?:and|or)\s')
 invalid_version_regex = re.compile('([0-9](?:rc|alpha|beta|pre).*)', re.IGNORECASE)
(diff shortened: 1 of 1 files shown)


keszybz:
> Sure, but on the other hand there are packages like id3lib, giflib, zlib, zziplib and others which would no longer be treated as lib packages with this change. Not sure if we can win here, but reporting more is better than reporting less -- known false positives can always be easily filtered out in configs.

OK, so let's start with a small first step and filter out python packages. This will suffice for python-matplotlib and python-joblib where I saw this false positive most often. I force pushed to replace the previous commit.

scop:
Note that I wrote "filtered out in configs", and I mean exactly that :) So instead of making any code changes whatsoever, this can be done in configs shipped by distro rpmlint packages.

keszybz:
I fail to see an advantage in emitting the warning and filtering it out. This change seems to be something that applies generally across distributions. I mean you can can add overrides and whatnot, but it'd be nicer to just make things work out of the box.

scop:
I don't think doing it in small steps like this while hardcoding those small steps one or few at a time, then waiting for an shipping a new rpmlint release etc is going to be a very efficient way of getting it all done. And there's never going to be enough knowledge about specific cases within rpmlint, whereas interest groups for lets say language foo could just drop in their configs in rpmlint's config dirs.

keszybz:
It seems just like a way to overcomplicate things. Having a SIG create (and distribute!) a config file instead of trimming down an overly generic test is not cost free. rpmlint has a test which is overly generic, so why not fix it here? Stay close to upstream, etc, etc :)

scop:
I'm afraid that as long as I'm practically equivalent to rpmlint upstream, we'll just have to agree to disagree on this.

keszybz:
Yeah, I can see that :(

I was planning to submit a patch to filter out common false-positive spelling errors from rpmlint. Would that also be outside of what you want to see in rpmlint?

keszybz:
I pretty much gave up on using rpmlint because of the prohibitive amounts of false postives ;( But yeah, if rpmlint could be updated to be at least semi-adequate for modern packages, I'd be interested.

scarabeusiv:
We are trying to change the state, overall most important aspect is really to have it testcased, so if you could generate lists of false postives for the lists we could basically be 'safe' when doing this change while validating against the list (I for one as maintainer of libreoffice really hate seeing this everytime I compile it :)).

keszybz:
FWIW, I tried running rpmlint, and I don't see this error. E.g. running `rpmlint python3-pandas` doesn't give any warning about the dependency on python3-matplotlib. This patch doesn't apply any more, so let's close this PR. If I still see the problem, I'll rebase and reopen.

PR #59, opened 2015-11-30
fix shared library matching

lnussel:
Avoids e.g.
[  332s] glib2-devel.i586: E: library-without-ldconfig-postun (Badness: 300) /usr/share/gdb/auto-load/usr/lib/libglib-2.0.so.0.4600.1-gdb.py


scop:
Could you add some regression tests for this? Extract the regex checks into separate functions to the extent it makes sense for testability and add them to test/test_files.py?

lnussel:
I guess I could but I'll be traveling the next two weeks so don't expect it to happen soon.

btw, for testing the openSUSE rpmlint package I also have a test suite. Maybe it could be useful for upstream as well, or at least give some inspiration: https://github.com/openSUSE/rpmlint-tests

Conan-Kudo:
@lnussel Could you rebase this against current master and add the tests as @scop previously requested?

scarabeusiv:
This is no longer applying and we have it as separate patch in the list of suse patches to review, thus hopefully this will get fixed when the porting part is complete.

PR #62, opened 2016-02-22
Avoid messing with the error encoding (Fixes #61)

### LINKED ISSUE #61: Crash with non ascii characters
If there are non ascii / utf8 characters in the message rpmlint crashes.
E.g. in my case a "non-utf8-filename" warning with latin1 characters:

    [  377s] Traceback (most recent call last):
    [  377s]   File "rpmlint.py", line 385, in &lt;module&gt;
    [  377s]   File "rpmlint.py", line 196, in main
    [  377s]   File "Filter.py", line 131, in printAllReasons
    [  377s]   File "Filter.py", line 40, in __print
    [  377s]   File "/usr/lib/python2.7/codecs.py", line 357, in write
    [  377s] UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 56: ordinal not in range(128)

A simple fix, but maybe not perfect would be:
http://pastebin.com/uN36hKnL

dirkmueller:
Convert str() to unicode() on python 2 before printing it.

### DIFF
--- a/Filter.py
+++ b/Filter.py
@@ -25,20 +25,20 @@ _diagnostic = list()
 _badness_score = 0
 printed_messages = {"I": 0, "W": 0, "E": 0}
 
-if sys.stdout.isatty():
-    def __print(s):
-        print(s)
-else:
-    __stdout = sys.stdout
-    if not __stdout.encoding:  # Python < 3 only?
-        import codecs
-        if hasattr(__stdout, "buffer"):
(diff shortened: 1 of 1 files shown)


dirkmueller:
Thanks for the comments, this version also passes the regression check

   python2 ./rpmlint -c TagsCheck test/binary/autofs-5.0.7-28.fc19.i686.rpm >/dev/null

PR #64, opened 2016-02-23
Remove always-false conditional block

Fak3:
The inner if-block is only true for python3. The outer block only true for python2


Fak3:
ok

scop:
Could you squash these two commits into a single one with the correct commit message and force-push that to your PR branch? Want cleaner history :)

Fak3:
I will create new PR

PR #68, opened 2016-03-13
BinariesCheck: lower memory requirements, fix chroot/chdir detection

StefanBruens:
Do not read whole output of objdump -d into memory, but read and process
the output while it is created (issue#67).
Also correct expression to find 'chdir@plt' in output (issue#66)

### DIFF
--- a/BinariesCheck.py
+++ b/BinariesCheck.py
@@ -216,23 +216,23 @@ class BinaryInfo(object):
                     chroot_index = -99
                     chdir_index = -99
                     for line in p.stdout:
-                        r = BinaryInfo.objdump_call_regex.search(line)
-                        if not r:
+                        res = BinaryInfo.objdump_call_regex.search(line)
+                        if not res:
                             continue
-                        if b'@plt' not in r.group(1):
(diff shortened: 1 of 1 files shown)


scop:
With python 3.4.3 I'm getting a `binaryinfo-objdump-failed` warning for all of these objdump calls. The same command succeeds when run directly on the console, there's nothing in stderr, and stdout looks ok (both separately and when run inside rpmlint), so I didn't spot an obvious reason why this occurs.

A minor style nit, the regex search return value variable name we tend to use is `res`, so changing `r` to that would be good for consistency.

StefanBruens:
Python 3 no longer disables signals for child processes:
http://bugs.python.org/issue1652#msg101040

In case both chroot/chdir are found, the loop is exited via break, the pipe is closed by the "with p.stdout" contextmanager, and objdump gets a SIGPIPE.

This behaviour is actually nicer than with python2, as any further output from objdump is irrelevant. The best solution is to ignore the return code if chroot_near_chdir is set.

scop:
Applied, thanks.

PR #69, opened 2016-03-17
Add check for validating file extensions

dirkmueller:
Sometimes upstream tarballs are inconsistently named
and use a different compression format than what the
extension indicates. Add a warning for that case.


SourceCheck.py
```diff
@@ -30,6 +37,12 @@ def check_source(self, pkg):
         # process file list
         spec_file = None
         for fname, pkgfile in pkg.files().items():
+            file_ext = fname.rpartition('.')[2]
+            if (file_ext in compressed_fileext_magic and
+                    pkgfile.magic and
+                    compressed_fileext_magic[file_ext] not in pkgfile.magic):
+                printWarning(pkg, 'inconsistent-fileextension', fname)
```
bmwiedemann: you could add pkgfile.magic as an additional hint here.

scop:
Could you split the actual test logic into a separate function and add some test cases for it? See e.g. `test/test_binaries.py` for some examples (this test would belong in `test/test_source.py`).

SourceCheck.py
```diff
@@ -60,6 +73,9 @@ def check_source(self, pkg):
 '''A file that you listed to include in your package has strange
 permissions. Usually, a file should have 0644 permissions.''',
 
+'inconsistent-fileextension',
```
scop: I think `inconsistent-file-extension` would be a better name for this.

dirkmueller:
will look at adding a test.

dirkmueller:
test added, passes testing, merging

PR #70, opened 2016-03-17
Fix file magic handling for magic 5.25+

dirkmueller:


scop:
If we're no longer using `magic.descriptor()`, there' no need for the check, and the line modified in the 2nd commit can be just removed.

Also, if this stuff goes away, 9d3464d4bf4c724b076e36b96a6075d2d48264ba should probably be reverted too.

@StefanBruens, comments?

dirkmueller:
actually the commit https://github.com/file/file/commit/f0adf3e6811aafde00dba8ba236609869a973769 isn't released yet, so maybe its premature to revert the workaround.

Conan-Kudo:
@dirkmueller Do we need this anymore?

dirkmueller:
don't remember, lets close this.

PR #79, opened 2016-09-13
shlib-with-non-pic-code: add text on how to figure files compiled without -fPIC

nmav:


scop:
Do you have an affected shared lib file somewhere, or a link to one that I could try this out with?

Also, `s/with no -fPIC/without -fPIC/` would be better in the added text I think.

nmav:
A library to test with is https://kojipkgs.fedoraproject.org//packages/gnutls/3.5.2/1.fc25/i686/gnutls-3.5.2-1.fc25.i686.rpm

scop:
Thanks!

PR #80, opened 2016-10-05
 Skip chroot-without-chdir check on non-x86_64

dirkmueller:
This is looking through disassembly and searches for callq, which
    is x86_64 specific assembler. the whole thing does not make sense
    on any architecture, so stop warning about it.

### DIFF
--- a/BinariesCheck.py
+++ b/BinariesCheck.py
@@ -207,7 +207,9 @@ class BinaryInfo(object):
 
             # check if chroot is near chdir (since otherwise, chroot is called
             # without chdir)
-            if self.chroot and self.chdir:
+            # Currently this implementation works only on x86_64 due to reliance
+            # on x86_64 specific assembly. Skip it on other architectures
+            if pkg.arch == 'x86_64' and self.chroot and self.chdir:
                 p = subprocess.Popen(
                     ['env', 'LC_ALL=C', 'objdump', '-d', path],
                     stdout=subprocess.PIPE, bufsize=-1)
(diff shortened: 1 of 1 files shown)


scop:
The description and code seem to be off sync. If the intent is to skip all non-x86, I suppose there's at least a bunch of ix86 archs missing from the code (and maybe pentium*, dunno)? Also, the description's latter sentence (the "does not make sense ... so stop warning") doesn't match what the code actually does. Could you clarify?

dirkmueller:
@scop sorry for the confusion. I updated the patch and the git commit message. the intention is to only run it on x86_64, since "callq" is not a valid assembly on any other architecture, hence does not occur and then creates false warnings.

scop:
@dirkmueller ok, thanks. The commit message still says "the whole thing does not make sense
on any architecture, so stop warning about it." -- shouldn't that comment be removed?

dirkmueller:
@scop I've removed the extra sentence

scop:
Thanks!

pihhan:
I think you have missed callq? pattern. It matches BOTH callq AND call, so this pull is in my opinion unnecessary and wrong.

dirkmueller:
@pihhan there are many architectures out there where subroutine calls are not matching "callq?" as a pattern, so they are still causing this warning incorrectly. If you assume anything about assembly then it should at least check for the architecture.

pihhan:
@dirkmueller They are causing this error incorrectly, because it does not handle situation where calls are not found at all. But chroot and chdir are dynamically linked. That might happen even on supported architecture. So it hides invalid test for failing architectures instead of not failing it in these cases. See issue #84

pihhan:
In my opinion i686 should be checked as before. Do not report error if you are not able to validate. Maybe it should be changed to warning, unless we have reliable detection.

PR #86, opened 2016-12-30
FilesCheck: Update Python 3.6 bytecode magic value again

AdamWill:
This is the value for 3.6rc1, which == 3.6.0.


dirkmueller:
Do you have any pointer to the cpython source equivalent of this value?

Conan-Kudo:
@dirkmueller See here: https://github.com/python/cpython/blob/master/Lib/importlib/_bootstrap_external.py#L230-L242

dirkmueller:
Thanks!

AdamWill:
@dirkmueller the comment just above the changed line actually includes the link. :)

PR #93, opened 2017-02-02
add /usr/lib/systemd to DEFAULT_HARDCODED_LIB_PATH_EXCEPTIONS

larsks:
Spec files that need to install files into, e.g.,
/usr/lib/systemd/system-generators will generate a
hardcoded-library-path error.  This commit adds /usr/lib/systemd to
the list of exceptions for this check.

### DIFF
--- a/SpecCheck.py
+++ b/SpecCheck.py
@@ -30,7 +30,7 @@ DEFAULT_BIARCH_PACKAGES = '^(gcc|glibc)'
 # Don't check for hardcoded library paths in packages which can have
 # their noarch files in /usr/lib/<package>/*, or packages that can't
 # be installed on biarch systems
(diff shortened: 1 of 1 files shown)


scop:
Seems otherwise legit, but those packages should not be hardcoding paths but instead use macros provided in `/usr/lib/rpm/macros.d/macros.systemd` (e.g. `%_systemdgeneratordir` for `/usr/lib/systemd/system-generators`). So I'd rather not apply this unless there's a compelling reason to not use these macros in some setups.

larsks:
These macros don't exist in eg centos/rhel 7, causing rpmlint to throw errors for which there is no alternative.

scop:
Ok, applied with that info amended to the commit message.

PR #94, opened 2017-02-14
add shared-lib-not-executable check for libraries without exec bit set

proyvind:


marxin:
@proyvind Are you really sure that shared libraries without executable bit set are so problematic? I see quite such libraries in my system.

scarabeusiv:
Followed up in #286

PR #96, opened 2017-02-14
Fix paths to extracted srpm files

proyvind:


scarabeusiv:
As this conflicts and failed to build even before, please if this is still something usefull create a new PR with tad longer explanation and we will try to get it merged.

PR #97, opened 2017-02-14
add a -T option for making badness below threshold return 0

proyvind:
This will make it more feasible to enable mandatory rpmlint checking
when one can operate with acceptable badness levels to avoid anal
bottlenecks (as rpmbuild will fail if returning anything else than 0)

### DIFF
--- a/rpmlint
+++ b/rpmlint
@@ -100,6 +100,7 @@ def main():
 
     packages_checked = 0
     specfiles_checked = 0
+    threshold_exceeded = False
 
     try:
         # Loop over all file names given in arguments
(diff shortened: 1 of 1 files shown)


rpmlint
```diff
@@ -321,6 +330,9 @@ for o in opt:
         Config.setOption('ExtractDir', extract_dir)
     elif o[0] in ('-n', '--noexception'):
         Config.no_exception = True
+    # terrible longname, but whatever..
+    elif o[0] in ('-T', 'zerothreshold'):
```
Conan-Kudo: I don't like this option, it's poorly named and doesn't explain what it does...

scarabeusiv:
This is no longer sadly applicable as the rpmlint is now behaving like proper entry_point. If it is something you would still like to see feel free to check out rpmlint/lint.py file and see how it behaves now.

Conan-Kudo:
Re-opening so that I know to port this over. It's apparently in use on Mandriva, ALT, PLD, and Mageia for their buildsystem rpmlint.

scarabeusiv:
Hmm, isn't this integrated currently? Take look on the threshod and badness scoring present in rpmlint/lint.py I guess I just did it when I was taking the logic from the rpmlint i had on the system when rewritting it.

Conan-Kudo:
@scarabeusiv Let me double check, but I thought that we don't have this behavior right now (any badness value triggers non-zero exit code...)

Conan-Kudo:
@scarabeusiv It looks like we still don't have this behavior:

```python
        if self.output.badness_threshold > 0 and self.output.score > self.output.badness_threshold:
            print_warning(f'(none): E: Badness {self.output.score} exceeeds threshold {self.output.badness_threshold}, aborting.')
            return 66
        if self.output.printed_messages['E'] > 0:
            return 64
```

scarabeusiv:
Yea, like here if threshold is set and it is not over the value it returns 0 even with errors, done by the `self.output.score > self.output.badness_threshold` part.
Instead of argument it is config option called threshold. But I suppose I really reflect it with https://github.com/rpm-software-management/rpmlint/pull/327/commits/cf92813c393d475926baf204e8b8101215c436dd where it goes llike:

Is there threshold and was it exceeded 66 otherwise zero
If there is no threshold and any errors 64 otherwise zero

Conan-Kudo:
Superseded by #384.

PR #98, opened 2017-02-14
add missing LGPL 2.1 license

proyvind:


proyvind:
Well, considering that LGPL 2.1 is the commonly used version of LGPL used rather than LGPL 2, that should speak for itself.

A question more sensible to ask is why listing LGPLv2 when LGPL 2.1 revision was published back in 1999.
So if anything, LGPLv2 should rather be LGPLv2.1, but since it's from a theoretical point of view could exist projects out there still using LGPL 2.0, or that would like to use it over 2.1, so keeping LGPLv2 for such (even if unlikely) possible cases isn't not not sensible either.. ;)

scop:
IIRC the thing that lead to 2.1 not specified separately was that there's not much practical difference between LGPL 2 and 2.1 and thus there's no need to separate them in license tags. Then again it's possible that I just misremember, but anyway I don't care much nor do I find maintaining a list of licenses within rpmlint too interesting, as that's pretty much a distro specific config thing. Maybe the defaults list should instead just be shipped empty, or changed to include [SPDX identifiers](https://spdx.org/licenses/) (most likely autogenerated from somewhere). Thoughts?

dirkmueller:
SUSE's copy of rpmlint does exactly that - automatically include the list of spdx defined licenses. I'm happy to upstream that change if it is no longer a problem for other distros. Afaik we're the only ones using spdx

Conan-Kudo:
@dirkmueller The SPDX thing is under consideration in Fedora, and is used by Clear and Yocto, so if you could upstream the change in a way that could make it optional to enable, that'd be stellar.

Conan-Kudo:
@proyvind The difference between LGPLv2 and LGPLv2.1 is that the former calls it "Library" and the latter calls it "Lesser". Thus it was never considered significant enough to differentiate.

scop:
@dirkmueller is the SUSE SPDX implementation just a matter of replacing the contents of the valid licenses list, or does it alter the matching rules in any way? If it's just a matter of updating the list and doesn't break existing license string splitting functionality, I'd say just go ahead with it. Or submit a pull request for review if you want comments first. (If it doesn't break any existing functionality e.g. for a distro that just defines their own license list, there's no need to make it optional.)

proyvind:
If the license used is LGPLv2.1, then the license specified should be LGPLv2.1, not LGPLv2 regardless of the insignificant differences between the two.
Replacing LGPLv2 with LGPLv2.1 seems to be the more sensible topic to discuss in this retard IMO.

Conan-Kudo:
I'm closing this, as the license list is already a mess, and I'd really like to have a better way to handle this in general.

@dirkmueller If you want to propose adding support for pulling SPDX identifiers automatically and evaluating those, then I'm happy to review that.

I'm thinking that I'd like to have a toggle for different license lists. SPDX, Fedora, and whatever the hell this list is...

Issue #84, opened 2016-11-21
False positives reporting missing-call-to-chdir-with-chroot 


pihhan:
Rpmlint reports errors about chroot also in cases, where really no such error is present. It reports for example on bind99-libs package on Fedora 24. It looks for @plt sections in disassembly. But for some libraries, especially on x86_64, they might not be present. If binary links both chroot and chdir AND @plt calls are missing, error is reported. And it is error, not only warning. And is wrong.

Example of disassembly for bind99-libs, function isc_dir_chroot, that only contains both chdir and chroot calls:
```   
    000000000003cf60 <isc_dir_chroot@@Base>:
       3cf60:       48 83 ec 08             sub    $0x8,%rsp
       3cf64:       48 85 ff                test   %rdi,%rdi
       3cf67:       74 47                   je     3cfb0 <isc_dir_chroot@@Base+0x50>
       3cf69:       e8 0a f4 fc ff          callq  c378 <_init@@Base+0x748>
       3cf6e:       85 c0                   test   %eax,%eax
       3cf70:       78 1e                   js     3cf90 <isc_dir_chroot@@Base+0x30>
       3cf72:       48 8d 3d 0d c2 00 00    lea    0xc20d(%rip),%rdi        # 49186 
    <isc_version@@Base+0x4b86>
       3cf79:       e8 0a ed fc ff          callq  bc88 <_init@@Base+0x58>
       3cf7e:       85 c0                   test   %eax,%eax
       3cf80:       78 0e                   js     3cf90 <isc_dir_chroot@@Base+0x30>
       3cf82:       31 c0                   xor    %eax,%eax
       3cf84:       48 83 c4 08             add    $0x8,%rsp
       3cf88:       c3                      retq   
```

pihhan:
Can be fixed by this:
```
diff --git a/BinariesCheck.py b/BinariesCheck.py
index e01c1ea..848e1e7 100644
--- a/BinariesCheck.py
+++ b/BinariesCheck.py
@@ -237,6 +237,8 @@ class BinaryInfo(object):
                 if p.wait() and not self.chroot_near_chdir:
                     printWarning(pkg, 'binaryinfo-objdump-failed', file)
                     self.chroot_near_chdir = True  # avoid false positive
+                elif chroot_index == -99 and chdir_index == -99:
+                    self.chroot_near_chdir = True  # avoid false positive
 
         else:
             self.readelf_error = True
```

scarabeusiv:
[...]

PR #100, opened 2017-02-28
Chroot pic fix

pemensik:
Do not report missing-call-to-chdir-with-chroot error if call positions are unknown. Resolves issue #84 and reverts #80, because it contains proper fix for the error.


pemensik:
This fix stops false reporting on Fedora packages bind99-libs and rsyslog

dirkmueller:
@pemensik I don't think the revert is related to that, because that change is making sure that we don't try to do call semantics on architectures where it is not called "call" (e.g. ARM, powerpc)

dirkmueller:
I've cherry-picked the other commit. thanks!

pemensik:
Ok, it is extended back to x86 as well by previous commit, so ok. Thank you for the merge.

PR #105, opened 2017-07-06
rpmdiff: Make rpmdiff library

Kratochvilova:
This is required for using/extending rpmdiff in third party python code.


scarabeusiv:
I implemented the solution in PR https://github.com/rpm-software-management/rpmlint/pull/164 Thus this can be considered done.

PR #107, opened 2017-07-10
Patch originally written by Mark Wielaard in Fedora bugzilla.

spotrh:
Add .build-id to regex to ignore for only-non-binary-in-usr-lib.
Ignore .build-id in hidden-file-or-dir checks.
.build-id was added in upstream RPM: https://github.com/rpm-software-management/rpm/commit/bbfe1f86b2e4b5c0bd499d9f3dd9de9c9c20fff2


BinariesCheck.py
```diff
@@ -300,7 +300,7 @@ def __init__(self, pkg, path, file, is_ar, is_shlib):
 bin_regex = re.compile(r'^(/usr(/X11R6)?)?/s?bin/')
 soversion_regex = re.compile(r'.*?([0-9][.0-9]*)\\.so|.*\\.so\\.([0-9][.0-9]*).*')
 reference_regex = re.compile(r'\.la$|^/usr/lib(64)?/pkgconfig/')
-usr_lib_exception_regex = re.compile(Config.getOption('UsrLibBinaryException', r'^/usr/lib(64)?/(perl|python|ruby|menu|pkgconfig|ocaml|lib[^/]+\.(so|l?a)$|bonobo/servers/)'))
+usr_lib_exception_regex = re.compile(Config.getOption('UsrLibBinaryException', r'^/usr/lib(64)?/(perl|python|ruby|menu|pkgconfig|ocaml|lib[^/]+\.(so|l?a)$|bonobo/servers/|\\.build-id)'))
```
scop: There's a double backslash escape here, note that the regex is a raw string nowadays. I'll push a fixed commit shortly.

PR #110, opened 2017-09-05
Fixes to rpmlint from Mageia

Conan-Kudo:
These are a pair of patches that @coling wrote for Mageia's rpmlint a few years back. As part of rebasing to the latest version, I'm sending these to merge upstream.

The third patch is a license content update that brings us more up to date with how licenses are tagged in Fedora/Mageia, done by @Remmy a while back...


scop:
Two first applied manually, with missing trailing periods in info messages added. Third one not, because wrt license identifiers, I'm not going to start or support maintaining distro specifics or a collection of them or their aliases in upstream rpmlint. SPDX is something I would support, see some discussion in #98.

Conan-Kudo:
@scop Okay, thanks. I had to give it a shot. :)

PR #116, opened 2017-09-29
Improve XDG Menu checks stability

dirkmueller:
Running RawConfigParser on untrusted input can cause a lot
of exceptions. Handle them gracefully and raise appropriate
rpmlint errors. Also separate the code a little and cleaning it up.


MenuXDGCheck.py
```diff
@@ -25,7 +25,44 @@ def __init__(self):
...
+        except cfgparser.DuplicateOptionError as e:
+            printError(
+                pkg, 'desktopfile-duplicate-option', filename,
+                '[%s]/%s' % (e.section, e.option))
+        except cfgparser.MissingSectionHeaderError:
+            printError(
+                pkg, 'desktopfile-missing-header', filename)
+        except (cfgparser.ParsingError, UnicodeDecodeError) as e:
+            printWarning(
+                pkg, 'invalid-desktopfile', filename)
```
scop: I think this deserves some information what made the desktopfile invalid, i.e. something from `e` in the message.

MenuXDGCheck.py
```diff
@@ -25,7 +25,44 @@ def __init__(self):
         # $ echo $XDG_DATA_DIRS/applications
         # /var/lib/menu-xdg:/usr/share
         AbstractCheck.AbstractFilesCheck.__init__(
-            self, "MenuXDGCheck", r"/usr/share/applications/.*\.desktop$")
+            self, "MenuXDGCheck", r'(?:/usr|/etc/opt|/opt/.*)/share/applications/.*\.desktop$')
```
scop: Since we're adding directories, I wonder if we should add subdirs (one level deep?) of these too, at least I have some  `/usr/share/applications/{kde,kde4}/*.desktop` on my Fedora 25 box.

dirkmueller: thats already covered by the regexp, right?

scop: Ah, right.

dirkmueller:
@scop better?

MenuXDGCheck.py
dirkmueller: the error isn't particularly helpful, it always says something like "Parse error while handling file: XYZ". I've included it.

MenuXDGCheck.py
```diff
@@ -25,7 +25,52 @@ def __init__(self):
...
+            printWarning(
+                pkg, 'invalid-desktopfile', filename,
+                e.message.partition(':')[0])
+        except cfgparser.Error as e:
+            # Only in Python >= 3.2
+            if (hasattr(cfgparser, 'DuplicateOptionError') and
+                    isinstance(e, cfgparser.DuplicateOptionError)):
+                printError(
+                    pkg, 'desktopfile-duplicate-option', filename,
+                    '[%s]/%s' % (e.section, e.option))
```
scop: Should probably output something also if it's not a DuplicateOptionError, some kind of generic else branch here?

dirkmueller: yeah, good point

dirkmueller:
@scop better?

scop:
Yup, looks good now on surface (not tested though).

PR #117, opened 2017-10-04
Handle %post scripts that contain non-ascii characters

dirkmueller:
when running LC_ALL=C and python3 this otherwise traces with:

  File "rpmlint/PostCheck.py", line 83, in check_syntax_script
    tmpfile.write(script)
UnicodeEncodeError: 'ascii' codec can't encode character '\xfc' in position 16: ordinal not in range(128)

Add test coverage.


scop:
Hm, I'm unable to reproduce the issue or to review this using the test subject package. Looks like at least my rpm (4.13.0.1-2.fc25 on Fedora 25) returns the supposedly non-utf8 content as utf8:

```
$ rpm -qp --scripts test/binary/Nonutfpostcheck-0-0.x86_64.rpm | file -i -
/dev/stdin: text/plain; charset=utf-8
```

dirkmueller:
Sorry, forgot reproducer: LC_ALL=C python3 rpmlint...

Conan-Kudo:
@dirkmueller So, to confirm, this breaks whenever a non-UTF-8 enabled locale is used, right? Does this really matter when Python will automatically coerce to a UTF-8 enabled locale?

dirkmueller:
@Conan-Kudo where does python automatically coerce to a utf-8 locale? if you run it in a non-utf8 locale and have utf8 embedded it will throw a decode exception..

Conan-Kudo:
@dirkmueller As of Python 3.7, it does automatic coercing to C.UTF-8 or equivalent locale: https://www.python.org/dev/peps/pep-0538/

Conan-Kudo:
That said, older Python versions wouldn't necessarily have this (it's backported in Fedora's Python 3, though).

The code looks fine, so I'll merge this.

PR #118, opened 2017-10-12
Binariescheck: Check for chroot/chdir on ARM, PPC

StefanBruens:
The assembly on ARM looks almost the same as the x86 assembly, but
with a branch mnemonic (bl) instead of an explicit call.

On PPC, library function calls use jump tables, so the vicinity check does
not work, but we can at least detect a sole chroot without chdir.

### DIFF
--- a/BinariesCheck.py
+++ b/BinariesCheck.py
@@ -54,8 +54,6 @@ class BinaryInfo(object):
     setuid_call_regex = create_regexp_call(r'set(?:res|e)?uid')
     setgroups_call_regex = create_regexp_call(r'(?:ini|se)tgroups')
     chroot_call_regex = create_regexp_call('chroot')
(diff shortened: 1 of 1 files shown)


BinariesCheck.py
```diff
@@ -96,6 +94,16 @@ def __init__(self, pkg, path, fname, is_ar, is_shlib):
         self.mktemp = False
 
         is_debug = path.endswith('.debug')
+        # Currently this implementation works only on specific
+        # architectures due to reliance on arch specific assembly.
+        if pkg.arch in ['armv6hl', 'armv7hl', 'aarch64']:
```
dirkmueller: should not be a list '[' ... ']' but a tuple '(' .. ')'

StefanBruens: changed to startswith('armv')

Issue #128, opened 2018-03-01
python-bytecode-inconsistent-mtime gives bogus result with Python 3.7 bytecode


hroncok:
The bytecode for 3.7 has changed and the way rpmlint tries to read mtime from the file is now bogus. See also https://bugzilla.redhat.com/show_bug.cgi?id=1550562

I'll try to prep a fix.

PR #129, opened 2018-03-01
Fix getting pyc mtime on Python 3.7

hroncok:
Fixes https://github.com/rpm-software-management/rpmlint/issues/128


FilesCheck.py
```diff
@@ -379,6 +379,25 @@ def py_demarshal_long(b):
...
+
+
+def pyc_mtime_from_chunk(chunk):
+    """From given chunk (beginning of the file), return mtime or None
+
+    From Python 3.7, mtime is not always present.
+    """
+    magic = pyc_magic_from_chunk(chunk)
+    second = py_demarshal_long(chunk[4:8])
+    if magic >= 3390:  # Python 3.7
```
encukou: Maybe you could add a link to the specification, i.e. https://www.python.org/dev/peps/pep-0552/#specification

FilesCheck.py
```diff
@@ -379,6 +379,27 @@ def py_demarshal_long(b):
...
+    """From given chunk (beginning of the file), return Python magic number"""
+    return py_demarshal_long(chunk[:4]) & 0xffff
+
+
+def pyc_mtime_from_chunk(chunk):
+    """From given chunk (beginning of the file), return mtime or None
+
+    From Python 3.7, mtime is not always present.
+
+    See https://www.python.org/dev/peps/pep-0552/#specification
```
hroncok: @encukou added here

FilesCheck.py
```diff
@@ -379,6 +379,27 @@ def py_demarshal_long(b):
...
+    """
+    magic = pyc_magic_from_chunk(chunk)
+    second = py_demarshal_long(chunk[4:8])
+    if magic >= _python_magic_values['3.7'][0]:
```
hroncok: changed to not hardcode the magic number here again.

test/pyc/.gitignore
```diff
@@ -0,0 +1 @@
+!*.pyc
```
hroncok: Forgot to add the pyc files, as they were gitignored (obviously), so added them now together with this.

FilesCheck.py
hroncok: added

Conan-Kudo:
This looks pretty good! I'll try to do some testing on something with Python 3.7 to be sure it doesn't break anything and then I'll merge it, if there are no issues.

hroncok:
bump?

Conan-Kudo:
Will take a look later today. I didn't have anything with Python 3.7 to use it with until recently.

Conan-Kudo:
Looks good to me, merging!

Conan-Kudo:
Thanks!

PR #130, opened 2018-03-27
Update TagsCheck.py, add FSFAP

c72578:
- Add FSFAP to DEFAULT_VALID_LICENSES
  For details see:
  https://fedoraproject.org/wiki/Licensing/FSFAP
  https://www.gnu.org/prep/maintain/html_node/License-Notices-for-Other-Files.html


Conan-Kudo:
LGTM.

PR #131, opened 2018-04-10
Add check of directory ownership

frenzymadness:
Hello.

We recently met a situation where package left some directories on a filesystem after uninstallation. After some investigation, we found out that this problem is caused by problematic definitions in %files section of specfile.

For example:
```
%{python3_sitelib}/pykickstart/commands/*py*
```

with this definition, rpm will install `commands` directory with its content but the package doesn't own `commands` directory itself which means that the directory will stay untouched after uninstallation.

I implemented check which checks ownership of all directories in a package and reports not owned directories as an error.

For testing, I have package python3-kickstart (just as an example) in two versions - one with right definitions in specfile and second one with wrong ones. I am using Fedora 27 in Docker container for testing.

```shell
$ dnf install /pykickstart/fixed_directory_ownership.rpm
...
Complete!
$ ./rpmlint -C . -c FilesCheck python3-kickstart
1 packages and 0 specfiles checked; 0 errors, 0 warnings.
$ dnf remove python3-kickstart
...
Complete!
$ dnf install /pykickstart/bad_directory_ownership.rpm
...
Complete!
$ ./rpmlint -C . -c FilesCheck python3-kickstart
python3-kickstart.noarch: E: not-owned-directory /usr/lib/python3.6/site-packages/pykickstart/handlers
python3-kickstart.noarch: E: not-owned-directory /usr/lib/python3.6/site-packages/pykickstart/commands
python3-kickstart.noarch: E: not-owned-directory /usr/lib/python3.6/site-packages/pykickstart
1 packages and 0 specfiles checked; 3 errors, 0 warnings.
```

Let me know if I can do anything or provide any further info.

Thank you and have a nice day.
Lumír


Conan-Kudo:
@frenzymadness This looks great, but could you please add a test for this?

See https://github.com/rpm-software-management/rpmlint/blob/master/test/test_files.py for our FilesCheck tests.

frenzymadness:
Thanks, I'll add tests ASAP. I'll create specfile, srpm and binary rpm of a package with wrong directory ownership and I'll prepare test.

Conan-Kudo:
Thanks!

frenzymadness:
I've added binary RPM with bad directory ownership issue but I am not sure how to design test.

As far as I understand, the tests in `test_files.py` work with binary RPMs as with files but to test directory ownership, the package has to be installed. Is there any way how can I install the package before a run of tests?

Conan-Kudo:
Probably not, I'm a bit surprised that there's no way to check with an rpm on whether it will own something on install, though.

@ffesti, @pmatilai, is there a way to check without installing?

frenzymadness:
> Probably not, I'm a bit surprised that there's no way to check with an rpm on whether it will own something on install, though.

I don't know how RPM works internally with its database. I tried to find a way how to check directory ownership without a installation of a package but from the source code of `rpm` tool, it seems that it uses its internal database to query some information from there and therefore I think that it's impossible to check directory ownership without the database. But I might be wrong of course.

frenzymadness:
Hello.

Is there any problem with this PR? I cannot find a way how to implement the same check without installing checked package so any help would be appreciated.

pmatilai:
A directory is owned by a package if it appears as a standalone entry in the package file list, that's all there is to it.

What makes it hard to *meaningfully* test and warn for unowned directories is that you need to filter out directories owned by dependencies, otherwise you'd only create so many false positives it becomes harmful rather than useful. And that's where the real problem lies: packages have all sorts of dependencies that are not required for building them, so they often would not be available in the rpmdb either. Hence, you need to install (simulate it) to calculate true unowned directories.

Conan-Kudo:
I'm wondering if this check is a good idea in rpmlint, then? Most distributions use rpmlint as part of the rpmbuild process (obviously Fedora is an exception), and it's probably not a good idea to do weird things with stuff like that.

frenzymadness:
Yep, I understand you. I wanted to add this check to Fedora Taskotron which checks all packages and updates but we decided that it would be nice to add this check to more RPM users and therefore rpmlint was chosen.

I know that for now is not possible to write some tests for this check but I still think that this check can be useful for a lot of users checking installed packages. And if a majority of users using rpmlint only during a build, this check won't have any effect.

frenzymadness:
Could you please take into consideration my last comment? Is there any chance that this check might be merged? If so, I can remove the WIP commit and we can proceed. Otherwise, I'll have to focus on another way how to deliver this check into out testing pipeline.

Conan-Kudo:
@frenzymadness Sorry, I think this check doesn't make sense in rpmlint due to how it _must_ function to work properly.

PR #132, opened 2018-05-03
get_default_valid_rpmgroups: Don't fail if rpm is partially installed

tmzullinger:
A catch-all exception was added to handle the case where the rpm package
was not installed dec4821 ("get_default_valid_rpmgroups: Don't fail if
the rpm package is not installed", 2015-07-24).  That catch-all was
later tightened in c1945e3 ("Avoid catch-all except statements",
2017-10-24).

The rpm package can also trigger a TypeError exception when installed
without docs or all languages, as is commonly done in container images.
Catch TypeError exception as well.

### DIFF
--- a/Pkg.py
+++ b/Pkg.py
@@ -227,7 +227,7 @@ def get_default_valid_rpmgroups(filename=None):
                 groupsfiles = [x for x in p.files() if x.endswith('/GROUPS')]
                 if groupsfiles:
                     filename = groupsfiles[0]
-        except KeyError:  # the rpm package might not be installed
+        except (KeyError, TypeError):  # the rpm package might not be installed
             pass
     if filename and os.path.exists(filename):
         with open(filename) as fobj:


tmzullinger:
This came up while looking into [rhbz#1574509](https://bugzilla.redhat.com/show_bug.cgi?id=1574509).  Since Fedora doesn't use the `Group:` tag and ships with rpm-4.14 which doesn't contain any GROUPS file, we're likely to simply disabled the `non-standard-groups` check by setting `ValidGroups = []`.

I don't know if that's worth making the default in rpmlint as well or if older rpm versions are still the larger target.  The GROUPS file was removed from rpm in [ 0d93b634b ("Delete GROUPS", 2013-01-16)](https://github.com/rpm-software-management/rpm/commit/0d93b634).  While the commit is 5 years old, it's only in `rpm >= 4.14`.

jasonmp85:
It's worth mentioning that d59bc2a1e2698040553c09610cb2befa1ef8d76a already addresses the `FileNotFoundError`… the intent seems to have been that if files were missing it would skip over them anyhow, but did so by erroneously catching `OSError` rather than `FileNotFoundError`.

This commit's not in any release yet, though.

tmzullinger:
Indeed.  I found that subsequently (only in the last half hour) while looking to patch the underlying issue, since it still causes issues when checking installed packages on systems where `%{_install_langs}` is restricted, as the Fedora container images are configured.  I suppose that makes this change unneeded. :)

tmzullinger:
Thanks for taking a look and providing the commit which fixed this issue at the root @jasonmp85.  I'll close PR.

PR #136, opened 2018-07-09
Patches from openSUSE that are always used there

scarabeusiv:
I tried to tweak and cherry pick all the relevant ones plus few minor fixes.

As discussed with Neal py2 should not be of concern thus I removed it where in compat, further planned changes are to thread the execution and it will work on py3 only anyway.


BinariesCheck.py
```diff
@@ -321,6 +321,7 @@ def __init__(self, pkg, path, fname, is_ar, is_shlib):
 srcname_regex = re.compile(r'(.*?)-[0-9]')
 invalid_dir_ref_regex = re.compile(r'/(home|tmp)(\W|$)')
 ocaml_mixed_regex = re.compile(r'^Caml1999X0\d\d$')
+usr_arch_share_regex = re.compile(r'/share/.*/(?:x86|i.86|x86_64|ppc|ppc64|s390|s390x|ia64|m68k|arm|aarch64)')
```
Conan-Kudo: We should also handle `riscv` and `mips` arches here...

NamingPolicyCheck.py
```diff
@@ -89,7 +89,8 @@ def check_binary(self, pkg):
 # if somone as a elegant solution, I will be happy to implement and test it.
 
 check.add_check('xmms', '^xmms(-|$)', '^/usr/lib(64)?/xmms/')
-check.add_check('python', '^python(-|$)', '^/usr/lib(64)?/python[1-9](-[1-9])?')
+check.add_check('python2', '^python(2)?(-|$)', '^/usr/lib(64)?/python[1-2](-[1-9])?')
```
Conan-Kudo: I don't know if it's reasonable to expect `python1.y` named as `python2` subpackage. 😜 

Could this check be made better to map major versions, such that when `python4` becomes a thing, it will be picked up correctly by policy?

SpecCheck.py
```diff
@@ -91,6 +88,8 @@ def re_tag_compile(tag):
 obsoletes_regex = re.compile(r'^Obsoletes:\s*(.*)', re.IGNORECASE)
 conflicts_regex = re.compile(r'^(?:Build)?Conflicts:\s*(.*)', re.IGNORECASE)
 
+lconfig_regex = re.compile(r'%{?run_ldconfig}?', re.IGNORECASE)
```
Conan-Kudo: What the heck is `%run_ldconfig`? Is it similar to this? https://fedoraproject.org/wiki/Changes/Removing_ldconfig_scriptlets#Upgrade.2Fcompatibility_impact

SpecCheck.py
```diff
@@ -632,12 +637,12 @@ def check_spec(self, pkg, spec_file, spec_lines=None):
...
 'invalid-spec-name',
-'''Your spec filename must end with '.spec'. If it's not the case, rename your
-file and rebuild your package.''',
+'''The spec file name (without the .spec suffix) must match the package name
+("Name:" tag). Either rename your package or the specfile.''',
 
 'non-utf8-spec-file',
-'''The character encoding of the spec file is not UTF-8.  Convert it for
-example using iconv(1).''',
+'''The character encoding of the spec file (or .changes file merged in the
```
Conan-Kudo: `<pkg>.changes` is SUSE-specific, please remove this.

SpecCheck.py
```diff
@@ -763,6 +768,14 @@ def check_spec(self, pkg, spec_file, spec_lines=None):
 '''Make check or other automated regression test should be run in %check, as
 they can be disabled with a rpm macro for short circuiting purposes.''',
 
+'deprecated-use-of-%run_ldconfig',
+'''According to the packaging conventions, the use of %run_ldconfig
+is deprecated. Please use /sbin/ldconfig instead, or
+
+%post(un) -p /sbin/ldconfig
+
+in the case where ldconfig is the only command to be executed.''',
```
Conan-Kudo: This has actually been changed in Fedora: https://fedoraproject.org/wiki/Changes/Removing_ldconfig_scriptlets

Please remove this change from this PR.

TagsCheck.py
```diff
@@ -1117,7 +1136,12 @@ def check_summary(self, pkg, lang, ignored_words):
 '''This rpm requires a specific release of another package.''',
 
 'no-url-tag',
-'''The URL tag is missing.''',
+'''The URL tag is missing. Please add a http or ftp link to the project location.''',
+
+'no-pkg-config-provides',
+'''The package installs a .pc file but does not provide pkgconfig(..) provides.
+The most likely reason for that is that it was built without BuildRequires: pkg-config.
```
Conan-Kudo: `BuildRequires: /usr/bin/pkg-config`

There are multiple compatible implementations of `pkg-config` binary.

FilesCheck.py
```diff
@@ -178,13 +178,14 @@
     '/var/tmp',
 )
 
+compressions = r'\.(gz|z|Z|zip|bz2|lzma|xz)'
```
Conan-Kudo: `zst` is also supported, since RPM 4.14.

FilesCheck.py
```diff
@@ -1042,7 +1054,7 @@ def check(self, pkg):
         for exe, paths in bindir_exes.items():
             if len(paths) > 1:
                 printWarning(pkg, "duplicate-executable", exe, paths)
-            if exe not in man_basenames:
+            if exe not in man_basenames and not exe.startswith("rc") and len(paths) is not 0:
```
Conan-Kudo: The `exe.startswith("rc")` I assume has to do with SUSE's weird `rcFOO` thing?

What is `len(paths) is not 0` for?

TagsCheck.py
scarabeusiv: At least in openSUSE we have provides, path requires should be last resort no?

SpecCheck.py
scarabeusiv: Nope that is some freaky old macro that is direct translation to /sbin/ldconfig

SpecCheck.py
scarabeusiv: Are those scriptlets in default rpm so it could be recommended globaly?

NamingPolicyCheck.py
scarabeusiv: not easily as it expect as first argument unique name for each check, and I kept the 1 there just for the fun of it as it can read the regular python- which were used before as python2 (and arguably even python1 arguments).

TagsCheck.py
Conan-Kudo: It depends. At least in this case, RPM itself generates `Requires: /usr/bin/pkg-config` with the pkgconfig dep generator for this reason.

It's generally consistent to just request the binary, since the package name does in fact differ across distros.

SpecCheck.py
Conan-Kudo: Not yet, no. It's being looked into, though.

FilesCheck.py
scarabeusiv: It is magical check making sure there sctually is executable and it aint empty, the if len(paths) > 1

Conan-Kudo: Ah.

BinariesCheck.py
```diff
@@ -321,6 +321,7 @@ def __init__(self, pkg, path, fname, is_ar, is_shlib):
 srcname_regex = re.compile(r'(.*?)-[0-9]')
 invalid_dir_ref_regex = re.compile(r'/(home|tmp)(\W|$)')
 ocaml_mixed_regex = re.compile(r'^Caml1999X0\d\d$')
+usr_arch_share_regex = re.compile(r'/share/.*/(?:x86|i.86|x86_64|ppc|ppc64|s390|s390x|ia64|m68k|arm|aarch64|mips|riscv)')
```
Conan-Kudo: mips can 32-bit or 64-bit, riscv can be 32-bit, 64-bit, or 128-bit.

TagsCheck.py
scarabeusiv: Yeah but both fedora and openSUSE state that it should be either BR: pkgconfig or BR: pkg-config. So we should use what's in those wikies

BinariesCheck.py
scarabeusiv: Wonder if it is worth it, basically it is relevant just for power and s390 all the others are sane and do not use /usr/share/archstuff/

MenuXDGCheck.py
```diff
@@ -8,10 +8,7 @@
 
 import codecs
 import os
-try:
-    import ConfigParser as cfgparser
-except ImportError:
-    import configparser as cfgparser
+import configparser as cfgparser
```
dirkmueller: While openSUSE only runs rpmlint via python 3, upstream still supports python 2.x.. I would prefer if we'd still keep that for now.

SpecCheck.py
```diff
@@ -9,10 +9,7 @@
 
 import re
 import sys
-try:
-    from urlparse import urlparse
-except ImportError:  # Python 3
-    from urllib.parse import urlparse
```
dirkmueller: Same here

SpecCheck.py
```diff
@@ -444,6 +445,10 @@ def check_spec(self, pkg, spec_file, spec_lines=None):
                                          'comparison-operator-in-deptoken',
                                          conf)
 
+            if current_section in ('post', 'postun'):
+                if lconfig_regex.search(line):
```
dirkmueller: What's the point of up streaming this patch? The macro isn't defined anymore, we should just drop it

rpmlint
```diff
@@ -17,12 +17,7 @@ import re
 import stat
 import sys
 import tempfile
-
-try:
-    import importlib
-except ImportError:  # Python < 2.7
-    importlib = None
-    import imp
+import importlib
```
dirkmueller: Same here

test.sh
```diff
@@ -2,7 +2,7 @@
 
 export PYTHONPATH=$(pwd)/tools:$(pwd)
 export TESTPATH="$(pwd)/test/"
-: ${PYTHON:=python} ${PYTEST:=py.test} ${FLAKE8:=flake8}
+: ${PYTHON:=python3} ${PYTEST:=py.test} ${FLAKE8:=flake8}
```
dirkmueller: Nope, still want to keep python 2 for now ;-)

dirkmueller:
@scarabeusiv as a hint, patches that start with "suse-" are downstream only and should be removed from this PR.. also it would be nice to have a little bit more descriptive commit message (+tests) for the changes that are being proposed.

SpecCheck.py
scarabeusiv: That macro is defined in 24 packages in tumbleweed.

scarabeusiv:
Oukey then this patchset does not make sense, nor my paralelization effort as that would work on python3 only. Well there goes my hackweek, I will just do something else.

dirkmueller:
@scarabeusiv well, I do appreciate the effort trying to upstream the opensuse packages.. I'm working on getting travis back to green (see #138 ) and then this could be revisited imho.

Conan-Kudo:
@dirkmueller @scarabeusiv I want to drop Python 2 support for rpmlint. Today, no users of rpmlint on the current release are using Python 2.

Conan-Kudo:
I am also okay with a number of these changes, I only have objections to a few specific ones, as I already detailed.

MenuXDGCheck.py
Conan-Kudo: I'd like to drop Python 2 support. Now is as good of a time as any, as there are no users of the current release with Python 2, and a number of distributions are simply not including Python 2 anymore if they can help it (RHEL 8, SLE 15, OMV Lx 4), while other distributions are working to kill off Python 2 in the next couple of years (Fedora, Mageia, etc.).

dirkmueller:
@Conan-Kudo I still need python 2.x compatibility until the next release, and I do think that supporting the latest community and enterprise OS is a good goal to have (and rhel7 still defaults to python 2.x, and so does SLE12). I've just reenabled the travis-ci checks and made them passing, but this pull request is breaking them again, so I wouldn't want to merge it as is. 

How about we continue the review in #140  instead.

Conan-Kudo:
@dirkmueller SLE 15 is Py3 only, and I don't really care about EL7 because they're frozen at rpmlint 1.5.

Keeping Python 2 compatibility means we need to use `trololio` for improving the performance of rpmlint, rather than leveraging asyncio directly.

Conan-Kudo:
RHEL 7 has Python 3.4 and 3.6 in EPEL, and SLE 12 has Python 3.4 in the distro itself. SLE 15 only has Python 3.6. Whenever RHEL 8 becomes a thing, it'll ship with only Python 3.

Conan-Kudo:
@scarabeusiv I merged some of this as #140, so if you'd like to rebase this PR on top, be my guest.

TagsCheck.py
Conan-Kudo: We don't actually have anything that mandates `BR: pkgconfig` in Fedora, because we include it in our buildroot.

FilesCheck.py
Conan-Kudo: I'd prefer to see the `exe.startswith("rc")` thing dropped, as it's SUSE-specific.

NamingPolicyCheck.py
Conan-Kudo: Perhaps we should leave this off the table for now, since it's a bit ugly and not easy to future-proof right now...

Conan-Kudo:
Looks great to me, merging!

PR #139, opened 2018-07-10
Add openSUSE Leap 42.3 travis testing

dirkmueller:
This is one of the older openSUSE releases that can be used
for regression testing against old rpm / python releases.


Conan-Kudo:
First, it doesn't look like Travis is working.

Second, is there a reasonable expectation that new rpmlint code would be deployed on Leap 42.3? From what I can tell, the answer to that is no.

Conan-Kudo:
I'd rather see openSUSE Leap 15.0 tested rather than Leap 42.3, because 15.0 actually _uses_ current rpmlint code.

dirkmueller:
this no longer makes sense for master, reopening against rpmlint-1.x branch

PR #142, opened 2018-07-11
Setuptools for distribution

scarabeusiv:
This makes the package behave as normal python software.

We install the binaries and all the data as needed. I verified the rpmlint itself behaves as expected, what I didn't do is to test importing of additional tests (Ie loading of the openSUSE/rpmlint-checks).

Questionable is where to put the default config file, atm I just put it to /usr/share/rpmlint...

Also what would be nice is to remove the scripts (rpmlint, rpmdiff) and just use entry points whlie having the normal classes.


Conan-Kudo:
This looks amazing! I'm looking over it now.

INSTALL
```diff
@@ -1,19 +0,0 @@
-You need the following utilities to run rpmlint:
```
dirkmueller: This information of non-python dependencies is useful to preserve. can you maybe add it to README.md if you want to get rid of INSTALL for some reason? otherwise please keep the file

setup.py
```diff
@@ -0,0 +1,55 @@
...
+"""Setup file for easy installation."""
+import glob
+
+from rpmlint import __version__
+from setuptools import setup
+
+setup(
+    name='rpmlint',
+    description='RPM file QA correctness validator',
+    long_description='Command-line tool for RPM files QA validation',
```
dirkmueller: Rpmlint is a tool to check common errors in RPMpackages.
Binary and source packages can be checked.

setup.py
```diff
@@ -0,0 +1,55 @@
...
+    url='https://github.com/rpm-software-management/rpmlint',
+    download_url='https://github.com/rpm-software-management/rpmlint',
```
dirkmueller: that one should poiint to the releases dir..

setup.py
```diff
@@ -0,0 +1,55 @@
...
+
+    version=__version__,
+
+    author='Frédéric Lepied',
+    author_email='flepied@mandriva.com',
+
+    maintainer='Neal Gompa',
+    maintainer_email='ngompa13@gmail.com',
```
dirkmueller: really? thats news to me.. :)

Conan-Kudo: Oi!

setup.py
```diff
@@ -0,0 +1,55 @@
...
+
+    license='License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)',
+    classifiers=[
+        # complete classifier list:
+        # http://pypi.python.org/pypi?%3Aaction=list_classifiers
+        'Development Status :: 5 - Production/Stable',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)',
+        'Operating System :: Unix',
+        'Operating System :: POSIX',
```
dirkmueller: Operating System :: POSIX :: Linux

setup.py
```diff
@@ -0,0 +1,55 @@
...
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 3',
```
dirkmueller: Intentionally leaving out the Python :: 2.7 marker? ;)

setup.py
```diff
@@ -0,0 +1,55 @@
...
+        'Topic :: Utilities',
+        'Topic :: Text Processing',
```
dirkmueller: Text processing?

how about 

Topic :: Software Development :: Quality Assurance
Topic :: System :: Archiving :: Packaging

test.sh
```diff
@@ -1,64 +0,0 @@
...
-echo "$PYTEST tests"
-$PYTEST -v || exit $?
-
-unset PYTHONWARNINGS
-
-echo "$FLAKE8 tests"
-$FLAKE8 --version
-$FLAKE8 . ./rpmdiff ./rpmlint || exit $?
-
-echo "man page tests"
```
dirkmueller: that part isn't converted anywhere? can the test coverage be restored?

test.sh
```diff
@@ -1,64 +0,0 @@
...
-    RET=$?
-    if [ $RET -ne 0 ]; then
-        exit $RET
-    fi
-done
-
-run_rpmlint="$PYTHON ./rpmlint -C $(pwd)"
-
-echo "Check that rpmlint executes with no unexpected errors"
-echo "...in default locale"
```
dirkmueller: Where is that being translated to?

test.sh
```diff
@@ -1,64 +0,0 @@
...
-$run_rpmlint test/*/*.rpm test/spec/*.spec >/dev/null
-rc=$?
-test $rc -eq 0 -o $rc -eq 64 || exit $rc
-echo "...in the C locale"
```
dirkmueller: And this?

test.sh
```diff
@@ -1,64 +0,0 @@
...
-LC_ALL=C $run_rpmlint test/*/*.rpm test/spec/*.spec >/dev/null
-rc=$?
-test $rc -eq 0 -o $rc -eq 64 || exit $rc
-echo "...with specfile from stdin"
-$run_rpmlint - < $TESTPATH/spec/SpecCheck.spec >/dev/null
```
dirkmueller: And this?

test/Dockerfile-centos7
```diff
@@ -1,5 +1,6 @@
 FROM centos:7
 
+RUN yum -y install epel-release
```
dirkmueller: its not possible to run this without EPEL?

test/Dockerfile-centos7
```diff
@@ -9,11 +10,16 @@ RUN yum -y install \
...
         /usr/bin/readelf \
         /usr/bin/xz \
+        python-pip \
         python-enchant \
-        python-magic
+        python-magic \
+        python-rpm
+
+# workaround of too old setuptools/pytest
+RUN pip install setuptools --upgrade
```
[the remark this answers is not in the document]
scarabeusiv: pyflake8 is not on the distribution, while pyflake8 in pypi needs new pytest new pytest needs new python-setuptools-scm and that needs new setuptools.

setup.py
scarabeusiv: The 3 is to be used if it works with any 3. you are supposed to diferentiate only if you can reasonably say it works only with ie 3.5+

test/Dockerfile-centos7
scarabeusiv: Missing python dependencies.

scarabeusiv:
Now the only pending request is to run the scripts themselves to see if it fails or not...

But the problem there is that I can't just launch any script. If it would be entry points I would load that point but that would take even longer (also now just calling the binary will not work of course as it needs the module properly available)...

Blindly doing something bellow simply won't work.
```
from subprocess import PIPE, Popen
def test_start():
    process = Popen('scripts/rpmlint.py', stdout=PIPE, stderr=PIPE, shell=False)
    out, err = process.communicate()
    print(out)
    print(err)
    assert not process.returncode
```

setup.py
scarabeusiv: So whom should I set there?

setup.py
scarabeusiv: Added

Conan-Kudo:
@scarabeusiv Please rebase this PR, as I merged #136. :)

test.sh
scarabeusiv: It is covered in test_mans.py now

setup.cfg
```diff
@@ -1,4 +1,13 @@
...
+[aliases]
+test=pytest
+
 [flake8]
 ignore = E122,E501,H101,H105,H201,H301,H404,H405,B001,W504
 import-order-style = google
 application-import-names = __isocodes__,__version__,AbstractCheck,AppDataCheck,BinariesCheck,Config,ConfigCheck,DistributionCheck,DocFilesCheck,FHSCheck,FilesCheck,Filter,I18NCheck,InitScriptCheck,LSBCheck,MenuCheck,MenuXDGCheck,NamingPolicyCheck,PamCheck,Pkg,PostCheck,RpmFileCheck,SCLCheck,SignatureCheck,SourceCheck,SpecCheck,TagsCheck,ZipCheck,Testing
+
+[tool:pytest]
+addopts = --flake8 --cov=rpmlint
```
dirkmueller: can we run flake8 separately, via tox perhaps? having flake8 as a test dependency is inconvenient for building the package in a distribution (pulls in all the flake dependencies).

test/Dockerfile-centos7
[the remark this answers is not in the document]
dirkmueller: I see, thats why there was a FLAKE8=true before. I would still prefer to test rpmlint against the *actual* distribution packages not some epel, and then split out flake8 testing into a separate target (doesn't have to be part of the testsuite imho), we can add it as a separate .travis.yml only job for example.

scarabeusiv:
Found a bug, the rpmlint worked for me as i still had old content in /usr/share...

Will have to adjust the lazy loading so it still keeps working as for now it will fail with:

```
scarabeus@bugaboo: ~/remote/rpmlint $ rpmlint test/*/*.rpm test/spec/*.spec
Traceback (most recent call last):
  File "/usr/bin/rpmlint", line 372, in <module>
    main()
  File "/usr/bin/rpmlint", line 95, in main
    loadCheck(c)
  File "/usr/bin/rpmlint", line 76, in loadCheck
    importlib.import_module(name)
  File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/lib/python3.6/site-packages/rpmlint/DistributionCheck.py", line 14, in <module>
    from . import AbstractCheck
ImportError: attempted relative import with no known parent package

```

scarabeusiv:
I had to rewrite the tests in this pull because it would not work at all otherwise and I think you would not be willing to merge the setuptools change with having tests failing.

setup.cfg
scarabeusiv: The distribution can still override the pytest and disable the flake8 run if they desire so. Yet it should not cause issues for us, and this way we don't need no tox to be present in the enviroment.

PR #145, opened 2018-07-12
Add openSUSE Leap 42.3 travis testing

dirkmueller:
This is one of the older openSUSE releases that can be used
for regression testing against old rpm / python releases.


scarabeusiv:
As we no longer support Leap 42.3 I suppose it does not make sense anymore right?

PR #151, opened 2018-07-14
rpmlint configuration parsing in a object

scarabeusiv:
This is slowly being done conversrion to config reading in an object file.
Everything is now stored in buffer and can be sorted/etc prior printout.
Also all list of list formats were redone to dictionaries.


test/test_helpers.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from rpmlint import helpers
+
+def test_warnprint(capsys):
+    """
+    Check we print stuff to stderr
+    """
+    message = 'I am writting to stderr'
```
shlomif: see https://en.wiktionary.org/wiki/writing

rpmlint/BinariesCheck.py
```diff
@@ -334,8 +324,13 @@ def dir_base(path):
 
 class BinariesCheck(AbstractCheck):
 
-    def __init__(self):
-        AbstractCheck.__init__(self, 'BinariesCheck')
+    def __init__(self, config):
+        AbstractCheck.__init__(self, config, 'BinariesCheck')
```
mimi1vx: use `super` to call parent object instance

rpmlint/DistributionCheck.py
```diff
@@ -15,41 +14,35 @@
...
-distribution = Config.getOption("Distribution")
-compress_ext = Config.getOption("CompressExtension", "bz2")
-
-
 class DistributionCheck(AbstractCheck):
 
-    def __init__(self):
-        AbstractCheck.__init__(self, "DistributionCheck")
+    def __init__(self, config):
+        AbstractCheck.__init__(self, config, "DistributionCheck")
```
mimi1vx: dtto

.travis.yml
```diff
@@ -7,7 +7,7 @@ services:
 
 env:
   - DIST=fedora28 PYTHON=python3 FLAKE='/usr/bin/flake8'
-  - DIST=fedoradev PYTHON=python3 FLAKE='/usr/bin/flake8'
+#  - DIST=fedoradev PYTHON=python3 FLAKE='/usr/bin/flake8'
```
Conan-Kudo: Why did this break?

scarabeusiv: It is broken even on master, the docker image will not start installing stuff, some conflicts in the rawhide files. We can enable it again when it is fixed by Fedora team.

dirkmueller:
Is there some way to break out some changes into a pr to make this one smaller?

scarabeusiv:
Not really sadly.
The configuration change was trigger to ensure the sanity of the imports, because it broke for me by just changing one import order. That change on its own won't work anywhere.
Then you need the filtering object to make the above work.
And in the end you need to make all objects working with the rework of those two.

I can remove noop commits for this but even then it is huge:
1487e3f
3ab46dd
0b4f87f
feab2a4

All in all we should not do any release prior issue #156 is covered. So mostly focus on the major logic changes and if something syntax wise really does not make sense.

.travis.yml
Conan-Kudo: Okay, then.

Conan-Kudo:
@scarabeusiv Merged, thanks!

PR #153, opened 2018-07-19
Treat Sphinx inventory files as binary

elyscape:
Sphinx, a tool commonly used to produce documentation for Python projects, produces inventory files with extension `.inv` that contain a plaintext header followed by raw zlib data. When the zlib data is short enough, this fools [the heuristic used][heuristic] to determine if a file is text or binary. Since these files are often included in documentation packages, this can produce false positives for the `wrong-file-end-of-line-encoding` check and the subsequent `file-not-utf8` check.

[heuristic]: https://github.com/rpm-software-management/rpmlint/blob/c1dbe704c42749abd1b2f3abec9fdf374c0ef986/rpmlint/FilesCheck.py#L280-L313

### DIFF
--- a/rpmlint/FilesCheck.py
+++ b/rpmlint/FilesCheck.py
@@ -304,6 +304,9 @@ def peek(filename, pkg, length=1024):
     # Ditto RDoc RI files
     if fl.endswith('.ri') and '/ri/' in fl:
         return (chunk, False)
+    # And Sphinx inventory files
+    if fl.endswith('.inv') and chunk.startsWith('# Sphinx inventory'):
+        return (chunk, False)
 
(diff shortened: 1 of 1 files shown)


elyscape:
The Travis failure is caused by the `fedora:rawhide` Docker image being broken and containing a number of packages built for F28. [Bugzilla #1601479][bugzilla] tracks this issue, and I'm told that 

[bugzilla]: https://bugzilla.redhat.com/show_bug.cgi?id=1601479

scarabeusiv:
@elyscape could you give me some small rpm that is exhibiting the issue so we could add a testcase for it?

scarabeusiv:
Replaced by #269

PR #160, opened 2018-07-31
Test covering for ZipCheck.py

scarabeusiv:
Added tests validating ZipCheck.py

Note that I also redid the ZipCheck to be function based. Here be picky on the changes and think if we should drop something or do stuff differently.


Conan-Kudo:
#151 is merged, please rebase.

Conan-Kudo:
LGTM, thanks!

PR #161, opened 2018-08-01
Add more flake8 checks

scarabeusiv:
Add more of the automatic flake8 checks and fix most of the offenders.

N802 and N806 are hidden for now as there are many issues in lint diff and pkg, but all 3 of those need to go via review and redesign and as such it makes sense to fix that when doing that action rather than here.


test/Dockerfile-tumbleweed
```diff
@@ -15,6 +15,8 @@ RUN zypper -n install \
         python3-pytest \
         python3-flake8 \
         python3-flake8-import-order \
+	python3-flake8-bugbear \
+	python3-flake8-quotes \
```
junaruga: The indent could be improved?

PR #168, opened 2018-08-17
Separate spellchecking object

scarabeusiv:
Simply rely on the enchant instead of inventing our own spellcheck.

I should do followup where it is properly integrated and has a configure option wether to spellcheck or not.

### DIFF
--- a/rpmlint/checks/TagsCheck.py
+++ b/rpmlint/checks/TagsCheck.py
@@ -12,281 +12,14 @@ import re
 import time
 from urllib.parse import urlparse
 
+import enchant
+import enchant.checker
 import rpm
 from rpmlint import Pkg
 from rpmlint.checks import FilesCheck
 from rpmlint.checks.AbstractCheck import AbstractCheck, macro_regex
 from rpmlint.helpers import byte_to_string
 
-
-_use_enchant = True  # FIXME: kill the option always enchant
-if _use_enchant or _use_enchant is None:
-    try:
-        import enchant
-        import enchant.checker
-    except ImportError:
-        enchant = None
-else:
-    enchant = None
-del _use_enchant
-
-BAD_WORDS = {
-    'alot': 'a lot',
-    'accesnt': 'accent',
-    'accelleration': 'acceleration',
-    'accessable': 'accessible',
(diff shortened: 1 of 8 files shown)


rpmlint/checks/TagsCheck.py
```diff
@@ -12,281 +12,14 @@
 import time
 from urllib.parse import urlparse
 
+import enchant
```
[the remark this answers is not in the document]
scarabeusiv: Yea I will create config option to turn it on/off, but without magical alternative spellchecking module simply to have it on/off

rpmlint/checks/TagsCheck.py
```diff
@@ -18,275 +18,14 @@
 from rpmlint.checks.AbstractCheck import AbstractCheck, macro_regex
 from rpmlint.helpers import byte_to_string
 
+try:
+    import enchant
+    import enchant.checker
+except ImportError:
+    # if the enchant is not present we simply continue byt without
```
Conan-Kudo: `byt` -> `but`

scarabeusiv:
You know what, I will split it to separate spellcheck helper and make sure it is tested, it seems like on openSUSE it didn't really work.

rpmlint/configspec.cfg
```diff
@@ -42,6 +42,8 @@ ReleaseExtension = string(default='')
 RpmGamesGroup = string(default='Games')
 # Doc files to which end of line and UTF-8 checks should not be applied
 SkipDocsRegexp = string(default='\.(?:rtf|x?html?|svg|ml[ily]?)$')
+# Enable spell checking
+SpellCheck = boolean(default=True)
 # Whether to use the Enchant spell checker for spell checking
 UseEnchant = boolean(default=True)
```
dirkmueller: isn't that redundant now? I would prefer to keep the existing setting rather than adding a new one that has the same behavior..

scarabeusiv:
Now it should work, only question stands if it will work in docker. :)

scarabeusiv:
Okay the TW breakage is bug in pyenchant, will do update there.

Conan-Kudo:
Travis said it was all good, so I merged. Thanks!

PR #177, opened 2018-09-19
Add gzip to deps on Leap 15 docker container

scarabeusiv:


scarabeusiv:
Otherwise the CI would fail now.

PR #181, opened 2018-09-27
Small tweaks to the container setup for CI

Conan-Kudo:


scarabeusiv:
Interesting fedoradev flake reports more stuff than tumbleweed one, yet they are on the same version...

Conan-Kudo:
@scarabeusiv I think the difference is that Fedora Rawhide is already on Python 3.7, which means we have a new Deprecation to deal with.

scarabeusiv:
Hmm looking at the fail it seems bogus actually :(

scarabeusiv:
Well since the same thing is exposed on latest TW lets merge this and figure it out.

PR #185, opened 2018-11-16
Fix exception in libtool handling and make testsuite work

dirkmueller:


scarabeusiv:
The fail is either bug in flake8 or something along the way, anyway the change here looks correct so lets merge it. :)

PR #187, opened 2018-11-16
Convert lint from script to normal object

scarabeusiv:
Instead of script let us have a lint object.


scarabeusiv:
So except of rpm-4.15 breakages all seems fine. Note the fails from the rpm are even before this PR just not exposed.

scarabeusiv:
@dirkmueller @mimi1vx @hroncok guys if you find time could you look up to check for some obvious errors? Otherwise this is probably ready.

scarabeusiv:
As it passes all the check I am going to merge this. If you guys spot something just file up bugs and we will see what happens.

Issue #179, opened 2018-09-19
DocFilesCheck.py - GPL2.0 only


scarabeusiv:
We have whole package licensed as GPL-2.0+ yet DocFilesCheck states itself as gpl-2.0-only and there is no way to contact the author of it (dead emails everywhere).

As such someone should rewrite it in order to have the whole rpmlint under one license only.

dirkmueller:
I found a working email address and asked for relicensing.

ensc:
I am ok with relicensing to GPL-2.0+.

scarabeusiv:
Splendid, thank you!

Conan-Kudo:
This was fixed a while ago...

PR #189, opened 2018-11-16
Remove gpl2.0-only blurb as per issue #179

scarabeusiv:
Authr acknowledged in the ticket that the file can use the general
license under which the package is distributed [GPL-2.0-or-later].


Conan-Kudo:
@scarabeusiv Please put the license header back, since his authorship is still distinct, just now GPLv2+ instead of GPLv2.

Conan-Kudo:
Gah, I had missed this originally. Oh, well, I'll just make an issue for adding a generic header to all files.

PR #195, opened 2018-12-20
Add /usr/lib*/firmware to UsrLibBinaryException

yaneti:


yaneti:
Hmm, I wonder if we should use UsrLibBinaryException to also avoid arch-independent-package-contains-binary-or-object

Conan-Kudo:
Merged, thanks!

PR #196, opened 2019-01-02
Backport d8f423b575e8be387d33bc3af176baf978efacbb:

marxin:
Come up with lto-bytecode check for ELF files.


BinariesCheck.py
```diff
@@ -484,6 +486,9 @@ def check_binary(self, pkg):
                 for ec in bin_info.exit_calls:
                     printWarning(pkg, 'shared-lib-calls-exit', fname, ec)
 
+            if bin_info.lto_sections:
+                printWarning('E', pkg, 'lto-bytecode', fname)
```
scarabeusiv: Shouldn't this be ...printError(pkg...

marxin: Yep, fixed.

BinariesCheck.py
```diff
@@ -484,6 +486,9 @@ def check_binary(self, pkg):
                 for ec in bin_info.exit_calls:
                     printWarning(pkg, 'shared-lib-calls-exit', fname, ec)
 
+            if bin_info.lto_sections:
+                printError('E', pkg, 'lto-bytecode', fname)
```
dirkmueller: this doesn't look correct, the proper usage would be:

```
printError(pkg, 'lto-bytecode', fname)
```

BinariesCheck.py
```diff
@@ -773,7 +778,11 @@ def check_binary(self, pkg):
 '''This executable should be stripped from debugging symbols, in order to take
 less space and be loaded faster. This is usually done automatically at
 buildtime by rpm. Check the build logs and the permission on the file (some
-implementations only strip if the permission is 0755).'''
+implementations only strip if the permission is 0755).''',
+
+'lto-bytecode',
+'''This executable contains a LTO section.  LTO bytecode is not portable
+and should not be distributes in static libraries or e.g. Python modules.''',
```
dirkmueller: typo: distributed

marxin: Fixed.

BinariesCheck.py
marxin: Fixed.

marxin:
> testcase missing - and how is that supposed to work?

If I'm correct @scarabeusiv told me the this branch does not support tests of BinariesCheck.py.

the error only says "should not be used in python modules" but I see no implementation of "only python modules" as a code check. what ensures that the warning does not show up in cases where it is safe? and when is it safe at all?

Full error message is:
```
his executable contains a LTO section.  LTO bytecode is not portable
and should not be distributed in static libraries or e.g. Python modules.
```

So we should never distribute these ELF sections in rpms.

dirkmueller:
@marxin there is tests/test_binaries.py which is testing this actually. I am wondering though, is this "bytecode" actually the result of -flto? can you add instructions on what to do about this error? e.g. build without -flto? build link step with lto? what is the actual issue, do we know? I would assume the lto bytecode is being removed in teh final link step?

marxin:
> @marxin there is tests/test_binaries.py which is testing this actually.
Ok, then let me add a test.
> I am wondering though, is this "bytecode" actually the result of -flto? can you add instructions on what to do about this error? e.g. build without -flto? build link step with lto? what is the actual issue, do we know? I would assume the lto bytecode is being removed in teh final link step?

All these are valid question. In general exposure of LTO bytecode happens when one distributes objects files (.a archive made from object files) and one should ideally build so called fat LTO objects (containing both LTO bytecode and assembly). And then LTO bytecode should be stripped.

Note that I'm first targeting to enable LTO in a staging project that can be a potential candidate for submission to Factory. In that time I'm planning to write a wiki section that will describe what to do with LTO and troubleshooting will be also included. In that time we can extent rpmlint to provide more information.

marxin:
Ok, now I added test-case and fixed the check. Should be fine now.

PR #214, opened 2019-07-12
Use docker volumes instead of copying data

scarabeusiv:
This could allow us to have all the results in travis and as such
ie. upload them to coveralls.

Do not ack this, I will try to get the coveralls uploads first completely working.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,13 @@
+[run]
+source = .
+
+; Run in parallel mode so that coverage can canonicalize the source paths
+; regardless of whether it runs locally or within a Docker container.
+parallel = True
+
+[paths]
+; the first path is the path on the local filesystem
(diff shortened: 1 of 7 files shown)


junaruga:
I tried this update for my app's repository.
I faced the volume directory's files are updated as root on the host side directory.
Also podman has different behavior from docker.
See https://github.com/containers/libpod/issues/3432 for detail.

scarabeusiv:
For us we should be fine we care just to grab the .coverage content and publish it, nothing more. Hopefully my last push made it play friendly.

scarabeusiv:
Ok now it works, but we need to grant reading to coveralls.io to actually be able to publish. @Conan-Kudo any ideas whom to poke? It needs to be owner of the group...

Conan-Kudo:
@scarabeusiv That would be @dmach.

PR #215, opened 2019-07-15
Fix coveralls reporting

scarabeusiv:
For now all stuff is properly submitted but the coveralls report 0 coverage.

As such trying to figure out why.


scarabeusiv:
And the relative vs absolute was to blame :)

PR #217, opened 2019-07-22
Switch from configobj to TOML

scarabeusiv:
The FIXMEs are quite plenty so this won't be ready for a bit :)

Most annoying part will be probably validation, as we should make sure that the regexps are okay/etc when reading the config to avoid crashing later on. On the bright side toml handles much better dictionaries/lists which we are using quite a lot.


rpmlint/config.py
```diff
@@ -88,108 +65,57 @@ def set_badness(self, result, badness):
...
                 # load all configs in the folders
                 confopts = sorted(confdir.glob('*config'))
                 self.conf_files += confopts
 
+        # As a last item load up the user configuration
+        if config:
+            if config.exists():
+                # load this only if it really exist
+                self.conf_files.append(config)
+                return
```
mimi1vx: why `return` here ?

scarabeusiv: Removed :)

PR #236, opened 2019-07-31
WIP: rewrite another part of BinariesCheck.

marxin:
This is work in progress.
@scarabeusiv: Can you please help me to mock up `test_soname_symlink` so that we'll see `invalid-ldconfig-symlink` and `no-ldconfig-symlink` errors?


marxin:
Good, I have a test-case for `no-ldconfig-symlink` for free now. So the other one will be needed.

rpmlint/checks/BinariesCheck.py
```diff
@@ -277,23 +279,82 @@ def check_lto_section(self, pkg, path):
...
             elif 'E' in stack_headers[0].flags:
                 self.output.add_info('E', pkg, 'executable-stack', path)
 
+    def check_shared_library(self, pkg, path):
+        if self.readelf_parser.is_shlib:
+            soname = self.readelf_parser.dynamic_section_info.soname
+            if not soname:
+                self.output.add_info('W', pkg, 'no-soname', path)
+            else:
+                if not BinariesCheck.validso_regex.search(soname):
```
scarabeusiv: self.validso_regex

rpmlint/checks/BinariesCheck.py
```diff
@@ -277,23 +279,82 @@ def check_lto_section(self, pkg, path):
...
+                    self.output.add_info('E', pkg, 'invalid-soname', path, soname)
+                else:
+                    (directory, base) = dir_base(path)
```
scarabeusiv: This is getting too nested, another function?

marxin: I reduced the level, I'll split it later.

rpmlint/pkg.py
```diff
@@ -934,12 +934,14 @@ def checkSignature(self):
...
 # Class to provide an API to a 'fake' package, eg. for specfile-only checks
 class FakePkg(AbstractPkg):
-    def __init__(self, name):
+    def __init__(self, name, files=None):
         self.name = str(name)
         self.arch = None
         self.current_linenum = None
         self.dirname = None
 
+        self._files = {f.name: f for f in files} if files else {}
```
scarabeusiv: please comment out why is this added

rpmlint/pkg.py
```diff
@@ -950,6 +952,9 @@ def cleanup(self):
         if self.dirname:
             getstatusoutput(('rm', '-rf', self.dirname))
 
+    def files(self):
+        return self._files
+
```
scarabeusiv: why not access the variable directly?

rpmlint/checks/BinariesCheck.py
```diff
@@ -277,26 +271,70 @@ def check_lto_section(self, pkg, path):
...
+        if not soname:
+            self.output.add_info('W', pkg, 'no-soname', path)
+        else:
+            if not self.validso_regex.search(soname):
+                self.output.add_info('E', pkg, 'invalid-soname', path, soname)
+            else:
+                (directory, base) = dir_base(path)
+                symlink = directory + soname
+                try:
+                    link = pkg.files()[symlink].linkto
```
scarabeusiv: please describe this in some comment to have grasp on what goes on, also would be nice as separate function

rpmlint/readelfparser.py
```diff
@@ -132,22 +148,101 @@ def parse(self):
...
+        for line in lines:
+            r = self.section_regex.search(line)
+            self.sections.append(ElfDynamicSection(r.group('key'), r.group('value')))
+
+    def parse_meta(self):
+        self.soname = None
+        soname = self['SONAME']
+        if len(soname) == 1:
+            value = soname[0]
+            token = 'Library soname: ['
```
scarabeusiv: seriously the assert does not work for you this way

test/test_readelf_parser.py
```diff
@@ -75,3 +91,39 @@ def test_executable_stack(binariescheck):
 def test_readelf_failure():
     readelf = readelfparser('no-existing-file')
     assert readelf.parsing_failed
+
```
scarabeusiv: is the libao. rpm still needed?

rpmlint/readelfparser.py
marxin: Yes, I know. I'm fine with the **soft** assert.

rpmlint/pkg.py
marxin: Because I want to follow the same interface as in `class Pkg`, where you also use `pkg.files()['my_file']`.

marxin:
@scarabeusiv I'll ping you once I'm ready for a review from you. Now it's WIP.

rpmlint/pkg.py
scarabeusiv: It is not really needed pkg.files['whatever'] works just fine,the pkg.py still needs these cleanups really.

scarabeusiv: Oth it can stay and be fixed in one huge chunk, okay...

PR #246, opened 2019-08-02
Drop online checks

scarabeusiv:


Conan-Kudo:
@scarabeusiv Why are we killing the optional network-based checks?

Conan-Kudo:
@scarabeusiv While SUSE distributions can't use these checks because of OBS, I'm pretty sure everyone else actually relies on them as part of the lint checking.

scarabeusiv:
nobody else uses the rpmlint coherently enough and the current implementation is quite sucky.

the https is handled by the spec-cleaner and the speccheck test if anything should be something like sourcestest and actually done properly like source validation service in obs.

Conan-Kudo:
@scarabeusiv No other distribution using rpmlint uses OBS. Checks in ROSA and OMV's ABF instances, for example, rely on rpmlint. Fedora's `fedora-review` uses rpmlint as part of the review process and leverages these checks. ALT Linux and PLD Linux also use rpmlint in similar manners.

OBS usage of rpmlint only works using a perverted build of rpmlint with a weird configuration and install path anyway.

scarabeusiv:
Again the https check is better offloaded to spec-cleaner because why warn about something when you can automatically fix it.
The two checks that are extra in speccheck are not entirely correct and should be done differently. I can re-create them properly if they are really used by some review process...

Conan-Kudo:
@scarabeusiv If spec-cleaner wasn't so fully SUSE-specific, I'd consider that argument.

But yes, they are used by some people for their review processes. The Mandriva family (where rpmlint comes from!) relies on it pretty heavily.

scarabeusiv:
I have no problem in merging support of other distros in spec-cleaner. Nobody ever bothered creating PRs. And it ain't like I didn't promote it.

Conan-Kudo:
> I have no problem in merging support of other distros in spec-cleaner. Nobody ever bothered creating PRs. And it ain't like I didn't promote it.

Well, actually I never heard of spec-cleaner until I got involved in openSUSE a few years ago... :(

scarabeusiv:
Ouky for the online md5 checks I will create new object that will really properly validate the stuff (not using md5 :P) and it will automatically detects if it can go online and do the check if it can..

Conan-Kudo:
@scarabeusiv I would greatly appreciate it. Thanks!

scarabeusiv:
In we go with the #254 putting it back later.

PR #247, opened 2019-08-02
Descriptions in toml

scarabeusiv:


Conan-Kudo:
So why do we want to move the hard-coded descriptions to TOML? DO we want to do some kind of configuration or override mechanism?

scarabeusiv:
Readability (50% of the check files are the descriptions) and easier possibility to tweak up the message by the distros.

scarabeusiv:
Also I want to write a test that will make sure that all the options do actually have some sort of description (which right now is not the case).

Conan-Kudo:
What do we have to gain by making it easier for distros to make it confusing by changing the descriptions?

scarabeusiv:
@Conan-Kudo it is mostly for better readability of the code, the possibility of the override is just a bonus and it does not make it way easier than it is even now as part of the code.

Conan-Kudo:
@scarabeusiv Okay then, I'll buy that. Let's see how this works out.

rpmlint/checks/BinariesCheck.py
```diff
@@ -137,9 +135,9 @@ def _check_shared_library(self, pkg, pkgfile_path, path):
             # libs only, but that has potential to generate lots of
             # false positives and noise.
             for symbol in self.ldd_parser.undefined_symbols:
-                self.output.add_info('W', pkg, 'undefined-non-weak-symbol', path, symbol)
+                self.output.add_info('E', pkg, 'undefined-non-weak-symbol', path, symbol)
```
mimi1vx: why this change ?

scarabeusiv: Should've been error previously missed it in previous commit. Could've added it as separate commit.

rpmlint/checks/MenuXDGCheck.py
```diff
@@ -22,7 +22,6 @@ def __init__(self, config, output):
         # /var/lib/menu-xdg:/usr/share
         AbstractFilesCheck.__init__(
```
mimi1vx: `super().__init__(config`.....

rpmlint/filter.py
```diff
@@ -24,14 +28,24 @@ def __init__(self, config):
...
         # Messages
-        self.results = list()
+        self.results = []
+
+    @staticmethod
+    def _load_descriptions():
+        descriptions = {}
+        descr_folder = Path(__file__).parent / 'descriptions'
+        try:
+            description_files = sorted(descr_folder.glob('*.toml'))
```
mimi1vx: why `sorted` ?

scarabeusiv: There can be dictionaries overrides this way because if you simly create ZZZmyoverride it will work. Also overshadowing if ever happens will be deterministic.

Conan-Kudo:
@scarabeusiv Can you please give me a chance to review things before merging in the future? I know that our lack of timezone overlap is kind of painful (UTC-4 vs UTC+2), but not getting a chance to review the final form until after it's merged kind of stinks...

PR #251, opened 2019-08-09
Simplify section_regex regex and handle .preinit_array in .a files.

marxin:


marxin:
The limitation is seen for `llvm8` package in current `:N` staging project.

PR #257, opened 2019-08-14
Add new check: linked-against-opt-library

marxin:
The check should replace
https://github.com/openSUSE/brp-check-suse/blob/master/brp-40-rootfs
in the future.


rpmlint/checks/BinariesCheck.py
```diff
@@ -141,6 +142,13 @@ def _check_shared_library(self, pkg, pkgfile_path, path):
                 self.output.add_info('W', pkg, 'unused-direct-shlib-dependency',
                                      path, dependency)
 
+    def _check_opt_library_dependency(self, pkg, pkgfile_path, path):
+        if not self.readelf_parser.is_archive:
+            for dependency in self.ldd_parser.dependencies:
```
mimi1vx: maybe:
```if any(d.startswith("/opt/") for d in self.ldd_parser.dependencies):
    self.output.add_info('E', pkg, 'linked-against-opt-library', path, dependency)
```

marxin: Probably yes, it's just a cosmetic change.

Issue #113, opened 2017-09-14
rpmlint --explain useless-provides is terrible


bkircher:
```
$ rpmlint --explain useless-provides 
useless-provides:
This package provides 2 times the same capacity. It should only provide it
once.
```

This is terrible. Any suggestion on what "useless-provides" is actually supposed to prevent?

scop:
Duplicate Provides, for example `foo` and `foo = 1.0`. Feel free to suggest a better explanation, even better if in a form of pull request :)

Conan-Kudo:
It's not really "preventing" anything. It's informing you that you have duplicate declarations of the same base information.

I don't see anything wrong with it. I'm happy to consider PRs with better wording, though.

dirkmueller:
I think we should update the explanation (I think I have a patch for that somewhere)

Conan-Kudo:
@dirkmueller Alright then, looking forward to that. ;)

scarabeusiv:
Bit improved I hope... :)

hroncok:
The new message is not always correct. If the RPM has two provides that include version, the error is present as well:

```
$ rpm -qp --provides python3-double-provides-0-0.fc33.noarch.rpm 
python-double-provides = 0-0.fc33
python-double-provides = 0-0.fc33
python3-double-provides = 0-0.fc33
python38-double-provides = 0-0.fc33
python38-double-provides = 0-0.fc33

$ rpmlint python3-double-provides-0-0.fc33.noarch.rpm 
...
python3-double-provides.noarch: E: useless-provides python-double-provides
python3-double-provides.noarch: E: useless-provides python38-double-provides
...
```

"This means versioned and unversioned symbols are provided at once" is not applicable in this case. 

In this case, the situation is "Identical automated and manual provides exist, the manual one is useless"

scarabeusiv:
Yeah feel free to file it as a new ticket Miro, the TagsCheck is not yet reviewed and is kinda hard to process at the moment tho :)

hroncok:
Here: https://github.com/rpm-software-management/rpmlint/issues/427

PR #266, opened 2019-08-15
Update bit description of useless-provides issue #113

scarabeusiv:


scarabeusiv:
@dirkmueller is the above good enoughTM? or do you have some patch somewhere still?

scarabeusiv:
Otoh if dirk has better patch he can replace this one later on if needed (the ticket was open for >1 year... :)

Issue #225, opened 2019-07-26
Review and convert the test logic to use functionalized approach


scarabeusiv:
The tests for individual issues should be separated as much as possible to be later paralelized.

All the tests should be reviewed and verified they are in good shape.
Also with this logic we should have at least 1 testcase per the function to make sure we do not regress to much.

Here is a list of current tests to review (new ones will hopefully be reviewed prior addition and adhere to this 'requirement):

- [ ] AbstractCheck.py
- [ ] AppDataCheck.py
- [ ] BinariesCheck.py
- [ ] ConfigCheck.py
- [ ] DistributionCheck.py
- [ ] DocFilesCheck.py
- [ ] FHSCheck.py
- [ ] FilesCheck.py
- [ ] InitScriptCheck.py
- [ ] I18NCheck.py
- [ ] LSBCheck.py
- [ ] MenuCheck.py
- [ ] MenuXDGCheck.py
- [ ] NamingPolicyCheck.py
- [ ] PamCheck.py
- [ ] PostCheck.py
- [ ] RpmFileCheck.py
- [ ] SCLCheck.py
- [ ] SignatureCheck.py
- [ ] SourceCheck.py
- [ ] SpecCheck.py
- [ ] TagsCheck.py
- [ ] ZipCheck.py

marxin:
This seems to be a candidate for milestone `2.1`. I don't see it fundamental for `2.0` release.

kstreitova:
_(Just commenting on this so Tomas can assign this issue to me)_

Issue #156, opened 2018-07-24
All the objects we have in place should have some basic tests


scarabeusiv:
We don't have any basic testcase validating various of the files, I will add it here whenever I notice and then we should write down some basic checks for them to ensure sanity.

- [ ] ConfigCheck.py

scarabeusiv:
In signature check we have test that outputs whitespace garbage while it should not and should be part of the signature check:

```apache-commons-collections-testframework-javadoc.noarch: E: no-signature
You have to include your pgp or gpg signature in your package.             For
more information on signatures, please refer to www.gnupg.org.
```

scarabeusiv:
This issue is really tied to issue #225 and probably should go hand in hand with that effort.

scarabeusiv:
Also there is rpmlint-tests in openSUSE namespace that could be incorporated in the main project, effectively obsoleting the openSUSE repo.

Current list of tests that could be added as a checklist:

 - [ ] buildroot.ref 
 - [ ] cvs-bad.ref 
 - [ ] dbus-policy.ref 
 - [ ] debug.ref 
 - [ ] debug1.ref 
 - [ ] debug2.ref 
 - [ ] desktopfile-bad-binary.ref 
 - [ ] desktopfile-bad-duplicate.ref 
 - [ ] desktopfile-bad-section.ref 
 - [ ] desktopfile-bad-unicode.ref 
 - [ ] desktopfile-good.ref 
 - [ ] empty.ref 
 - [ ] erlang.ref 
 - [ ] foo1-developer-plugins.ref 
 - [ ] foo1-devel.ref 
 - [ ] game.ref 
 - [ ] gethostbyname.ref 
 - [ ] ghost1.ref 
 - [ ] ghost2.ref 
 - [ ] ghost3.ref 
 - [ ] good.ref 
 - [ ] chroot.ref 
 - [ ] libzork42.ref 
 - [ ] license1.ref 
 - [ ] logrotate1.ref 
 - [ ] logrotate2.ref 
 - [ ] noarch.ref 
 - [ ] opt.ref 
 - [ ] opt2.ref 
 - [ ] pam1.ref 
 - [ ] permissions1.ref 
 - [ ] permissions2.ref 
 - [ ] pie.ref 
 - [ ] polkit.ref 
 - [ ] rclink2.ref 
 - [ ] rclink4.ref 
 - [ ] rclink5.ref 
 - [ ] rclink6.ref 
 - [ ] shlib1.ref 
[...]

PR #270, opened 2019-08-18
Remove LSBCheck

kstreitova:
Separate the check() method to multiple functions and clean up the code. Also add a test for LSBCheck and adjust LSBCheck.toml description to be more precise (#225 #156).


rpmlint/checks/LSBCheck.py
```diff
@@ -18,15 +9,50 @@
```
scarabeusiv: The regexps could be class variable maybe, but regardless the a-zA-Z0-9... should be \w+ see the https://docs.python.org/3/howto/regex.html

kstreitova:
After a discussion with Tomas, I suggest removing this check. See reasons in my commit message.

scarabeusiv:
It is not enabled on any distro so LGTM. @Conan-Kudo your opinion?
Today it seems we delete shitaton of stuff ;-)

Conan-Kudo:
I can explain where the rules come from. The rules are the intersection of compatible NVRs between RPM v3 and DPKG for supporting portability across RPM and Debian systems through tools like `alien` and other things.

The check is disabled by default in Fedora because we have stopped enforcing this compatibility requirement, but there are corporate users that have this check turned on for their own RPMs.

kstreitova:
Then we should rename it as it's not connected to LSB at all. Any suggestions?

Are the regexps correct so I can adjust description according to it?

Conan-Kudo:
@kstreitova It is related to LSB, as LSB specified RPM v3 as the standardized package format.

PR #272, opened 2019-08-19
Remove 'filename-too-long-for-joliet' check

kstreitova:
This check is disabled in Fedora since 2009 [1] and for openSUSE it
fails for 145 packages so evidently nobody cares.

Also, Joliet is a quite old specification so I suppose that modern
tools either truncate or use something like 'joliet-long' extension
(allow Joliet filenames to be up to 103 Unicode characters) anyway.

[1] https://www.redhat.com/archives/fedora-packaging/2009-January/msg00042.html


scarabeusiv:
@Conan-Kudo what's your take. As Fedora disable this and the test after updates to the specification really does not make sense we could really live without it...

Conan-Kudo:
Blech, let's get rid of this.

PR #273, opened 2019-08-19
Remove 'use-old-pam-stack' check

kstreitova:
The purpose of this check was to validate whether pam files contain
deprecated 'pam_stack' module instead of 'include' directive.

Pam_stack module was deprecated in PAM 0.78 (2005)[1] so it was
enough time for the migration. Currently, there is no package
in openSUSE that would break this check.

[1] https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/U32WZEPTQBBCEPV6KJY35L2RZ7WKWTII/


scarabeusiv:
LGTM. @Conan-Kudo ^^ :)

Conan-Kudo:
I hope by now that nobody is using old PAM...

PR #275, opened 2019-08-24
Add ldd checks also for executables (#5).

marxin:


test/test_ldd_parser.py
```diff
@@ -59,7 +59,16 @@ def test_unused_dependency_in_package(binariescheck):
...
-    assert 'unused-direct-shlib-dependency ' in out
+    assert 'E: unused-direct-shlib-dependency ' in out
+
+
+def test_unused_dependency_in_package_for_executable(binariescheck):
+    output, test = binariescheck
+    test.run_elf_checks(FakePkg('fake'), get_full_path('appletviewer'), '/usr/bin/appletviewer')
+    assert not test.readelf_parser.parsing_failed()
+    assert not test.ldd_parser.parsing_failed
+    out = output.print_results(output.results)
```
scarabeusiv: I suppose debug output does not need to stay.

marxin: Hm, which debug output do you mean? I can't see it :P

PR #277, opened 2019-08-25
Add static-library-without-debuginfo check.

marxin:


scarabeusiv:
Conflicts, but I have question, can it even happen to not have the debuginfo? If it should almost never happen then it should be error, no?

marxin:
> Conflicts, but I have question, can it even happen to not have the debuginfo? If it should almost never happen then it should be error, no?

Yep, let's make it an error.

PR #278, opened 2019-08-25
Add static-library-without-symtab check.

marxin:


rpmlint/descriptions/BinariesCheck.toml
```diff
@@ -135,3 +135,7 @@ linked-against-usr-library="""
 Libraries and executables under /bin, /sbin, /lib and /lib64 may not link
 against a shared library in /usr folder.
 """
+static-library-without-symtab="""
+The static library doesn't contain any symbols and therefore can't be linked
+against. This may indicated that it was strip.
```
Conan-Kudo: `it was strip`? What does that mean?

marxin: It should be **stripped**, which indicate that `strip` tool was applied to a static library.

Issue #253, opened 2019-08-13
Replace os.path with pathlib where applicable


scarabeusiv:
We are using both approaches and it would make sense to have everything from pathlib.

Most notorous issue is that pkg.py needs to use pathlib but it is used also in following files:
 - [ ] test/test_scl.py
 - [ ] test/test_ldd_parser.py
 - [ ] test/test_filter.py
 - [ ] test/test_readelf_parser.py
 - [ ] rpmlint/filter.py
 - [ ] rpmlint/pkg.py
 - [ ] rpmlint/checks/TagsCheck.py
 - [ ] rpmlint/checks/RpmFileCheck.py
 - [ ] rpmlint/checks/InitScriptCheck.py
 - [ ] rpmlint/checks/SCLCheck.py
 - [ ] rpmlint/checks/SpecCheck.py
 - [ ] rpmlint/checks/ZipCheck.py
 - [ ] rpmlint/checks/MenuXDGCheck.py
 - [ ] rpmlint/checks/FilesCheck.py

marxin:
@scarabeusiv 
I've just made a pull request. What's remaining is the usage of:
- `os.path.normpath` - can we replace it with `Path.resolve()`? I know it's not 1:1
- `safe_normpath` - any suggestion what to do with this function?

marxin:
@scarabeusiv Please assign that to me.

scarabeusiv:
Looking at safe_normalpath it seems that we could kill that using the resolve tbh.

marxin:
> Looking at safe_normalpath it seems that we could kill that using the resolve tbh.

Good, I updated the pull request (#280). Please take a look.

marxin:
This is fixed, please close it.

PR #280, opened 2019-08-25
Remove partially os.path.* usage (#253).

marxin:


scarabeusiv:
@mimi1vx could you review this?

marxin:
`pathlib` does not provide a function like `os.path.normpath`, so we need to stay with it.

Issue #227, opened 2019-07-26
Merge in tests/testsuite from openSUSE rpmlint-tests


scarabeusiv:
There is shell testsuite residing at https://github.com/openSUSE/rpmlint-tests. This should not be needed and we should have all the coverage inplace in the rpmlint tests here.

This is tied to https://github.com/openSUSE/rpmlint-checks which contain some extra checks and need to be merged too (the enablement later on is simply tied to the distribution configuration, yet the checks can all be at one place with no problem).

The merge needs to happen prior 2.0 release that changed the syntax so none of the openSUSE checks will work otherwise.

@dirkmueller what would be your suggestion how to tackle this; it will need to happen after we finish with pkg.py changes but then all the tests should be tweakable independently and the api on how they should work should not change anymore.

Conan-Kudo:
None of the SUSE-specific checks should be pulled in. That means the SUSE FHS restrictions, BSD RC link enforcement, kmod packaging rules, etc. should not be incorporated upstream unless they are rationalized with the other major RPM distros (Fedora and Mageia).

scarabeusiv:
@Conan-Kudo I am not saying to enable them by default just to have them in one repo, even now not all of the checks here are enabled by default.

Conan-Kudo:
I don't want all the checks from openSUSE merged wholesale, even disabled.

scarabeusiv:
Care to explain why? They are not going to cause issues as it will be just checks with tests.

Even now some of the checks are distro specific only.

Conan-Kudo:
Mainly because it's going to be hard to keep up with all the extra code from those checks. And some of those checks are *really* irrelevant to us all.

A quick look through indicates the following checks are okay for contributing upstream:
* `BashismsCheck` (disable by default)
* `Check4Xinetd` (disable by default)
* `CheckBuildDate` (enable by default)
* `CheckBuildRoot` (enable by default)
* `CheckDBusPolicy` (needs some refactoring, but enable by default after)
* `CheckExecDocs` (remove `-lang` package related checks, enable by default)
* `CheckFilelist` (needs reworking to match unified base FHS, drop SUSE prefix, enable by default)
* `CheckIconSizes` (enable by default)
* `CheckLogrotate` (needs reworking to drop SUSE prefix, drop mandate for `%dir /var/log` in every package, disable by default)
* `CheckPkgConfig` (needs tweaking, also incorporate #95, enable by default)
[...]

PR #285, opened 2019-08-26
Add BashismsCheck from openSUSE/rpmlint-checks (#227).

marxin:


rpmlint/checks/BashismsCheck.py
```diff
@@ -0,0 +1,36 @@
+#############################################################################
+# File          : BashismsCheck.py
+# Purpose       : check /bin/sh shell scripts for bashisms
+#############################################################################
```
scarabeusiv: The header is kinda how to say it, pointless.

rpmlint/checks/BashismsCheck.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, config, output):
+        super().__init__(config, output, r'.*')
+
+    def check_file(self, pkg, filename):
+        root = pkg.dirName()
+        pkgfile = pkg.files()[filename]
+        fullpath = root + filename
+
+        if not (stat.S_ISREG(pkgfile.mode) and
+                pkgfile.magic.startswith('POSIX shell script')):
```
scarabeusiv: Isn't 'POSIX shell script' everything that ends with .sh ?
We have to really implement queries for the pkg.py something like pkg.configs, pkg.headers, pkg.binaries,...

marxin: > Isn't 'POSIX shell script' everything that ends with .sh ?

Nope:
```
$ file /usr/bin/* | grep 'POSIX shell script' | grep -v '\.sh' | wc -l
500
```

scarabeusiv: Interesting, i am not disputing that but I created sample .sh file where I set shebang to point to various areas:
 * bash ` Bourne-Again shell script, ASCII text executable`
 * env sh `a /usr/bin/env sh script, ASCII text executable`
 * /bin/sh `POSIX shell script, ASCII text executable`
 * /usr/bin/sh `a /usr/bin/sh script, ASCII text executable`

So probably it is correct right?

marxin: > So probably it is correct right?

I guess so.

test/test_bashisms.py
```diff
@@ -0,0 +1,25 @@
...
+
+@pytest.mark.parametrize('package', ['binary/bashisms'])
+def test_bashisms(tmpdir, package, bashismscheck):
+    output, test = bashismscheck
+    test.check(get_tested_package(package, tmpdir))
+    out = output.print_results(output.results)
+    # TODO: enable once we fix checkbashisms on Leap 15
+    # which is broken due to a Perl regex error:
+    # https://build.opensuse.org/request/show/725994
+    # assert 'W: potential-bashisms /bin/script1' in out
```
scarabeusiv: Just enable this, we will fix it SOON TM.

marxin:
@scarabeusiv  Should I disable the check by default?

scarabeusiv:
@marxin yes please this one should be off by default.

marxin:
> @marxin yes please this one should be off by default.

Done.

PR #286, opened 2019-08-26
Add shared-lib-not-executable check for libraries without exec bit se…

marxin:
…t (#94).


marxin:
Related issue #7.

marxin:
Patch has been updated to report only for shared libs that have `INTERP` header set.

rpmlint/checks/BinariesCheck.py
```diff
@@ -261,6 +262,12 @@ def _check_forbidden_functions(self, pkg, pkgfile_path, path):
         for fn in forbidden_functions_filtered:
             self.output.add_info('W', pkg, fn, path, forbidden_functions[fn]['f_name'])
 
+    def _check_executable_shlib(self, pkg, pkgfile_path, path):
+        if not self.is_exec and self.readelf_parser.is_shlib:
+            interp = [h for h in self.readelf_parser.program_header_info.headers if h.name == 'INTERP']
+            if len(interp):
```
scarabeusiv: `if interp:` is good enough here, empty lists are False so you don't need to calculate the length of it.

rpmlint/checks/BinariesCheck.py
```diff
@@ -261,6 +262,12 @@ def _check_forbidden_functions(self, pkg, pkgfile_path, path):
...
+                self.output.add_info('W', pkg, 'shared-lib-not-executable', path)
```
scarabeusiv: Also as it is really limited on which libs need this and they must be executable I would say this should be error.

PR #293, opened 2019-08-28
Add DBusPolicyCheck from openSUSE/rpmlint-checks (#227).

marxin:


rpmlint/checks/DBusPolicyCheck.py
```diff
@@ -0,0 +1,48 @@
...
+            if f in pkg.ghostFiles():
+                continue
+
+            # catch xml exceptions
+            try:
+                if f.startswith('/etc/dbus-1/system.d/'):
+                    send_policy_seen = False
+                    lf = pkg.dirName() + f
+                    xml = parse(lf)
+                    for p in xml.getElementsByTagName('policy'):
```
scarabeusiv: This is getting way too nested, could it be separated to functions please?

marxin: Yep. Fixed.

scarabeusiv:
This one I would recommend for default enabled treatment.

rpmlint/checks/DBusPolicyCheck.py
```diff
@@ -0,0 +1,54 @@
...
+                send_policy_seen = True
+
+            if (allow.hasAttribute('receive_sender') or
+                    allow.hasAttribute('receive_interface')):
+                self.output.add_info('W', pkg, 'dbus-policy-allow-receive', allow.toxml(), f)
+
+        for deny in policy.getElementsByTagName('deny'):
+            if (deny.hasAttribute('send_interface') and not
+                    deny.hasAttribute('send_destination')):
+                self.output.add_info('E', pkg, 'dbus-policy-deny-without-destination', allow.toxml(), f)
```
scarabeusiv: This for loop could be just one separate function as it is not tied to the policy_seen at all.

marxin: Done, actually I found an issue in `dbus-policy-deny-without-destination` error message where I used `allow.toxml()`.

PR #294, opened 2019-08-29
Add ExecDocCheck from openSUSE/rpmlint-checks (#227).

marxin:


rpmlint/descriptions/ExecDocCheck.toml
```diff
@@ -0,0 +1,8 @@
+executable-docs="""
+Documentation should not be executable.""""
```
scarabeusiv: There is one " too many

rpmlint/checks/ExecDocCheck.py
```diff
@@ -0,0 +1,44 @@
...
+            self.output.add_info('W', pkg, 'package-with-huge-docs: %3d%%' % (doc_size * 100 / complete_size))
+
+        for f in pkg.docFiles():
+            mode = files[f].mode
+            if not stat.S_ISREG(mode) or not mode & 0o111:
+                continue
+            path = Path(f)
+            for ext in ['txt', 'gif', 'jpg', 'html',
+                        'pdf', 'ps', 'pdf.gz', 'ps.gz']:
+                if path.suffix == '.' + ext:
```
scarabeusiv: Wouldn't this be nicer to use the ext with . before in the name? `for ext in [ '.txt'...`

rpmlint/checks/ExecDocCheck.py
```diff
@@ -0,0 +1,44 @@
...
+            if stat.S_ISREG(pkgfile.mode):
+                complete_size += pkgfile.size
+
+        doc_size = 0
+        for f in pkg.docFiles():
+            if stat.S_ISREG(files[f].mode):
+                doc_size += files[f].size
+
+        if doc_size * 2 >= complete_size and doc_size > 100 * 1024 and not self.ignore_pkg(pkg.name):
+            self.output.add_info('W', pkg, 'package-with-huge-docs: %3d%%' % (doc_size * 100 / complete_size))
```
scarabeusiv: This whole calculation above starting with line 18 should be separate function which gets called if not ignoring the package.

scarabeusiv:
Also this one should be by default on I suppose.

rpmlint/checks/ExecDocCheck.py
scarabeusiv: Still not what I ment, now from line 22 form this file until line 34 this hould be separate function only called when self.ignore_pkg for it is false.

rpmlint/checks/ExecDocCheck.py
```diff
@@ -0,0 +1,50 @@
...
+        for f in pkg.docFiles():
+            mode = files[f].mode
+            if not stat.S_ISREG(mode) or not mode & 0o111:
+                continue
+            path = Path(f)
+            extensions = ['.txt', '.gif', '.jpg', '.html', '.pdf', '.ps', '.pdf.gz', '.ps.gz']
+            if path.suffix in extensions:
+                self.output.add_info('E', pkg, 'executable-docs', f)
+
+            for name in ['README', 'NEWS', 'COPYING', 'AUTHORS']:
```
scarabeusiv: Maybe consider LICENSE, and LICENCE and all the variants we have in the list case insensitive? It is not a bug per-se but I think looking on the current spec files it is quite used...

PR #295, opened 2019-08-29
Add IconSizesCheck from openSUSE/rpmlint-checks (#227).

marxin:


Conan-Kudo:
@marxin What happens if an SVG file is installed?

rpmlint/checks/IconSizesCheck.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def check(self, pkg):
+        if pkg.isSource():
+            return
+
+        for fname, pkgfile in pkg.files().items():
+            if '/animations/' in fname:
+                continue
+
+            res = self.file_size_regex.search(fname)
```
scarabeusiv: This does not need to be so nested, just get both of the 'res' values at once and then if both contain results do the comparisons, no need to do if, anther if, another if....

scarabeusiv:
@Conan-Kudo it should not find in size-specific folders svg images, and if they are there they should be throwing a warning as they should be in the scalable dir.

rpmlint/checks/IconSizesCheck.py
```diff
@@ -0,0 +1,28 @@
+import re
+
+from rpmlint.checks.AbstractCheck import AbstractCheck
+
+
+class IconSizesCheck(AbstractCheck):
+    def __init__(self, config, output):
+        super().__init__(config, output)
+        self.file_size_regex = re.compile(r'/icons/[^/]+/(?P<x>\d+)x(?P<y>\d+)/')
+        self.info_size_regex = re.compile(r'(?P<x>\d+) x (?P<y>\d+)')
```
scarabeusiv: These two could be class variables declared without init and then init itself can be dropped.

Conan-Kudo:
@scarabeusiv Unfortunately, this is not what is happening. There are icon themes that are shipping SVGs into all the size directories. For example, in Fedora, we have `elementary-icon-theme` doing this.

scarabeusiv:
Hmm in that case we need to update the regexp to skip svgs

marxin:
> Hmm in that case we need to update the regexp to skip svgs

That should be fine as `pkgfile.magic` will contains something like:
`/usr/.../route-transit-tram-symbolic.svg: SVG Scalable Vector Graphics image`

scarabeusiv:
Right so the regexp wont catch up and pass up and the test is skipped. LGTM then.

PR #296, opened 2019-08-29
Add LogrotateCheck from openSUSE/rpmlint-checks (#227).

marxin:


marxin:
> Can you please readd the copyright that was in the original file?

That's definitely possible. However, I know that @scarabeusiv asked me to remove the file header with the copyright when I did a port from `rpmlint-check`.
@scarabeusiv: What do you suggest?

scarabeusiv:
That was about those current ones that are listed in authors, if there are new peepz we need to add them there.

marxin:
> That was about those current ones that are listed in authors, if there are new peepz we need to add them there.

Do we have somewhere a list of authors? Or do you mean `git grep '# Author'`?

marxin:
> Do we have somewhere a list of authors? Or do you mean `git grep '# Author'`?

Looking at the history, AUTHORS file was removed in c80f9feac1b83b52e3a7541e4009006cd40c1ead.
So the list of authors will be generated from `git log`.

scarabeusiv:
@marxin ok then I would stick with my other proposal, 2 commits:
1) initial with the orginal author
2) rewrite

marxin:
> @marxin ok then I would stick with my other proposal, 2 commits:
> 
> 1. initial with the orginal author
> 2. rewrite

1

> 1. initial with the orginal author
> 2. rewrite

And will be the author line removal part of **rewrite**?

marxin:
Ok, I made the rebase.

PR #297, opened 2019-08-30
Add PkgConfigCheck from openSUSE/rpmlint-checks (#227).

marxin:


rpmlint/checks/PkgConfigCheck.py
```diff
@@ -0,0 +1,35 @@
...
+class PkgConfigCheck(AbstractFilesCheck):
+    def __init__(self, config, output):
+        super().__init__(config, output, r'.*/pkgconfig/.*\.pc$')
+        # currently causes too many failures (2008-03-05)
+        self.suspicious_dir = re.compile(r'[=:](?:/usr/src/\w+/BUILD|/var/tmp|/tmp|/home)')
+
+    def check(self, pkg):
+        # check for references to /lib when in lib64 mode
+        if pkg.arch in ('x86_64', 'ppc64', 's390x'):
+            self.wronglib_dir = re.compile(r'-L/usr/lib\b')
```
scarabeusiv: aarch64?

rpmlint/checks/PkgConfigCheck.py
```diff
@@ -0,0 +1,35 @@
...
+import stat
+
+from rpmlint.checks.AbstractCheck import AbstractFilesCheck
+
+
+class PkgConfigCheck(AbstractFilesCheck):
+    def __init__(self, config, output):
+        super().__init__(config, output, r'.*/pkgconfig/.*\.pc$')
+        # currently causes too many failures (2008-03-05)
+        self.suspicious_dir = re.compile(r'[=:](?:/usr/src/\w+/BUILD|/var/tmp|/tmp|/home)')
```
scarabeusiv: could be as a class variable and you could kill the init.
also any idea uf the comment makes sense after 10 years?

rpmlint/checks/PkgConfigCheck.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def check_file(self, pkg, filename):
+        if pkg.isSource() or not stat.S_ISREG(pkg.files()[filename].mode):
+            return
+
+        if pkg.grep(self.suspicious_dir, filename):
+            self.output.add_info('E', pkg, 'invalid-pkgconfig-file', filename)
+
+        pc_file = open(pkg.dirName() + '/' + filename, 'r',
+                       encoding='utf-8', errors='ignore')
```
scarabeusiv: shouldn't we rather report an error if the pc file cannot be read?

rpmlint/checks/PkgConfigCheck.py
marxin: I recommend to remove the comment and we'll see how much it will generate errors.

rpmlint/checks/PkgConfigCheck.py
```diff
@@ -0,0 +1,37 @@
...
+
+        try:
+            pc_file = open(pkg.dirName() + '/' + filename, 'r', encoding='utf-8')
+            for l in pc_file:
+                if l.startswith('Libs:') and self.wronglib_dir.search(l):
+                    self.output.add_info('E', pkg, 'pkgconfig-invalid-libs-dir', filename, l)
+                if '//' in l and '://' not in l:
+                    self.output.add_info('E', pkg, 'double-slash-in-pkgconfig-path', filename, l)
+        except Exception as e:
+            self.output.add_info('E', pkg, 'pkgconfig-exception', filename, str(e))
```
scarabeusiv: no description for this error?

PR #300, opened 2019-09-02
Disable temporarily broken environments.

marxin:


scarabeusiv:
Nah we should fix the Leap, which I have on my todo still and the Fedora should get fixed soon TM; they just updated for python3.8.

PR #301, opened 2019-09-02
Add SystemdInstallCheck from openSUSE/rpmlint-checks (#227).

marxin:


marxin:
> This needs to be reworked to orient around the upstream systemd macros.

Can you be please more concrete here? Are you talking about `PRE_POST_PATTERN`, `PREUN_PATTERN`, `POSTUN_PATTERN` regexes?

Conan-Kudo:
@marxin Yes. Also the messages, etc.

The openSUSE specific messages, hints, and structures need to be replaced with ones reflecting what systemd upstream recommends.

marxin:
> @marxin Yes. Also the messages, etc.
> 
> The openSUSE specific messages, hints, and structures need to be replaced with ones reflecting what systemd upstream recommends.

Ok, we as SUSE use following systemd macros:
https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines

These are different from these which are defined in https://github.com/systemd/systemd/blob/master/src/core/macros.systemd.in.

About the rpm expansion. It seems to me not nice to seek for expanded values of the macros. Example:
`%{systemd_preun XYZ}`:
```
if [ $1 -eq 0 -a -x /usr/bin/systemctl ]; then						
	# Package removal, not upgrade							
	/usr/bin/systemctl --no-reload disable XYZ || :				
	
	test -z "$DISABLE_STOP_ON_REMOVAL" -a -f /etc/sysconfig/services &&\
		. /etc/sysconfig/services						
	test "$DISABLE_STOP_ON_REMOVAL" != yes -a "$DISABLE_STOP_ON_REMOVAL" != 1 &&\
		/usr/bin/systemctl stop XYZ || :					
		
fi											
```

It's a multiline string and XYZ has to be replaced with a regex with service name.
That said, I would incline to disable it by default and we use it only in SUSE.

Conan-Kudo:
@marxin openSUSE ships both variants, as I understand it. But if you don't want to adapt this for upstream macros which are used in most distributions, then this is not acceptable to merge.

marxin:
> @marxin openSUSE ships both variants, as I understand it. But if you don't want to adapt this for upstream macros which are used in most distributions, then this is not acceptable to merge.

I'm fine with adapting both macros (openSUSE and upstream). But what I see problematic is to expand macros and then search for the needles in rpms. Any suggestions how to do it?

Issue #292, opened 2019-08-28
Tracker for to-rebase openSUSE patches


scarabeusiv:
This is not mostly for downstream openSUSE as we have list of patches we will have to review prior the release, maybe even some can be upstreamed, but it is mostly for downstream review we have to do.

List of stuff to check:
 - [x] devel-provide-is-devel-package.diff
 - [x] docdata-examples.diff
 - [x] check-for-self-provides.diff
 - [x] invalid-filerequires.diff
 - [x] no-doc-for-lib.diff
 - [x] only-reg-files-are-scripts.diff
 - [x] rpmgroup-checks.diff
 - [x] rpmlint-slpp-NUM-NUM.patch
 - [x] suse-binarieschecks.diff
 - [x] suse-pkg-config-check.diff
 - [x] suse-python3-naming-policy.diff
 - [x] suse-rpmlint-all-pie.patch
 - [x] suse-shlib-devel-dependency.diff
 - [x] suse-skip-macro-expansion.diff
 - [x] suse-spdx-license-exceptions.patch
 - [x] suse-version.diff

scarabeusiv:
@mimi1vx @kstreitova @marxin ^^ :)

marxin:
@scarabeusiv Thanks for it. Can you please also upload the patches to https://gist.github.com/ ?
What's bad about the issue is that I can't modify your commend with the checklist :/

scarabeusiv:
@marxin kein ahnung. Just tell me on irc what I should check :)) The gist will take me a bit as I am still doing the training so it might take me a bit :)

marxin:
Ok, there's a gist for all the patches: https://gist.github.com/marxin/4345da142ebf76117b6acd08d015c37e

marxin:
> @marxin kein ahnung. Just tell me on irc what I should check :)) The gist will take me a bit as I am still doing the training so it might take me a bit :)

Ok, please check `suse-binarieschecks.diff`, all stuff in the patch has been ported. And please sort the list of the patches ;)

scarabeusiv:
But the list is sorted alphabetically. :)

marxin:
About `only-reg-files-are-scripts.diff`: Can we drop `./rpmlint/checks/InitScriptCheck.py` as `init.d` is dead?

scarabeusiv:
I think @Conan-Kudo said some distro still uses them, but I wonder if we really have such a product somewhere that is rpm based yet non-systemd.

[...]

PR #305, opened 2019-09-03
Remove commented out package-provides-itself check.

marxin:
The check is already covered with useless-provides check and
I added test for it.


marxin:
Once installed it will cover `devel-provide-is-devel-package.diff` in #292.

PR #306, opened 2019-09-03
Make an exception for spurious-executable-perm.

marxin:
Except files that live in %docdir/examples.


marxin:
It's covering `docdata-examples.diff` in #292.

rpmlint/descriptions/FilesCheck.toml
```diff
@@ -123,6 +123,8 @@ spurious-executable-perm="""
 The file is installed with executable permissions, but was identified as one
 that probably should not be executable. Verify if the executable bits are
 desired, and remove if not.
+NOTE: example scripts should be packaged under %docdir/examples, which will avoid
```
Conan-Kudo: Isn't this `%{_docdir}`?

test/test_files.py
```diff
@@ -149,3 +149,12 @@ def test_lib_regex():
...
          '/usr/lib64/rsocket/binary',))
+
+
+@pytest.mark.parametrize('package', ['binary/executableperm'])
+def test_executable_permission(tmpdir, package, filescheck):
+    output, test = filescheck
+    test.check(get_tested_package(package, tmpdir))
+    out = output.print_results(output.results)
+    assert 'W: spurious-executable-perm /usr/share/doc/packages/bwidget/Widget.html' in out
+    assert 'W: spurious-executable-perm /usr/share/doc/packages/bwidget/script/Widget2.html' not in out
```
Conan-Kudo: Ugh, please don't allow `/usr/share/doc/packages` to enter upstream rpmlint. Nobody else does such a boneheaded modification.

Conan-Kudo:
@marxin I'm not sure if examples should be permitted to be executable, though... Why not just install them as regular binaries, then?

test/test_files.py
marxin: What wrong with that? It's only a test-case that I added here..

Conan-Kudo:
I'm inclined to say this is not okay at all. If random binary files are in `/usr/share/doc` then that's a problem.

marxin:
> I'm inclined to say this is not okay at all. If random binary files are in `/usr/share/doc` then that's a problem.

I'm fine with that.
@scarabeusiv: please check the patch in #292

PR #307, opened 2019-09-03
Add devel-package-with-non-devel-group warning (#292).

marxin:
It covers `rpmgroup-checks.diff` patch.


Conan-Kudo:
@marxin Remove the patch name from the commit message, and then I'll merge it.

marxin:
> @marxin Remove the patch name from the commit message, and then I'll merge it.

Done, thanks.

PR #310, opened 2019-09-03
Add new check no-pkg-config-provides (#292).

marxin:


rpmlint/descriptions/TagsCheck.toml
```diff
@@ -168,3 +168,7 @@ devel-package-with-non-devel-group="""
 The package ends with -devel but does not have a RPM group starting with
 Development/.
 """
+no-pkg-config-provides="""
+The package installs a .pc file but does not provide pkgconfig(..) provides.
+The most likely reason for that is that it was built without BuildRequires: pkg-config.
```
Conan-Kudo: `BuildRequires: pkgconfig`

rpmlint/checks/TagsCheck.py
```diff
@@ -225,6 +228,15 @@ def check(self, pkg):
...
                             self.output.add_info('W', pkg, 'no-provides', prov)
 
+                if has_pc:
+                    found_pkg_config_dep = False
+                    for p in (x[0] for x in pkg.provides()):
+                        if p.startswith('pkgconfig('):
+                            found_pkg_config_dep = True
+                            break
+                    if not found_pkg_config_dep:
+                        self.output.add_info('W', pkg, 'no-pkg-config-provides')
```
scarabeusiv: This should be fatal imo. I just looked on the report and we have no reports for it and when i triggered it manualy it is pretty fatal to lack the provides.

Conan-Kudo: It basically should never happen since a pkgconfig implementation should be in the buildroot in most distributions.

PR #311, opened 2019-09-03
Add SUSE version checks in spec files (#292).

marxin:


Conan-Kudo:
This is not acceptable for upstream.

marxin:
> This is not acceptable for upstream.

@scarabeusiv 
Why? The macro names are not clashing with anything else. Yes, it's SUSE specific, but it's not invasive or limiting any other distribution.

Conan-Kudo:
@marxin When @scarabeusiv and I first talked about this, I said that I'd take as many of the SUSE checks as possible that are generally applicable, because they are broadly useful.

I have no intention of taking a check that is only useful for one distribution, especially since adding your own checks is still going to be supported in rpmlint.

PR #312, opened 2019-09-04
Add FilelistCheck from openSUSE/rpmlint-checks (#227).

marxin:


rpmlint/checks/FilelistCheck.py
```diff
@@ -0,0 +1,410 @@
...
+            '/var/yp/',
+            # those are not in FHS!
+            '/var/adm/',
+            '/var/nis/',
+            '/emul/',
+            '/run/',
+        )
+
+    def init_checks(self):
+        self._checks = [
```
scarabeusiv: This whole crazy dictionary could be for readability converted in toml file(s) specifying the format as it really does not need to be in here.

rpmlint/checks/FilelistCheck.py
```diff
@@ -0,0 +1,410 @@
...
+        ]
+
+    def check(self, pkg):
+        if pkg.isSource():
+            return
+
+        files = pkg.files()
+
+        if not files:
+            self.output.add_info('W', pkg, 'filelist-empty')
```
scarabeusiv: This does not have a description, and I can't imagine this situation happening tbh.

rpmlint/checks/FilelistCheck.py
```diff
@@ -0,0 +1,410 @@
...
+                else:
+                    invalidfhs.add(pfx)
+
+            if f.startswith('/opt'):
+                try:
+                    provider = f.split('/')[2]
+                except Exception:
+                    continue
+                # legacy exception
+                if provider == 'kde3':
```
scarabeusiv: this looks too legacy...

rpmlint/checks/FilelistCheck.py
marxin: Good idea. Will we then use 'ignorefileif' with string names of functions that will then found and called?

marxin: It's doable easily with `getattr(self, 'notsymlink')`.

marxin:
> At a glance, there are several things wrong with it:
> 
> * This has SUSE-specific stuff in it

Can you be please more specific?

> * This hasn't been updated for FHS 3.0

Yep, that can be added as a follow up patch.

> * `/usr/libexec` is not in here, despite most distributions using it

What exactly should be checked in the folder?

> * There are checks in here repeated by other checks (like the systemd, udev, and xinetd ones)

Ok, I can remove the overlap.

> 
> This needs a lot of fixing before I'd consider it for merging in rpmlint.

Can you be please more specific?

marxin:
> * There are checks in here repeated by other checks (like the systemd, udev, and xinetd ones)

I verified this and I see only an overlap with a `udev` check. I added one commit that addresses that.

Conan-Kudo:
>> This has SUSE-specific stuff in it

> Can you be please more specific?

* SUSE specific paths (SuSEfirewall2, fillupdir, YaST2, SuSE capitalization, etc.)
* SUSE vendor check (which I had just noticed, and now I feel paranoid about checking the others for that...)
* `/usr/src/packages`, `/usr/share/doc/packages`
* Advocacy of `/srv` (all distributions _except_ SUSE use `/var` for this)

There's probably more in here, but those are what I see at a glance...

>> `/usr/libexec` is not in here, despite most distributions using it

> What exactly should be checked in the folder?

The directory needs to be in the good prefix list.

>> There are checks in here repeated by other checks (like the systemd, udev, and xinetd ones)

> I verified this and I see only an overlap with a udev check. I added one commit that addresses that.

I want those remaining checks moved to the appropriate existing checks we have for those. Especially now that I've merged so many of them...

PR #313, opened 2019-09-06
Remove useless ld --version and c++filt --version.

marxin:
We can set self.parsing_failed even w/o --version call.


Conan-Kudo:
@marxin Looks good, I'll wait for CI before merging...

PR #315, opened 2019-09-09
Separate TmpFilesCheck to functions and add a test

kstreitova:
Separate the check() method to multiple methods and clean up the code.
Also add a test (systemd-tmpfiles-1.0-0.x86_64.rpm) that contains
correct handling of tmpfiles configuration files and test that these
tmpfiles warnings are issued correctly.

Don't check %pre section for "systemd-tmpfiles --create" call as it
should be placed in the %post section.

Remove "now check remaining ghost files that are not already" part
as the same code is already present in PostCheck.py and it's not related
to tmpfiles anyway. This means that "ghost-files-without-postin" and
"postin-without-ghost-file-creation" checks were removed from
TmpFilesCheck.py.

Rename postin-without-tmpfile-creation to post-without-tmpfile-creation
as we want to address %post section (there is no %postin).

#225 #156


kstreitova:
I'm not sure about the tmpfiles and %files section. Currently, it checked just the tmpfiles presence in %files but the description of the warning said that it should be %ghost. However, neither Fedora nor SUSE documentation mentions that it should be %ghost. What do you think about it?

Also, SUSE documentation says that the  "systemd-tmpfiles --create" call should be in the %post section. So I removed checking of the %pre section. Please validate.

Thanks!

scarabeusiv:
Was that removed %ghost check used in openSUSE? Is rpmlint.opensuse.org reporting it somewhere?

Also for the tmpfiles, it actually makes sense to do it in the post rather than in %pre so it should even be reported like bad idea to do it in %pre phase.

kstreitova:
> Was that removed %ghost check used in openSUSE? Is rpmlint.opensuse.org reporting it somewhere?

Yes, it was used. According to rpmlint.opensuse.org there were around 80 packages that violate this rule.

> Also for the tmpfiles, it actually makes sense to do it in the post rather than in %pre so it should even be reported like bad idea to do it in %pre phase.

Ok, I will add this check.

Conan-Kudo:
Wait, so openSUSE still isn't using the systemd file triggers? https://github.com/systemd/systemd/blob/master/src/core/triggers.systemd.in

scarabeusiv:
@Conan-Kudo seems so; you can ask the systemd maintainers tho to confirm...

kstreitova:
I added the check for %pre. Also, I suggest leaving the %ghost situation as-is for now. I will open an issue so we can discuss it there. Ready to merge.

Conan-Kudo:
@kstreitova I'll merge after the pipeline passes.

PR #316, opened 2019-09-09
Separate ZipCheck to functions a little bit more

kstreitova:
Separate the check() method to multiple methods and clean up
the code. Also add a test for ZipCheck
(bad-crc-uncompressed-1.0-9.1.x86_64.rpm) that tests crc issues
and uncompressed zipfiles (#225 #156).

Also fix _check_jarindex() method that had swapped return values.


rpmlint/checks/ZipCheck.py
```diff
@@ -19,72 +19,82 @@ def check(self, pkg):
...
 
+        # the META-INF is optional so skip if it is not present
         mf = 'META-INF/MANIFEST.MF'
-        # The META-INF is optional so skip if it is not present
-        if mf not in zipfile.namelist():
-            return False
-        # otherwise check for the classpath
-        manifest = zipfile.read(mf).decode()
+        if mf not in jarfile.namelist():
+            return
```
scarabeusiv: As this function and the one bellow check for this meta-inf I guess this could even be detected prior going to these two functions.

kstreitova: Actually this meta-inf check is not necessary for _check_jarindex() method so I removed it completely.

PR #318, opened 2019-09-11
Add FilelistCheck from openSUSE/rpmlint-checks (#227).

marxin:


marxin:
> There's probably more in here, but those are what I see at a glance...

Yep, it's a downstream pull request, so it may contain SUSE-specific stuff.

rpmlint/checks/FilelistCheck.py
```diff
@@ -0,0 +1,144 @@
...
+                    invalidfhs.add(f)
+                else:
+                    invalidfhs.add(pfx)
+
+            if f.startswith('/opt'):
+                try:
+                    provider = f.split('/')[2]
+                except Exception:
+                    continue
+                if is_suse and (provider == 'suse' or provider == 'novell'):
```
scarabeusiv: Any idea what for is this is_suse thingy?

marxin: Dunno. I would take is as-is.

Conan-Kudo: It's an exception for `/opt`. If the package has `SUSE` or `Novell` and the host system running rpmlint is a SUSE system, `/opt` rules don't apply.

scarabeusiv: ok, lets keep it ugly as-is.

PR #324, opened 2019-09-13
Remove DistributionCheck

kstreitova:
Remove DistributionCheck as "invalid-vendor" and "invalid-distribution"
checks are useless (neither openSUSE nor Fedora uses these tags).

"manpage-not-compressed" and "infopage-not-compressed" checks were moved
to FilesCheck as it makes more sense to have it there.


rpmlint/checks/DistributionCheck.py
```diff
@@ -1,68 +0,0 @@
...
-
-
-class DistributionCheck(AbstractCheck):
-
-    def __init__(self, config, output):
-        super().__init__(config, output)
-        self.man_regex = re.compile(r'/man(?:\d[px]?|n)/')
-        self.info_regex = re.compile(r'(/usr/share|/usr)/info/')
-        self.vendor = self.config.configuration['Vendor']
-        self.distribution = self.config.configuration['Distribution']
```
scarabeusiv: If vendor and distribution config options are no longer used we should kill them from configuration too.

Conan-Kudo: they are used, but they're auto-set by our buildsystems (Fedora by Koji, openSUSE by OBS). I'm pretty sure this was a check that was more meaningful in the era of mixed buildsystems and such. Nowadays, there's not really a problem.

kstreitova: Ok, removed.

Conan-Kudo:
@kstreitova Could you please split the move of the manpage/infopage checks into a separate commit before removing DistributionCheck? If someone complains and wants it restored, I want an easy path to revert.

Issue #38, opened 2013-04-02
RFE: add documentation about which checks are run


scop:
(From http://rpmlint.zarb.org/cgi-bin/trac.cgi/ticket/180, scop)

Copied from  https://bugzilla.redhat.com/show_bug.cgi?id=521622:

"While running rpmlint on a specfile and then on srpm the sets of spec tests differ. It is mentioned within the manpage and the readme file as CAVEATS:

"All checks do not apply to all argument types. For best check coverage, run rpmlint on source rpms instead of plain specfiles, and installed binary rpms instead of uninstalled binary rpm files."

But there is no list, or reference to other documentation, saying which checks apply for the particular input type. Please add some aid to tell what is going to be checked."

Original comment by: scop

scarabeusiv:
Actually this sounds nice. Something like the printout that is done by pytest (header listing all the loaded tests, list of configs that are in use, then all the errors/warnings and at the end nice summary.
Pytest does it nicely so lets get some inspiration :)

PR #327, opened 2019-09-17
Generate information header prior running the rpmlint wrt issue #38

scarabeusiv:
It prints out version and configs/rpmlintrc everytime.
And in verbose mode it prints out the loaded checks and requested files.

### DIFF
--- a/rpmlint/color.py
+++ b/rpmlint/color.py
@@ -1,0 +1,20 @@
+import sys
+
+if hasattr(sys.stderr, 'isatty') and sys.stderr.isatty():
+    class Color(object):
+        """
+        Colors used when doing printouts with rpmlint
+        """
+        Bold = '\x1b[1m'
+        Red = '\x1b[31m'
+        Yellow = '\x1b[33m'
+        Reset = '\x1b[0m'
+else:
+    class Color(object):
+        """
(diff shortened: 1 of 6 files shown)


scarabeusiv:
I wonder if we should use python-wcwidth to detect term lenght and do some wrapping, also if we should use colorama for some nice coloring there...

Conan-Kudo:
> I wonder if we should use python-wcwidth to detect term lenght and do some wrapping, also if we should use colorama for some nice coloring there...

I don't have anything against it, so if you'd like to add it, I don't have a problem with it.

scarabeusiv:
@Conan-Kudo I decided to not add any extra pkg dep and I will just expect nobody will try to use the rpmlint to display chinese/japanese characters.

Conan-Kudo:
@scarabeusiv I don't know if that's a good expectation. The Fedora and openSUSE zh communities do tend to do this, but it's up to you whether it should be fixed now or not...

scarabeusiv:
@Conan-Kudo it will print it out, it will just not be 100% centered as the width of some chars is not 1 character...

Conan-Kudo:
@scarabeusiv Ah, well, then I'm not too concerned right now.

PR #331, opened 2019-09-19
Come up with position-independent-executable-suggested (#292).

marxin:
Right now we have default value of self.pie_exec_re == ''. Which
will always succeed and so that we always report non-position-independent-executable
for any file. I'm suggesting to split the error and fallback to a new
warning.


scarabeusiv:
Aw darn, this is my fuckup, lgtm.

PR #333, opened 2019-09-19
Remove flaky macro expansion of private_so_paths (#292).

marxin:
The macro expansion leads to a unstable expected locations.


marxin:
#292

Conan-Kudo:
@marxin That code seems like it works as intended...?

marxin:
> @marxin That code seems like it works as intended...?

Ok, then I make a PR to the `opensuse` branch.

Conan-Kudo:
@marxin Well, I want to know what's wrong that led to this patch being made. I'm not necessarily saying "no" out of the gate here.

marxin:
@scarabeusiv Can you please help us here?

scarabeusiv:
This expansion was flaky and caused trouble for us in openSUSE because it is evaluated in some case to other stuff than what was anticipated. Now even fedora people complain that it expands to stuff with output to stdout. As such it is better to just not do it (like we do in openSUSE) and come up with better way how to achieve the result if desirable...

Conan-Kudo:
@scarabeusiv Sounds good to me. Could you create an issue for doing this better? I'll merge this for now.

PR #337, opened 2019-09-19
Generalize shared library naming policy.

marxin:
In order to cover also openSUSE SLPP, I would like to extend
the base_or_libs_regex.

### DIFF
--- a/rpmlint/checks/TagsCheck.py
+++ b/rpmlint/checks/TagsCheck.py
@@ -193,10 +193,10 @@ class TagsCheck(AbstractCheck):
                     if pkg_config_regex.match(fname) and fname.endswith('.pc'):
                         has_pc = True
                 if has_so:
-                    base_or_libs = base + '/' + base + '-libs/lib' + base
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@marxin What is the effect of this change?

marxin:
> @marxin What is the effect of this change?

It will catch library names as `libfoo-1.2.3.so`.

scarabeusiv:
It is just allowing slpp package names. As was the pull #59

Conan-Kudo:
@scarabeusiv I don't know what SLPP is, so that's not helpful. :(

scarabeusiv:
Just for the ref:
https://www.debian.org/doc/debian-policy/ch-sharedlibs.html
https://en.opensuse.org/openSUSE:Shared_library_packaging_policy

Conan-Kudo:
This was always allowed, as Mandriva did it that way. OpenSUSE's version is a little weird in that underscores are used instead of decimal points (IMO, that should be fixed...)

PR #343, opened 2019-09-24
Add RCLinksCheck.

marxin:
The check is about interaction in between `systemd` and `init.d`.


Conan-Kudo:
Not permitting this for upstream. You should also retire this as policy in openSUSE.

marxin:
> Not permitting this for upstream. You should also retire this as policy in openSUSE.

Yep, it was intended for `opensuse` branch.
@scarabeusiv: Do we still want to follow the policy in openSUSE?

Conan-Kudo:
@marxin The systemd shadowed initscript probably belongs in `SysVInitOnSystemdCheck` instead of here.

scarabeusiv:
Yeah the rclink policy is still adhered, and most of our documentation sadly points to it so I don't see that being deprecated any time soon.

Conan-Kudo:
@scarabeusiv If rpmlint doesn't enforce it anymore, then it'll slowly go away as it's not needed.

scarabeusiv:
It is enforced as of now on Tumbleweed so we should not regress from there. Maybe there could be some discussion on -factory if it is really needed again...

PR #345, opened 2019-09-26
Add RCLinksCheck.

marxin:


rpmlint/checks/RCLinksCheck.py
```diff
@@ -0,0 +1,44 @@
...
+                    rccandidates.add(basename.rpartition('.')[0])
+            elif fname.startswith('/etc/init.d/'):
+                basename = Path(fname).name
+                if not (basename.startswith('rc') or
+                        basename.startswith('boot.')):
+                    initscripts.add(basename)
+
+        for fname in sorted(initscripts):
+            if fname in rccandidates:
+                self.output.add_info('E', pkg, 'systemd-shadowed-initscript', fname)
```
scarabeusiv: This should for sure go to upstream at:
  https://github.com/rpm-software-management/rpmlint/blob/master/rpmlint/checks/SysVInitOnSystemdCheck.py

Conan-Kudo: Didn't I say this in the original rclinks PR?

marxin: If so, then the 2 checks should be probably merged. So that I'm closing the PR for now.

PR #346, opened 2019-09-29
Add cProfile to travis

mimi1vx:


Conan-Kudo:
@mimi1vx Looks like the Python 3.6 in Leap 15.1 doesn't have the patch to auto-coerce to `C.UTF-8` locale... It's not handling things well..

Conan-Kudo:
@scarabeusiv If it helps, here's the patch from Fedora: https://src.fedoraproject.org/rpms/python36/blob/master/f/00262-pep538_coerce_legacy_c_locale.patch

scarabeusiv:
@mimi1vx please, rebase the new python3 is already published these days.

scarabeusiv:
As this also breaks leap that has pending MU it is just SUSE internal screwup it takes us ages I will merge it too.

PR #348, opened 2019-09-30
Move DocFilesCheck.py checks to DocCheck.py

kstreitova:
Move all DocFilesCheck.py checks to DocCheck.py as it logically belongs
there. Add comments, test cases and two test rpms:
  * install-file-in-docs-1.0-0.x86_64.rpm
  * dogtail-0.9.0-5.29.noarch.rpm

(#225 #156)


Conan-Kudo:
`dogtail`? Isn't that a real package?

kstreitova:
Yes, it is a real package. It was the fastest solution on how to obtain a package with additional dependencies in the documentation for me right now.

Conan-Kudo:
@kstreitova Can we please not do that? That adds licensing complexity that I just don't want to deal with.

kstreitova:
Sure, ready to merge.

PR #351, opened 2019-10-03
Try to use podman instead of docker to have it work on the enterprise…

scarabeusiv:
…y stuff too.

If it works then it is 1:1, on local build it works, even with the rootless mapping, but I dunno if travis will actually manage to run it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,28 +2,40 @@ language: python
 
 sudo: required
 
-services:
-  - docker
-
 env:
   - DIST=fedora30
   - DIST=fedoradev
   - DIST=opensuselp15
   - DIST=opensusetw
 
+# podman is currently not in Ubuntu
+addons:
+  apt:
+    update: true
+    sources:
+      - sourceline: "ppa:projectatomic/ppa"
+    packages:
+      - podman
+
 # the python here is basically no-op just used for coveralls
 python:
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@scarabeusiv `podman` doesn't exist in Ubuntu.

scarabeusiv:
And I am looking if it could be made to work and for that I need this open so the travis triggers it for me...

scarabeusiv:
So it seems to be operable, the only issue is that it uncovered a bug on Tubmleweed I have to fix first.

junaruga:
I am happy to see this PR.
If you find a issue related to podman, I recommend to set `dist: xenial` to `.travis.yml`.
I am not sure which distribution version is running without setting `dist`.

podman on `dist: bionic` by root user (`sudo`) has a bug.
See my investigation on my repository.
https://github.com/junaruga/libpod/blob/feature/add-travis-ci/.travis.yml
https://travis-ci.org/junaruga/libpod/builds/592103783

scarabeusiv:
@Conan-Kudo see, I told ya it will work :) And I even fixed a bug in TW as a result from that :)

Conan-Kudo:
Yeah, yeah. I'll merge it once the pipeline passes.

Issue #354, opened 2019-10-14
check for control characters in header fields


M0ses:
As we have now

https://github.com/rpm-software-management/createrepo_c/commit/317ded66bb94e6d3dc80bde6c22c5a2cec8f897d

in place, to avoid broken xml file created by createrepo_c, we now should have some checks in rpmlint also, so we can detect failures ASAP.

PR #355, opened 2019-10-23
check tags for forbidden control characters

M0ses:
The newer versions of createrepo_c break on forbidden control characters
in several tags. This patch implements this behavior for rpmlint also, to
fail as early as possible.

Fixes: #354


scarabeusiv:
It does not pass flake8 (see the travis).

M0ses:
fixed

Conan-Kudo:
Still some failures:

```
./rpmlint/pkg.py:180:1: E302 expected 2 blank lines, found 1
./rpmlint/pkg.py:197:1: E303 too many blank lines (3)
./rpmlint/checks/TagsCheck.py:429:9: E303 too many blank lines (2)
```

scarabeusiv:
Could you also add testing .spec file that would exhibit the behaviour so we could test the detection?

scarabeusiv:
There is `test/test_tags.py` which already checks some of the behaviour, but you can add trivial trigger case for this too I suppose.

M0ses:
I have two binary packages as reallife examples (92K and 64K). Would this also be sufficient?

Conan-Kudo:
I don't think we need binary packages, just spec files.

M0ses:
Hmmm, after trying to implement test cases based on spec files I realized that my patch only checks binary packages - when checking a spec file, it does not find any issues.

I can understand, that checks based on a spec file might be good also but I would like to treat this as a different issue.

The main point of my patch is to find issues caused by automatic Requires/Provides which can only be found in the binary packages.

Conan-Kudo:
Your check should be checking both spec and binary packages, but I guess we can have more binary packages for test cases... :/

rpmlint/checks/SpecCheck.py
```diff
@@ -420,14 +435,25 @@ def check_spec(self, pkg):
...
                                                  conf)
 
             if current_section == 'changelog':
+                e = Pkg.has_forbidden_controlchars(line)
+                if e:
+                    self.output.add_info('E', pkg,
+                                         'forbidden-controlchar-found',
+                                         '%%changelog: %s' % e)
                 for match in self.macro_regex.findall(line):
+                    print(line)
```
scarabeusiv: This honestly looks like debug output.

M0ses: My fault, sorry

PR #360, opened 2020-01-01
Functionalize BinariesCheck

kstreitova:
Review and rewrite checks to be more "functionalized". Add test rpms and test cases.


scarabeusiv:
@kstreitova fails the testsuite in one of the tests.

rpmlint/descriptions/BinariesCheck.toml
```diff
@@ -12,6 +12,9 @@ noarch-with-lib64="""
 This package is marked as noarch but installs files into lib64.
 Not all architectures have this in path, so the package can't be noarch.
 """
+no-soname="""
+The library has no soname.
```
Conan-Kudo: This should be `no-soversion`. Shared object names do not require versions. You're attempting to mandate that a soname contains a soversion. Adjust accordingly.

rpmlint/descriptions/BinariesCheck.toml
```diff
@@ -63,8 +66,8 @@ themselves.
 incoherent-version-in-name="""
 The package name should contain the major version of the library.
 """
-invalid-directory-reference="""
-This file contains a reference to /tmp or /home.
+invalid-la-file="""
+This .la file contains a reference to /tmp or /home that is not allowed.
```
Conan-Kudo: I think `that is not allowed` is redundant. Drop it.

rpmlint/descriptions/BinariesCheck.toml
```diff
@@ -155,3 +158,13 @@ This library doesn't have the executable bit set. Without this bit set,
 rpm for instance won't be able identify the file as a library and not
 generate dependencies or strip debug symbols from it.
 """
+library-not-linked-against-libc="""
+The library is not dynamically linked against libc.
+"""
+program-not-linked-against-libc="""
+The binary is not dynamically linked against libc.
+"""
+shared-lib-without-dependency-information="""
```
Conan-Kudo: `lib`->`library`

rpmlint/descriptions/BinariesCheck.toml
Conan-Kudo: Alternatively, it can be titled `no-version-in-soname`.

kstreitova: I don't think so. This 'no-soname' check tests the library for an occurrence of the SONAME entry in the dynamic section (readelf -d <library>). 
Maybe we should just extend the description to be more verbose (something like: "The library has no SONAME entry in its dynamic section.")? But I'm not sure if it will help users or if it confuses them even more.

Conan-Kudo: Hmm, we probably should have that check. That description isn't terrible, but it's not too helpful either (what's a dynamic section?).

kstreitova: Yep, it's too complicated for the user. I would vote for leaving it as is while the user can find more information in the wiki (e.g. https://fedoraproject.org/wiki/Common_Rpmlint_issues#no-soname)

kstreitova:
Ready to be merged.

scarabeusiv:
Yea I think this looks okay as is, if the description is misleading we can tweak it in followup PRs or if someone complains via opening a ticket.

Issue #125, opened 2018-02-17
RFE: accept also "$(dirname "$specfile")/.rpmlintrc" config file


praiskup:
Seems like only .rpmlintrc from home directory has been accepted so far, but it would be nice to be able to ignore rpmlint  warnings on per-package basis.

praiskup:
For security reasons, `.rpmlintignore` with list of regular expressions would be a better approach?

Conan-Kudo:
I don't want any dotfiles for this. They should be fully visible. My feeling is that it should be `<srcpkgname>.rpmlintrc`.

As for where the file is or its name, as far as I'm aware, rpmlint itself doesn't care. This is up to the tools that automatically run rpmlint.

praiskup:
Dotfile is perfectly fine with me... it is low priority file which not everyone is interested in;
same as pylintrc, .gitignore, etc.

Though it shouldn't be python "evaluated" file, since it is insecure.  One could get malicious
code executed without noticing.

scarabeusiv:
I wrote logic for loading old configs here https://github.com/rpm-software-management/rpmlint/pull/151/commits/699ec5e4981c6a4fa2f3e3a3d0af69fcaa54ec9e

Possible question is if the rpmlint should do this on its own or if we leave it as another config passed as -f. If changes to be done should go when doing issue#155.

scarabeusiv:
Hmm now one of the interesting thing is, if we run rpmlint on some list of items where should the found rpmlint files apply.

`path/file1.rpm`
`path/file1-rpmlintrc`
`path2/file2.rpm`

In the above scenario should the rpmlintrc be applied on the `path2/file2.rpm`?

Also alternatively:

`path/file1.rpm`
`path/file1-rpmlintrc`
`path/file2.rpm`
`path/file3.rpm`
`path/file3-rpmlintrc`

Should the rpmlints in the above merged and applied for everything?

Conan-Kudo:
I would expect we'd apply rules based on source package name rpmlintrc files. But if we were reading multiple files, I guess it would merge them as you say.

praiskup:
Well, I'm repeating myself :-) but I still think `.rpmlintrc` file near the spec file (in dist-git it would be stored along the `.gitignore` file) is better idea.  Rpmlint can then simply look into the directory where where the linted file is (be that spec or srpm file), and check for dotfile.

Anything else sounds a bit overcomplicated to me.

Conan-Kudo:
[...]

PR #364, opened 2020-01-23
Implement rpmlintrc loading for issue 125

scarabeusiv:
This will take some time to complete, nothing much to see atm.

### DIFF
--- a/rpmlint/lint.py
+++ b/rpmlint/lint.py
@@ -24,8 +24,7 @@ class Lint(object):
             self.config = Config(options['config'])
         else:
             self.config = Config()
-        if options['rpmlintrc']:
-            self.config.load_rpmlintrc(options['rpmlintrc'])
+        self._load_rpmlintrc()
         if options['verbose']:
             self.config.info = options['verbose']
         if options['strict']:
@@ -80,6 +79,31 @@ class Lint(object):
             else:
                 print_warning(f'(none): E: there is no installed rpm "{name}".')
         return existing_packages
+
+    def _load_rpmlintrc(self):
+        """
+        Load rpmlintrc from argument or load up from folder
+        """
+        if self.options['rpmlintrc']:
+            self.config.load_rpmlintrc(self.options['rpmlintrc'])
+        else:
+            # load only from the same folder specname.rpmlintrc or specname-rpmlintrc
+            # do this only in a case where there is one folder parameter or one file
+            # to avoid multiple folders handling
+            rpmlintrc = []
+            if not len(self.options['rpmfile']) == 1:
+                return
+            pkg = self.options['rpmfile'][0]
+            if pkg.is_file():
+                pkg = pkg.parent
+            rpmlintrc += sorted(pkg.glob('*.rpmlintrc'))
+            rpmlintrc += sorted(pkg.glob('*-rpmlintrc'))
+            if len(rpmlintrc) > 1:
+                # multiple rpmlintrcs are highly undesirable
+                print_warning('There are multiple items to be loaded for rpmlintrc, ignoring them: {}.'.format(' '.join(map(str, rpmlintrc))))
+            elif len(rpmlintrc) == 1:
+                self.options['rpmlintrc'] = rpmlintrc[0]
+                self.config.load_rpmlintrc(rpmlintrc[0])
 
     def _print_header(self):
         """
--- a/test/rpmlintrc/multiple/sample.rpmlintrc
+++ b/test/rpmlintrc/multiple/sample.rpmlintrc
@@ -1,0 +1,1 @@
+setBadness('suse-dbus-unauthorized-service', 0)
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ dist/
 .eggs
 *.egg-info
 .coverage
+.coverage.*
 *.pyo
 *.pyc
 *.flog
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
 
 script:
   - podman build -t $DIST -f test/Dockerfile-$DIST .
-  - podman run -v $(pwd):/usr/src/rpmlint/ $DIST python3 setup.py test
+  - podman run -v $(pwd):/usr/src/rpmlint/ $DIST python3 -m pytest
   - podman run -v $(pwd):/usr/src/rpmlint/ $DIST flake8
 
 after_success:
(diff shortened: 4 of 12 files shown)


rpmlint/lint.py
```diff
@@ -81,6 +80,36 @@ def _load_installed_rpms(self, packages):
...
+            # to avoid multiple folders handling
+            rpmlintrc = []
+            if not len(self.options['rpmfile']) == 1:
+                return
+            pkg = self.options['rpmfile'][0]
+            if pkg.is_dir():
+                files = pkg.iterdir()
+            if pkg.is_file():
+                files = pkg.parent.iterdir()
+            for item in files:
```
mimi1vx: https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob i think will be better

rpmlint/lint.py
```diff
@@ -81,6 +80,36 @@ def _load_installed_rpms(self, packages):
...
+            # load only from the same folder specname.rpmlintrc or specname-rpmlintrc
+            # do this only in a case where there is one folder parameter or one file
+            # to avoid multiple folders handling
+            rpmlintrc = []
+            if not len(self.options['rpmfile']) == 1:
+                return
+            pkg = self.options['rpmfile'][0]
+            if pkg.is_dir():
+                files = pkg.iterdir()
+            if pkg.is_file():
```
mimi1vx: `elif` should be better

scarabeusiv: ok

rpmlint/lint.py
scarabeusiv: done

rpmlint/lint.py
```diff
@@ -81,6 +80,31 @@ def _load_installed_rpms(self, packages):
...
+            # do this only in a case where there is one folder parameter or one file
+            # to avoid multiple folders handling
+            rpmlintrc = []
+            if not len(self.options['rpmfile']) == 1:
+                return
+            pkg = self.options['rpmfile'][0]
+            if pkg.is_file():
+                pkg = pkg.parent
+            rpmlintrc += sorted(pkg.glob('*.rpmlintrc'))
+            rpmlintrc += sorted(pkg.glob('*-rpmlintrc'))
```
Conan-Kudo: I would have liked a deprecation notice for any `rpmlintrc` files using `*-rpmlintrc`...

test/rpmlintrc/multiple/sample.rpmlintrc
```diff
@@ -0,0 +1 @@
+setBadness('suse-dbus-unauthorized-service', 0)
```
Conan-Kudo: Is this a valid test? We don't have this lint rule in master...

Conan-Kudo:
@scarabeusiv @mimi1vx Please give me a chance to review these things too... 😕

test/rpmlintrc/multiple/sample.rpmlintrc
scarabeusiv: It could've been just empty file, I copied just parsable line thats it.

rpmlint/lint.py
scarabeusiv: scarabeus@bugaboo: ~/tmp/openSUSE:Factory $ grep '\-rpmlintrc' *.spec  |wc -l
793
scarabeus@bugaboo: ~/tmp/openSUSE:Factory $ grep '\.rpmlintrc' *.spec  |wc -l
67
The one with dash is winning by a huge margin.

Conan-Kudo: So? That doesn't make it right. It just means we have to support it while we tell everyone to fix their files.

scarabeusiv: And what is the 'right' approach? Your feeling? SUSE is at the moment the only vendor doing this stuff and by majority it seems to be accepted that bla-rpmlintrc is the way how to denote them.
Yet you find .rpmlintrc more to your liking so we should force everyone to rename the files?

DimStar77: > So? That doesn't make it right. It just means we have to support it while we tell everyone to fix their files.

it makes it right according to the openSUSE packaging guidelines

https://en.opensuse.org/openSUSE:Packaging_checks


we should simply ignore .rpmlintrc and consider them, according to the guidelines, as inexisting for the rpmlint checker)

Conan-Kudo: > And what is the 'right' approach? Your feeling? SUSE is at the moment the only vendor doing this stuff and by majority it seems to be accepted that bla-rpmlintrc is the way how to denote them.

Geez man, what the hell? Firstly, outside of the SUSE bubble, `<pkg>.rpmlintrc` is way more common. Secondly, in Fedora's infrastructure, you can use `.rpmlintrc` as a dot file for a package in Dist-Git. I know that this doesn't work in OBS VCS because its broken and can't handle dotfiles, but it's a thing. Thirdly, It's a bloody file extension to make it obvious that it's a file type "rpmlint runtime configuration".

It's not just a bloody feeling, it's based on the fact that it makes sense! Hell, before now, rpmlint didn't care at all what it was named, we just told it what to look up. You _changed_ that behavior without giving me a chance to respond until after the fact.

scarabeusiv: I didn't change this approach, you can still denote it by your liking, but I added feautre to be able to process it based on the directory files if nothing is specified.
As we discussed in the ticket half a year agon we agreed to have -rpmlintrc and .rpmlintrc which will now work. It will parse even '.rpmlintrc' file.
So what actual problem your warning on -rpmlintrc will solve? Except you don't personaly like it?

Conan-Kudo: I didn't specifically agree to `<pkg>-rpmlintrc` in #125. But whatever.

I just wanted to compromise and give people time to migrate to a single predictable file name (`<srcpkgname>.rpmlintrc`) so that there'd be no chance of collisions. Allowing loading multiple `rpmlintrc` files can lead to weirdness, and the best way to eliminate that is to just have a single file we load per source package (+collection of binary packages).

scarabeusiv: Actually loading multiple rpmlintrc files will not work ever, because what you want to shadow on one package will be needed on another.
Thats why I just make sure we found only one or warn otherwise, the same is that I don't look for them if multiple packages/folders are specified we don't look for them.

If you wan't to stricten the loader to be always `<srcpkg>.rpmlintrc` then be my guest but that should be another step where all the distros that are using the directory loading approach should update their policies.

PR #366, opened 2020-02-10
Cleanup pkg.py

mimi1vx:


rpmlint/pkg.py
```diff
@@ -511,25 +511,25 @@ def _extract(self):
             print_warning('Unable to access dir %s' % self.dirname)
             return None
```
scarabeusiv: There is return cmd bit down removed so this one should not return either.

scarabeusiv:
@mimi1vx https://lgtm.com/projects/g/rpm-software-management/rpmlint/rev/pr-d440e4457c1b92c8077be31a8b33a1df3cce51ae

scarabeusiv:
./rpmlint/pkg.py:522:13: F841 local variable 'cmd' is assigned to but never used

mimi1vx:
> ./rpmlint/pkg.py:522:13: F841 local variable 'cmd' is assigned to but never used

in progress with removal of `getstatusoutput`

PR #367, opened 2020-02-10
Flake8 by pytest

scarabeusiv:
To avoid people simply not running flake and failing in travis.


Conan-Kudo:
@scarabeusiv wouldn't this force flake8 runs if the unit tests were run in a package build?

scarabeusiv:
@Conan-Kudo same like we force coverage run there. The only way is for packagers to delete the line or add the deps.
But since everyone forgets to run the flake this will get them to use it.

Conan-Kudo:
Fair point.

PR #368, opened 2020-02-11
[opensuse] Add rc* compat simlink checking wrt #227

scarabeusiv:
Implements SUSE specific CheckRCLinks.py part by expanding
SysVInitOnSystemdCheck upstream check.

### DIFF
--- a/rpmlint/checks/SysVInitOnSystemdCheck.py
+++ b/rpmlint/checks/SysVInitOnSystemdCheck.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+import stat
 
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
@scarabeusiv As an aside, not a fan of it being an error... One day I'd like to see rcFOO stuff phased out in openSUSE in favor of just teaching everyone to use standard commands with `systemctl(1)`.

scarabeusiv:
I agree that this check is pointless. But as long as the policy wants it and we can safely validate it we should do so (we did even before, but with few bugs, so now it should be better).
But again, first the policy should be changed, than rpmlint can drop the check part.

PR #369, opened 2020-02-11
[opensuse] Rebase update-alternatives check for new rpmlint

scarabeusiv:
Redo the update-alternatives handling upon new rpmlint approach.

We are talking about https://github.com/openSUSE/rpmlint-checks/blob/master/CheckUpdateAlternatives.py

Still needs testcase/review of functionality prior merging.

### DIFF
--- a/rpmlint/checks/BashismsCheck.py
+++ b/rpmlint/checks/BashismsCheck.py
@@ -18,13 +18,11 @@ class BashismsCheck(AbstractFilesCheck):
             return
 
         try:
-            r = subprocess.run(['dash', '-n', fullpath],
-                               stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+            r = subprocess.run(['dash', '-n', fullpath])
             if r.returncode == 2:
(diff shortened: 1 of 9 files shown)


Conan-Kudo:
Hmm, something is tingling here... @decathorpe, do we have any guidelines specifically about alternatives in Fedora? This might be useful as a more generic check to implement policy about how alternatives should be used...

scarabeusiv:
We use the update-alternatives from dpkg that behave slightly differently from the Fedora version of this thing.

Conan-Kudo:
@scarabeusiv blech, really? ☹️

scarabeusiv:
Well looking on https://docs.fedoraproject.org/en-US/packaging-guidelines/Alternatives/ if it is spot on the requirements look the same so apart from it being different implementation the results delivered should be the same?

Conan-Kudo:
@scarabeusiv I'd think so, yes, at least from that perspective.

scarabeusiv:
Ok I will close this PR and create new one to point to master rather than opensuse branch and we can check it up. in my view i still need to do tests anyway :)

PR #370, opened 2020-02-11
Add Makefile error to FilesCheck

scarabeusiv:
This is the only really relevant check from CheckCommonFiles.py
that make sense to be ported.

We check for Makefile(am/in/) in the installed folder and warn if it is present.
We should probably check for CMakelists.txt in the same manner.

Needs a testcase prior merging.


rpmlint/checks/FilesCheck.py
```diff
@@ -554,6 +555,8 @@ def check(self, pkg):
                 self.output.add_info('W', pkg, 'manifest-in-perl-module', f)
             elif f == '/usr/info/dir' or f == '/usr/share/info/dir':
                 self.output.add_info('E', pkg, 'info-dir-file', f)
+            elif makefile_regex.search(f):
+                self.output.add_info('E', pkg, 'makefile-junk', f)
```
Conan-Kudo: I'm not sure this makes sense as an error. For example, the SELinux policy package ships Makefiles as `-devel` content, and there are Makefile-based build systems and tools (blech). I'd downgrade this to a warning.

scarabeusiv: Does the selinux put it in some determined location? If by design we could whitelist it and error out on everything else it is probably better.

Conan-Kudo: There are various Makefiles in `/usr/share/selinux/devel` and subdirectories.

scarabeusiv: So just excluding /usr/share/selinux would be golden right?

Conan-Kudo: Probably.

scarabeusiv: Added there, now just the tests.

PR #372, opened 2020-02-11
Rebase update-alternatives check for new rpmlint

scarabeusiv:
Followup of #369 as it should be distro agnostic.

Still needs tests before merging and actually some code cleanup&checking.


rpmlint/checks/UpdateAlternativesCheck.py
```diff
@@ -0,0 +1,79 @@
...
+
+import rpm
+from rpmlint.checks.AbstractCheck import AbstractCheck
+from rpmlint.pkg import b2s
+
+
+class UpdateAlternativesCheck(AbstractCheck):
+    """
+    Check for compliance with SUSE's update-alternatives guidelines:
+      http://en.opensuse.org/openSUSE:Packaging_Multiple_Version_guidelines
```
Conan-Kudo: Can we make this comment more generic?

rpmlint/descriptions/UpdateAlternativesCheck.toml
```diff
@@ -0,0 +1,17 @@
...
+not marked as ghost. Mark it as %ghost.
+"""
+alternative-link-missing="""
+The file %{_sysconfdir}/alternatives/$(basename generic-name) is missing
+in the file list. Mark it as %ghost and add it to the file list.
+"""
+suse-alternative-generic-name-missing="""
+The update-alternatives generic name is not in the filelist. Create it as
+a symlink to %{_sysconfdir}/alternatives/$(basename generic-name) and add it
+to the file list.
```
Conan-Kudo: We should drop the SUSE prefix here...

Conan-Kudo:
@scarabeusiv Can we squash these two into a single commit?

scarabeusiv:
I plan to squash it at the end. So i could see what i messed up and what was there from the begining.

Conan-Kudo:
@scarabeusiv Makes sense to me. 👍

rpmlint/descriptions/UpdateAlternativesCheck.toml
scarabeusiv: done, copy&pasto.

rpmlint/checks/UpdateAlternativesCheck.py
scarabeusiv: Adjusted it somehow, still i think we should rather point to the docs than describe it again in here? Or should i copy the wiki content there?

Conan-Kudo: If it's not too much content, perhaps inlining it makes sense? I'm not sure.

scarabeusiv:
After I wrote down what the checker should do it seems there is quite huge room for improvement to actually have it do what it is supposed...

rpmlint/checks/UpdateAlternativesCheck.py
scarabeusiv: added there.

Conan-Kudo:
@scarabeusiv Let's rename this to `AlternativesCheck`, as that's a better name for this.

scarabeusiv:
@Conan-Kudo okay.

scarabeusiv:
And in we go. I should probably mark the leap build as expected fail as it will take us few weeks to get that in.

PR #374, opened 2020-02-11
Add 'comprehensions' and 'builtins' flake8 tests and fix reported issues

mimi1vx:


rpmlint/spellcheck.py
```diff
@@ -59,7 +59,7 @@ def spell_check(self, text, fmt, lang='en_US', pkgname='', ignored_words=None):
         ignored_words: words to be ignored by the spellchecker
         """
         warned = set()
-        suggestions = dict()
```
scarabeusiv: set() stays?

mimi1vx: ja, because `{} == dict`,  set() hasn't empty representation

scarabeusiv:
lgtm

PR #376, opened 2020-02-12
Implement ErlangCheck from upstream

scarabeusiv:
Checks if the erlang was properly compiled with debug information and if not just pre-generated erlang files are shipped.


scarabeusiv:
@Conan-Kudo opinion? The CI is green, rawhide hickupped...

Conan-Kudo:
`pybeam` is broken, see [RH#1796224](https://bugzilla.redhat.com/show_bug.cgi?id=1796224).

This looks good otherwise, so I'm okay with merging this.

PR #380, opened 2020-02-13
Refactored MenuXDGCheck (#225)

MeggyCal:


rpmlint/checks/MenuXDGCheck.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, config, output):
         # desktop file need to be in $XDG_DATA_DIRS
         # $ echo $XDG_DATA_DIRS/applications
         # /var/lib/menu-xdg:/usr/share
-        super().__init__(config, output, r'(?:/usr|/etc/opt|/opt/.*)/share/applications/.*\.desktop$')
+        super().__init__(config, output, r'(?:/usr/share)/applications/.*\.desktop$')
```
scarabeusiv: Why instead of '(?:/usr/share)/applications/' ust not use '/usr/share/applications/'

rpmlint/checks/MenuXDGCheck.py
```diff
@@ -17,51 +17,33 @@
...
-                self.output.add_info('E', pkg, 'desktopfile-duplicate-option', filename,
-                                     '[%s]/%s' % (e.section, e.option))
-            else:
-                self.output.add_info('W', pkg, 'invalid-desktopfile', filename,
-                                     e.message.partition(':')[0])
+            self._handle_parser_error(pkg, filename, e)
         except UnicodeDecodeError as e:
+            if not is_utf8(f):  # not optimal, but working
+                self.output.add_info('E', pkg, 'non-utf8-desktopfile', filename)
             self.output.add_info('W', pkg, 'invalid-desktopfile', filename, 'Unicode error: %s' % (e))
```
scarabeusiv: Why double check, if one tool reports unicod decoding then just throw out unicode parsing error and be done with it.

rpmlint/checks/MenuXDGCheck.py
```diff
@@ -77,7 +59,43 @@ def check_file(self, pkg, filename):
...
+        """
+        # I would love to use switch, however, each warning is printed differently
+        if (isinstance(e, cfgparser.MissingSectionHeaderError)):
+            self.output.add_info('E', pkg, 'desktopfile-missing-header', filename)
+        elif (isinstance(e, cfgparser.DuplicateSectionError)):
+            self.output.add_info('E', pkg, 'desktopfile-duplicate-section', filename,
+                                 '[%s]' % e.section)
+        elif (isinstance(e, cfgparser.DuplicateOptionError)):
+            self.output.add_info('E', pkg, 'desktopfile-duplicate-option', filename,
+                                 '[%s]/%s' % (e.section, e.option))
```
scarabeusiv: please use here and on other above/bellow the normal python formated string: `f'broken stuff {variable}'`

rpmlint/checks/MenuXDGCheck.py
```diff
@@ -77,7 +59,43 @@ def check_file(self, pkg, filename):
...
+    def _has_binary(self, root, cfp, binary):
+        """
+        Check whether there is a binarry assigned to the desktop file.
+
+        Needs configparser instance, it is assumed to be called in parse_desktop_file.
+        """
+        if cfp.has_option('Desktop Entry', 'Exec'):
+            binary.append(cfp.get('Desktop Entry', 'Exec').partition(' ')[0])
+        # If there is no binary mentioned it is OK
+        if not binary[0]:
```
scarabeusiv: why are you checking just binary0, if it uses multiple binaries should't it check more? or what is the iterator for?

rpmlint/checks/MenuXDGCheck.py
MeggyCal: because I am confused :)

rpmlint/checks/MenuXDGCheck.py
MeggyCal: well, yeah, this is uggly...

PR #383, opened 2020-02-14
Fix --rpmlintrc "user specified rpmlintrc" check

lbt:
The warning should be issued if *not* Path.exists()


scarabeusiv:
Thanks :)

PR #384, opened 2020-02-14
Add option to only return error when badness threshold is exceeded

Conan-Kudo:
When rpmlint is configured for mandatory tests in package builds, it is desirable to allow rpmlint to "pass" the build unless the the badness threshold has been exceeded, even with errors.

This option permits this behavior for this circumstance.

### DIFF
--- a/rpmlint/__init__.py
+++ b/rpmlint/__init__.py
@@ -87,7 +87,9 @@ def process_lint_args(argv):
     parser.add_argument('-V', '--verbose', action='store_true', help='provide detailed explanations where available')
     parser.add_argument('-p', '--print-config', action='store_true', help='print the settings that are in effect when using the rpmlint')
     parser.add_argument('-i', '--installed', nargs='+', default='', help='installed packages to be validated by rpmlint')
-    parser.add_argument('-s', '--strict', action='store_true', help='treat all messages as errors')
+    lint_modes_parser = parser.add_mutually_exclusive_group()
+    lint_modes_parser.add_argument('-s', '--strict', action='store_true', help='treat all messages as errors')
+    lint_modes_parser.add_argument('-P', '--permissive', action='store_true', help='treat individual errors as non-fatal')
 
     # print help if there is no argument
     if len(argv) < 1:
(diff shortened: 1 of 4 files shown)


scarabeusiv:
This still does not make sense:

User configured treshold in config is compared first. If that is not specified then it goes to check errors and if there are any it errors out.

With your commit if it finds any errors it simply will not error out.

Or is this supposed to be inverse option to strict where we treat all messages as errors?

Conan-Kudo:
@scarabeusiv Yes, I think this is basically the inverse option, let me restructure that accordingly.

scarabeusiv:
I checked the code of previous solver and the logic now is:

If badness treshold defined:
 codeblock
Else if errors found


But previously it was:
If badness treshold defined:
  do stuff
If errors found:
  do other stuff

scarabeusiv:
So all in all I think the option makes sense, but I would call it something sensible like '--permissive' 'Do not error out even if there were errors located in the file'...

Conan-Kudo:
@scarabeusiv That's probably a much better option name. I kind of struggled figuring out what to name it 😄

rpmlint/__init__.py
```diff
@@ -87,7 +87,9 @@ def process_lint_args(argv):
     parser.add_argument('-V', '--verbose', action='store_true', help='provide detailed explanations where available')
     parser.add_argument('-p', '--print-config', action='store_true', help='print the settings that are in effect when using the rpmlint')
     parser.add_argument('-i', '--installed', nargs='+', default='', help='installed packages to be validated by rpmlint')
-    parser.add_argument('-s', '--strict', action='store_true', help='treat all messages as errors')
+    lint_modes_parser = parser.add_mutually_exclusive_group()
+    lint_modes_parser.add_argument('-s', '--strict', action='store_true', help='treat all messages as errors')
+    lint_modes_parser.add_argument('-P', '--permissive', action='store_true', help='error only if badness threshold is exceeded')
```
scarabeusiv: Not sure if just not say 'treat all errors as non-fatal and do not fail' or something along the way.

Conan-Kudo: Sure, makes sense.

scarabeusiv:
@Conan-Kudo for the tests you need to add it to test/test_lint.py options_preset dict.

Conan-Kudo:
@scarabeusiv Thanks, done!

PR #381, opened 2020-02-13
Introduce Packit CI to enable easy local testing of PRs

Conan-Kudo:
[Packit](https://packit.dev/) will generate pull-request specific package builds for Fedora Rawhide, Mageia Cauldron, and openSUSE Tumbleweed to help make it easier to do local testing of pull requests as part of reviewing proposed changes.

PR #386, opened 2020-02-14
Add dependency on RPM Python bindings

Conan-Kudo:
RPMLint heavily leverages the RPM Python bindings in order to function properly. As those bindings are shipped with Python metadata, we can and should declare the dependency properly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,7 @@ setup(
 
     install_requires=[
         'pyxdg',
+        'rpm',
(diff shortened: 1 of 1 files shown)


scarabeusiv:
Mhm, As the python3-rpm provides egginfo it should be fine :)

Conan-Kudo:
@scarabeusiv I'd discovered this problem when working on #381. 💯

PR #389, opened 2020-02-14
Remove NamingPolicyCheck

scarabeusiv:
As described in the PR it is quite pointless and actively harmful.

If someone wants to codify the naming sequences it coudl be reintroduced but at the moment 80% of the reported content is actually wrong.

Also as a trivial change disable InitCheck by default (it can still be enabled by those using them).


Conan-Kudo:
Perhaps in the future we should reintroduce it, but yeah, I agree...

PR #390, opened 2020-02-14
Remove compat for old rpms from SpecCheck

scarabeusiv:
The opening error was fixed in rpm since 2011 based on the bug and
we can reduce bit the conditions as rpm < 4.8 is of no interest to
us anyway.


Conan-Kudo:
@scarabeusiv If you can take care of the LGTM notice, then it'll be good to merge.

Conan-Kudo:
Meh, who cares...? Merging.

PR #394, opened 2020-02-17
Move the rpmlint CLI logic into its own module file

Conan-Kudo:
Having the CLI processing logic in the init module file made it
get evaluated on every single rpmlint import. This is completely
unnecessary and causes an import loop in setuptools, introducing
dependencies where there should not be when trying to "build" the
rpmlint code with setuptools.

This should also speed up imports across the board in rpmlint, since
this code only needs to be evaluated once.


Conan-Kudo:
@scarabeusiv I'm confused at this failure in Travis CI, any idea what it means?

Conan-Kudo:
@scarabeusiv Nevermind, I think I figured it out. :)

Issue #341, opened 2019-09-20
Add a check for consistent OPTFLAGS usage


msmeissn:
At SUSE we store the used RPM_OPT_FLAGS / %optflags in a ELF section in the .o files.

A check that verifies / compares the used options with the wanted options would help to identify
files not built with RPM_OPT_FLAGS.

This would spot files compiled without stack-protector, fortify, pointer protections or warnings etc etc

Conan-Kudo:
This is also done using `annobin` on RH/Fedora, but I'm not sure how to implement such a check across the board. I'm sure everyone is using different mechanisms to record this data...

msmeissn:
perhaps it could be standardized in some form?

Conan-Kudo:
I'd be happy if it was.

All I know about it is from these pages:
* https://fedoraproject.org/wiki/Changes/Annobin
* https://developers.redhat.com/blog/2018/02/20/annobin-storing-information-binaries/
* https://developers.redhat.com/blog/2019/02/04/annocheck-examining-the-contents-of-binary-files/

Does the SUSE solution look anything like this?

scarabeusiv:
it is stored in the debuginfo for -record-gcc-switches option that is enabled by default. So we will write extractor for that  if that is present it should be stored there even for fedora.

The major problem for the fedora is that the project wanted a way to check it even for non-debuginfo packages which would need to be done by gcc plugin.

Conan-Kudo:
@scarabeusiv Might it make sense for SUSE to also adopt the `annobin` approach?

Conan-Kudo:
@scarabeusiv @msmeissn Might it make sense for SUSE to also adopt the `annobin` approach?

scarabeusiv:
why the debiginfo pkgs contain the data even on fedora, we can just scan that and be done with it, it will just not work on the general rpms which is what the fedora tooling alows

msmeissn:
The compiler writes the optflags into the elf section as I said.

Our scanner so far was just grepping over the buildlog, and did not use the optflags section.

Currently our GCC builds do not support GCC plugins.

marxin:
> it is stored in the debuginfo for -record-gcc-switches option that is enabled by default. So we will write extractor for that if that is present it should be stored there even for fedora.

I tested the `-frecord-gcc-switches` and it's problematic with LTO where we have merged compilation units and so that we'll miss all original compilation unit flags.

[...]

PR #397, opened 2020-03-05
Add new {missing-mandatory,forbidden}-optflags checks (#341). 

marxin:
The check is based on parsing of debug info and
-grecord-gcc-switches compiler option.

### DIFF
--- a/rpmlint/checks/BinariesCheck.py
+++ b/rpmlint/checks/BinariesCheck.py
@@ -6,6 +6,7 @@ import stat
 import rpm
 from rpmlint.checks.AbstractCheck import AbstractCheck
 from rpmlint.lddparser import LddParser
+from rpmlint.objdumpparser import ObjdumpParser
 from rpmlint.readelfparser import ReadelfParser
 from rpmlint.stringsparser import StringsParser
 
@@ -56,7 +57,8 @@ class BinariesCheck(AbstractCheck):
                                 self._check_rpath,
                                 self._check_library_dependency,
                                 self._check_forbidden_functions,
-                                self._check_executable_shlib]
+                                self._check_executable_shlib,
+                                self._check_optflags]
 
     @staticmethod
     def create_nonlibc_regexp_call(call):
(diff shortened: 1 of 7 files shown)


rpmlint/checks/BinariesCheck.py
```diff
@@ -466,6 +478,12 @@ def run_elf_checks(self, pkg, pkgfile_path, path):
                 self.output.add_info('E', pkg, 'ldd-failed', path, failed_reason)
                 return
 
+            self.objdump_parser = ObjdumpParser(pkgfile_path, path)
+            failed_reason = self.objdump_parser.parsing_failed_reason
+            if failed_reason:
+                self.output.add_info('E', pkg, 'objdump-failed', path, failed_reason)
```
scarabeusiv: This seem not to have description added.

rpmlint/checks/BinariesCheck.py
```diff
@@ -452,6 +454,16 @@ def _check_executable_shlib(self, pkg, pkgfile_path, path):
             if interp:
                 self.output.add_info('E', pkg, 'shared-library-not-executable', path)
 
+    def _check_optflags(self, pkg, pkgfile_path, path):
+        if not self.readelf_parser.is_archive:
+            mandatory_optflags = self.config.configuration['MandatoryOptflags']
+            if mandatory_optflags:
```
scarabeusiv: This could be less nested preferably.

rpmlint/checks/BinariesCheck.py
```diff
@@ -452,6 +454,16 @@ def _check_executable_shlib(self, pkg, pkgfile_path, path):
...
+                for dwarf_unit in self.objdump_parser.compile_units:
+                    tokens = dwarf_unit['producer'].split(' ')
+                    missing = [mo for mo in mandatory_optflags if mo not in tokens]
+                    if missing:
+                        self.output.add_info('W', pkg, 'missing-mandatory-optflags', path, ' '.join(missing))
```
scarabeusiv: Should even be error I dare to say if it does not do too many false positives.

rpmlint/checks/BinariesCheck.py
```diff
@@ -452,6 +454,16 @@ def _check_executable_shlib(self, pkg, pkgfile_path, path):
             if interp:
                 self.output.add_info('E', pkg, 'shared-library-not-executable', path)
 
+    def _check_optflags(self, pkg, pkgfile_path, path):
+        if not self.readelf_parser.is_archive:
+            mandatory_optflags = self.config.configuration['MandatoryOptflags']
```
scarabeusiv: There should also be list of forbidden flags we could define and inform about.

rpmlint/checks/BinariesCheck.py
marxin: I would wait with making it an error, we can do it later.

Conan-Kudo:
Will this work with annobin-generated information, too?

Conan-Kudo:
@nickclifton Could you please review this to see if `annobin` would be covered sufficiently?

marxin:
> Will this work with annobin-generated information, too?

No. I don't know where exactly `annobin` saves command line options.
Note that there's one another possible solution that can remove need
of the `annobin` plugin:
https://gcc.gnu.org/ml/gcc-patches/2020-03/msg00294.html

rpmlint/checks/BinariesCheck.py
```diff
@@ -452,6 +454,24 @@ def _check_executable_shlib(self, pkg, pkgfile_path, path):
...
+            return
+
+        for dwarf_unit in self.objdump_parser.compile_units:
+            tokens = dwarf_unit['producer'].split(' ')
+            missing = [mo for mo in mandatory_optflags if mo not in tokens]
+            forbidden = [f for f in forbidden_optflags if f in tokens]
+            if missing:
+                self.output.add_info('W', pkg, 'missing-mandatory-optflags', path, ' '.join(missing))
+            if forbidden:
+                self.output.add_info('W', pkg, 'forbidden-optflags', path, ' '.join(forbidden))
```
scarabeusiv: This can be error for sure, if you set out to detect optflag and go you way to set to detect it you should explode if it is not allowed.

rpmlint/descriptions/BinariesCheck.toml
```diff
@@ -168,3 +171,9 @@ shared-library-without-dependency-information="""
 The listed shared library doesn't include information about which other
 libraries the library was linked against.
 """
+missing-mandatory-optflags="""
+This executable was not compiled with $Optflags.
```
scarabeusiv: "This executable was not compiled with expected flag." instead for readability?

PR #402, opened 2020-03-14
Rename CheckForXinetd to XinetdDepCheck

### LINKED ISSUE #399: Rename CheckForXinetd
We name all the checks as blablaCheck and this one is inverse.

For similarity we should simply rename it.

Conan-Kudo:
This more accurately represents what the check contains
and follows the same naming convention as the rest of the
checks.

Fixes #399


scarabeusiv:
Yarp :+1:

PR #403, opened 2020-03-14
Check for debugsource subpackages by default

Conan-Kudo:
Fedora, openSUSE, Mageia, and most others have these, so it's a
reasonable default.


Conan-Kudo:
@scarabeusiv I'm starting to review the settings as part of preparing a change proposal to move to rpmlint 2.0 for Fedora 33.

PR #405, opened 2020-03-14
Reduce binary packages size

scarabeusiv:


scarabeusiv:
Merging as it is mostly noop.

PR #406, opened 2020-03-17
[opensuse] KMP check

scarabeusiv:
So far just quickly ported, I need to review what it does and add testcases.


rpmlint/checks/KMPPolicyCheck.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class KMPPolicyCheck(AbstractCheck):
+    def check(self, pkg):
+        if pkg.isSource() or pkg.name.find('-kmp-') < 0:
+            return
+
+        pkg_requires = set(map(lambda x: x[0].split('(')[0], pkg.requires()))
+
+        kernel_flavour = "kernel-" + pkg.name.partition('-kmp-')[2]
```
Conan-Kudo: Please consistently use `flavor` here (American English is used throughout rpmlint)

scarabeusiv: sure.

Conan-Kudo:
@scarabeusiv Why drop the `suse-` prefix? This is specific to SUSE kernel module packaging, since SUSE doesn't follow the standard kmod packaging convention.

scarabeusiv:
Because it is in our branch anyway and we don't prefix the others either as it eats quite bit space on the rpmlint output line.

Issue #371, opened 2020-02-11
[opensuse] review and port remainder of the rpmlint-checks that are SUSE specific to 2.0 syntax


scarabeusiv:
When going over the review of what to port and review there still remain following items (counting open PRs as "done"):

openSUSE:
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/KMPPolicyCheck.py
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/BrandingPolicyCheck.py

Could be generalized, see https://github.com/rpm-software-management/rpmlint/blob/master/rpmlint/checks/PAMModulesCheck.py vs https://github.com/openSUSE/rpmlint-checks/blob/master/CheckPAMModules.py
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/CheckPolkitPrivs.py
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/CheckSUIDPermissions.py
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/CheckCronJobs.py
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/CheckDBUSServices.py

should be upstreamed:
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/LibraryPolicyCheck.py
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/CheckBuildRoot.py
- [x] https://github.com/openSUSE/rpmlint-checks/blob/master/ErlangCheck.py
- [x] https://github.com/openSUSE/rpmlint-checks/commits/master/MixedFileOwnerships.py

unless we manage to figure out a way how to make the test generic enough for the openSUSE ones they go only to opensuse branch, otherwise upstream first

scarabeusiv:
BuildRoot was already done, I overlooked that 1e2d6e3abfb7419cd081df79b15c926228f2f052

scarabeusiv:
KMPPolicyCheck is now WIP by me.

scarabeusiv:
@Conan-Kudo what would be your opinion of https://github.com/openSUSE/rpmlint-checks/pull/55/files#diff-736af2f9cac9ffa626fddbff2dc31c73 I would say this suseism errorer could be upstream, so people by accident don't copy and add this?
I would of course extend it to check all other variables too not just supplements.

Conan-Kudo:
@scarabeusiv Yeah, makes sense to me.

[...]

PR #413, opened 2020-03-28
Add test for SUSEisms that are not needed

scarabeusiv:
Found when checking #371

Implements https://github.com/openSUSE/rpmlint-checks/pull/55/files#diff-736af2f9cac9ffa626fddbff2dc31c73


scarabeusiv:
What could've happened to travis here...

scarabeusiv:
@Conan-Kudo anything wrong with the above? If it is okay for you i would merge it as the travis seem to be stuck and localy it works just fine.

rpmlint/checks/ZyppSyntaxCheck.py
```diff
@@ -0,0 +1,18 @@
+from rpmlint.checks.AbstractCheck import AbstractCheck
+
+
+class ZyppSyntaxCheck(AbstractCheck):
+    def check(self, pkg):
+        # We care only about the names, versions are pointless here
+        pkg_supplements = list(map(lambda x: x[0], pkg.supplements))
```
mimi1vx: I dont think map and lambda is needed `pkg_supplements = [x for x in pkg.supplements]` and for others same

PR #414, opened 2020-03-30
Use namedTuple for dependency info in Pkg

mimi1vx:


scarabeusiv:
Looks good, I would also swap the 'nice' usage in packages too so people won't keep using 0, 1, 2 keys forever.

scarabeusiv:
For some reason the travis does not report back about the success:
  https://travis-ci.org/github/rpm-software-management/rpmlint/builds/668635076?utm_medium=notification&utm_source=github_status

PR #416, opened 2020-03-30
Move version to __init__.py and avoid rpmlint as dependency for setup.py

mimi1vx:


Conan-Kudo:
@mimi1vx I do have a way to fix this, though. I'll make a pull request for it shortly.

mimi1vx:
> No. This causes import loops. That's why I got rid of this in the first place.

where ?

mimi1vx:
ok, Your solution

PR #417, opened 2020-03-30
setuptools: Remove remaining import loop in the code

Conan-Kudo:
This change eliminates the final import loop that made it difficult
to install rpmlint from PyPI, as now we have no implicit self dependencies
that turns all our runtime dependencies into build-time dependencies.


scarabeusiv:
travis still misbehaves.

PR #418, opened 2020-03-30
Try to get travis pass again

scarabeusiv:


rpmlint/version.py
```diff
@@ -1,6 +1,6 @@
 try:
-    from importlib.metadata import version as implib_metadata_version
+    from importlib.metadata import version
 except ImportError:
-    from importlib_metadata import version as implib_metadata_version
+    from importlib_metadata import version
 
-__version__ = implib_metadata_version('rpmlint')
+__version__ = version(__name__)
```
Conan-Kudo: Is it safe to use the `version()` name like this? And where does `__name__` come from?

scarabeusiv: Nope, i was just trying if it would resolve it.

scarabeusiv:
@Conan-Kudo bigger problem the importlib seem not to work without the egg-info available.

So it works if rpmlint is installed but i would have to catch another exception and set fake version if not...

scarabeusiv:
The error we see is basically the same like https://github.com/pyinstaller/pyinstaller/issues/4258

@mimi1vx do you happen to have any suggestions?

Conan-Kudo:
@scarabeusiv We probably should just have a fake version fallback of `0.0.0`.

scarabeusiv:
Ok added the fallback there.

scarabeusiv:
The importlib will be added on Leap 15 with pytest MU that is being now processed by QA.

PR #420, opened 2020-04-01
Add MixedOwnershipCheck #371

scarabeusiv:


scarabeusiv:
@Conan-Kudo anything against above being enabled by default?

Conan-Kudo:
@scarabeusiv No, this is an excellent generic check.

Conan-Kudo:
Bah, Travis still seems to be broken...

PR #421, opened 2020-04-02
WIP: Add SLP check #371

scarabeusiv:


Conan-Kudo:
@scarabeusiv Don't we already have a variation of this?

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,311 @@
...
+        # Verify non-lib stuff does not add dependencies
+        if libs:
+            for dep in pkg_requires.difference(_essential_dependencies):
+                if '.so.' in dep and dep not in libs and dep not in libs_needed:
+                    printError(pkg, 'shlib-policy-excessive-dependency', dep)
+
+        # Check for non-versioned directories beyond sysdirs in package
+        sysdirs = ('/lib', '/lib64', '/usr/lib', '/usr/lib64',
+                   '/usr/share', '/usr/share/licenses',
+                   '/usr/share/doc/packages')
```
Conan-Kudo, line 300: This should be `/usr/share/doc` instead of `/usr/share/doc/packages`

scarabeusiv:
Nope the SLP is not tested elsewhere. There is the weird %mklibname in speccheck.

In BinaryCheck we have test for the soname and if the library filename matches that only.

rpmlint/checks/SharedLibraryPolicyCheck.py
scarabeusiv: Yep, there is also quite bit of /opt and kde3 prefixes that must go away :)

Conan-Kudo:
`%mklibname` check is the Mandriva SLP enforcement package macro: https://wiki.mageia.org/en/Libraries_policy

scarabeusiv:
Finished up in #560

Issue #326, opened 2019-09-16
Tmpfiles.d %files rules


kstreitova:
In rpmlint we currently check if the tmpfile created via tmpfiles.d is listed in the %files section and we advise to list it as a %ghost.

However, in [openSUSE tmpfiles.d packaging wiki ](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Creating_files_and_subdirectories_in_.2Fvar.2Frun_and_.2Frun) is no mention about %ghost. The same applies for 
[Fedora](https://fedoraproject.org/wiki/Packaging:Tmpfiles.d#Example_spec_file) but it uses %verify(not size mtime md5) instead (for files).

So my question is what are the correct rules we want to check in rpm lint regarding tmpfiles.d and %files section.

Thanks!

Conan-Kudo:
@jasontibbitts @decathorpe Are the Fedora rules on that wiki page still correct? I see people using `%ghost` more than I do using the odd `%verify` stuff...

decathorpe:
@Conan-Kudo wow, I've never seen `%verify(foo)` anywhere, and didn't even know it existed ... I only know of `%ghost`, and it's what I would use in this case.

By the way, the authoritative source of these packaging guidelines is now [here](https://docs.fedoraproject.org/en-US/packaging-guidelines/Tmpfiles.d/), barring formatting issues that happened when converting them from the wiki page. But looking at the source for the `tmpfiles.d` page, it hasn't been updated or verified yet since the initial conversion from wiki markup to AsciiDoc.

Conan-Kudo:
@kstreitova @decathorpe I'm not sure `%ghost` is necessary for stuff in `/run` anymore, since it's more or less guaranteed to be tmpfs now. `%verify` is definitely overkill.

Conan-Kudo:
@kstreitova @decathorpe is [bringing this up to the Fedora Packaging Committee](https://pagure.io/packaging-committee/issue/927), so we'll have an answer soon.

kstreitova:
@Conan-Kudo Any news regarding creating a tmpfile macro that was mentioned in the Fedora meeting? Thanks!

Conan-Kudo:
@decathorpe has been entrusted to design Fedora's `tmpfiles.d` policy...

@decathorpe, have you figured out how it should work yet?

decathorpe:
@Conan-Kudo I didn't have much time to look into this yet, sorry.

[...]

PR #422, opened 2020-04-02
Tmpfiles %ghost

kstreitova:
We want rpmlint to check if a tmpfile is listed as a %ghost in the %files. In fact, it can't be there without %ghost directive (rpm build would fail) so it's enough to check if it's in the filelist. I added this fact to the docstring so it should be more clear now.


Conan-Kudo:
> In fact, it can't be there without %ghost directive (rpm build would fail)

Why do you think this is true?

kstreitova:
Because the tmpfile file doesn't exist in the time of the build. Therefore (without %ghost) it fails in the "Processing files" section with "File not found" RPM build error.

Conan-Kudo:
That assumes someone actually puts the file in the file list. Most of the time, they don't. Unless you have a brp that evaluates the tmpfiles and forces them to exist, which would be bad for the reasons you stated earlier.

scarabeusiv:
Yes and we agreed in that referenced ticket that the files created by systemd should be marked as %ghost to get them pruned on pkg uninstall, so what is the issue, this implements exactly what was the decision in #326 ...

Conan-Kudo:
I'm just confused how this would be enforced.

kstreitova:
@Conan-Kudo can you be more specific about what concerns you in this PR exactly, please?

Conan-Kudo:
@kstreitova Is this reading the tmpfiles listing and checking if they're also in the rpm files list?

PR #426, opened 2020-04-03
Make 'only-non-binary-in-usr-lib' an error

kstreitova:


scarabeusiv:
Technically speaking the systemd and others should be in the known whitelist.
Also I wonder if the check does not need rethink, as it is now it loads up content of /usr/lib and if for whole pkg there is nothing else there that is binary it warns, but otherwise if there is i.e. one library and rest is just some txt files it will keep silent...

scarabeusiv:
@kstreitova could you split the .gitignore update as that can go in so we can keep this one tracking just the nonbinary check?

marxin:
Closing as there's no activity. If needed, we can adjust badness in `openSUSE` configuration.

PR #428, opened 2020-04-08
Rewrite SignatureCheck and add tests

kstreitova:
Separate SignatureCheck to functions and add tests. Also, rewrite the
logic as SignatureCheck was obsolete and didn't work properly
(e.g. it searched for "MISSING KEYS" and "pgp|gpg" strings that are no
longer used in 'rpm -K' output).

Now it checks the output of 'rpm -Kv' which provides more detailed output.
Thanks to that we could add another check ('E: invalid-signature' that
checks if the signature is corrupted or not).

The test_signature.py file was added so now all 3 signature checks are
tested properly.


scarabeusiv:
LGTM

Issue #423, opened 2020-04-03
Python packages testing check


scarabeusiv:
With the openSUSE packages when looking on what we do we should do following checks on python packages:

1. verify the egg-info is folder (setuptools based) and not file (distutils) to make pip works
2. validate requires really match up the egg-info requires
3. move the forbidden dirs from https://github.com/rpm-software-management/rpmlint/blob/opensuse/rpmlint/checks/FilelistCheck.toml#L86 to the check and expand also on subfolders ; there should be no need to ship tests i.e. (except few known cases, so this should be warning)

Conan-Kudo:
I think number 2 is hopefully not going to be needed anymore once the dependency generator is fully enabled in openSUSE like it is in RH/Fedora and Mageia. In Fedora, Mageia, OpenMandriva, ROSA, etc., we don't *have* this problem.

I do not think we should codify deleting content from Python packages in rpmlint upstream. For a couple of reasons: there are enough valuable reasons to ship test code and we can't reasonably determine the test code is _just_ test code.

But number 1 would make openSUSE fail this check. 😉 There are a number of packages that have been deliberately forced from setuptools to distutils (which you are likely aware of that!), even though I think that's a bad idea...

scarabeusiv:
We will never use that dependency generator in openSUSE, we can't and won't do those cycles that it causes.

There are 2 cases:
1. tests are directly in site-pacakges/tests so all the packages conflicts, that we already check and error
2. tests in the proper subfolder, that can be warning as we should be running those tests arleady.

And we actually already fake the egg-info if needed, so it is more like to fix the packages that accidentaly didn't pull the setuptools or were overlooked to have the egg-info properly.

Conan-Kudo:
> We will never use that dependency generator in openSUSE, we can't and won't do those cycles that it causes.

It only causes those cycles because you _made_ it that way. That is _definitely_ fixable and you know it. And I'm looking at removing setuptools as a dependency for Python 3.8+ anyway. But you _need_ to fix this problem _openSUSE_ created with setuptools.

> tests are directly in site-pacakges/tests so all the packages conflicts, that we already check and error

This part I agree with. We should definitely have an error here.

darix:
This is also a problem because we do automatic rebuilds. while most other distros dont. so you dont see the cycles.

Conan-Kudo:
[...]

PR #430, opened 2020-05-07
Add PythonCheck (#423)

jgrassler:
This adds a Python specific check that performs 3 tasks:

1) Verify the egg-info is folder (setuptools based) and not file based
   (distutils style)

2) Check package's Requires against egg-info requires.txt to make sure the
   package pulls in everything specified by Upstream.

3) Check for Python specific forbidden dirs and output various warnings and errors about them.


scarabeusiv:
In general I would recommend usimg pathlib where possible.
There are unused imports. 
You should adjust the dockerfiles in the test folder adding the new dep.
The testing rpms were acked so you can just add the tests here too :-)

Conan-Kudo:
I also had said previously ***very firmly*** that ensuring that requires are present for Python match between setuptools data and rpm requires is out of scope for rpmlint.

Conan-Kudo:
Also note, while this isn't relevant for upstream rpmlint, the policy changes you've proposed here will force setuptools into the ring 0 in openSUSE Factory. You should just do that first.

hellcp:
This will cause setuptools to be required in ring0, so maybe it makes sense to enable deps generator in Factory, before you break it with this change ;)

jgrassler:
Thanks for the feedback! Some comments inline.

Conan-Kudo wrote:
> Please do not include checking whether dependencies exist between requirements.txt/setuptools and rpm requires. There are too many things that go wrong here:

I'm well aware of these two...

>     * No guarantee of name consistency or even matching (e.g. straight-plugin vs straight.plugin)
> 
>     * Naming conventions may obfuscate or break identification

...and that's why the check emits warnings and not errors. It's strictly an advisory kind of "Hey, there might be something off here, why don't you take a look and double-check the Requires are really shipshape". If the warnings turn out to be bogus they can be ignored, either through the moral equivalent of `grep -v W-python-egginfo-require-not-in-spec ` running in people's visual cortex or by adding a rpmlintrc for filtering the bogus warnings if one is a stickler for detail.

>     * Dependencies may be intentionally filtered out

Now that one is a problem, yes. Right now the check does go through there in seek-and-destroy mode ignoring any and all conditionals. That is going to turn up false positives. I was toying with the idea of adding some heuristics (such as ignoring `extra_requires` conditional on `win32`), but I opted against that in the first iteration - that's something that will require refinement over time anyway.

hellcp wrote:
> This will cause setuptools to be required in ring0, so maybe it makes sense to enable deps generator in Factory, before you break it with this change ;)

It will cause setuptools to be required by virtue of banning distutils based egg-info via you mean?

Anyway, both could be addressed by making this particular check configurable with the default being "off". Is that a compromise we all could live with? I'd hate to toss it out entirely because it does give me a quick and dirty read on what Requires a package should probably have - that _is_ useful.

hellcp:
> ...and that's why the check emits warnings and not errors. It's strictly an advisory kind of "Hey, there might be something off here, why don't you take a look and double-check the Requires are really shipshape". If the warnings turn out to be bogus they can be ignored, either through the moral equivalent of `grep -v W-python-egginfo-require-not-in-spec ` running in people's visual cortex or by adding a rpmlintrc for filtering the bogus warnings if one is a stickler for detail.

And also is the first check that can entirely fall outside of packager's capabilities to actually fix. Yes there are SUSE Security checks, which require a bug report, but try to report a bug for renaming a package to something else (especially something that contains an invalid character in the Name tag) and it will be instantly closed as RESOLVED WONTFIX. I also intentionally remove every rpmlintrc I ever see because I am a stickler for detail in every way, including this one.

jgrassler:
Alright. I guess I'll remove the requires check then. If anybody _does_ want to make the case for it, by all means do...the earliest I'll be able to get back to this is probably Thursday next week. For now we've got working tests for openSUSE but not Fedora since `requirements-parser` is not available there. If we do decide to retain that check after all, I'll have to modify to work with what's available everywhere (e.g. `python3-packaging`).

Setting to draft until next week for this is very much work in progress, still.

Conan-Kudo:
@jgrassler I'm happy to accept this into rpmlint once the requirements check is removed. The rest of it looks great, and given the pending removal of that code, I expect that the missing dependency won't be a problem.

Conan-Kudo:
@jgrassler As for `requirements-parser`, if it's required for other parts, then I'll be happy to get it packaged in Fedora. Just let me know...

jgrassler:
Alright, I removed the requirements check now. Ready for review/merge as far as I'm concerned. The Travis check fails due to what appears to be a problem with the opensuselp15 test suite. Not sure where exactly the problem is, but the same `INTERNALERROR> OSError: could not get source code` is thrown when running against unmodified `master`...

```
podman build -t opensuselp15 -f test/Dockerfile-opensuselp15 .
 podman run -v $(pwd):/usr/src/rpmlint/ opensuselp15 python3 -m pytest
```

Also, thanks for the offer to get `requirements-parser` packaged in Fedora, but it's no needed without the requirements check.

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,25 @@
+python-tests-in-package="""
+test/ or tests/ directory in package directory. Tests should not be included in %{python_sitelib}/<pkgname>."""
+python-doc-in-package="""
+doc/ or docs/ directory in package directory. Documentation should go into %{docdir}, not %{python_sitelib}/<pkgname>"""
+python-src-in-package="""
+src/ directory included in %{python_sitelib}/<pkgname>. This should be in a separate -src package.
+"""
```
Conan-Kudo, line 7: What guidelines are enforcing this? There doesn't seem to be anything in Fedora or openSUSE for it.

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,25 @@
...
+python-egginfo-require-not-in-spec="""
+Requires pulled in by the spec do not match upstream requires.txt. Please double-check."""
```
Conan-Kudo, line 9: This needs to be removed.

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,25 @@
...
+this is a global name space not owned by any particular package.
+"""
+python-doc-in-site-packages="""
+doc/ or docs directory installed to %{python_sitelib}. This should never happen
+since this is a global name space not owned by any particular package.
+"""
+python-src-in-site-packages="""
+src/ directory installed to %{python_sitelib}. This should never happen
+since this is a global name space not owned by any particular package.
+"""
```
Conan-Kudo, line 25: Same question as with the stuff in package checks...

Conan-Kudo:
@hroncok Could you review this as well?

rpmlint/descriptions/PythonCheck.toml
scarabeusiv: Guess what happens if two separate packages install site-packages/src or site-packages/test....

rpmlint/descriptions/PythonCheck.toml
scarabeusiv: Why?

jgrassler: It does, indeed - missed that one when removing the Requires check. Thanks for spotting!

rpmlint/descriptions/PythonCheck.toml
Conan-Kudo: That's not what this check is about? This seems to be blocking `%python_sitelib/<pkgname>/{src,tests,docs}`?

rpmlint/descriptions/PythonCheck.toml
Conan-Kudo: In retrospect, this makes sense! 😅

rpmlint/descriptions/PythonCheck.toml
jgrassler: Removed.

hroncok:
In general:

> Verify the egg-info is folder (setuptools based) and not file based (distutils style)

This is needed why? Distributing distutils based projects is not good?

What about dist-info directories?

> Check package's Requires against egg-info requires.txt to make sure the package pulls in everything specified by Upstream.

This is wrong idea. The dependency generator should generate those, not rpmlint check those. More reasons already provided by @Conan-Kudo.

> Check for Python specific forbidden dirs and output various warnings and errors about them.

Will see what this means.

rpmlint/descriptions/PythonCheck.toml
scarabeusiv: It ain't blocking but warning about those, as it is in most cases just wasted disk space esp. on containers.

scarabeusiv:
> > Verify the egg-info is folder (setuptools based) and not file based (distutils style)
> 
> This is needed why? Distributing distutils based projects is not good?
> 
The problem is with pip, it cries a lot when it encounters non-setuptools egg-info.

rpmlint/descriptions/PythonCheck.toml
hroncok: that is indeed weird.

test/test_python.py
```diff
@@ -0,0 +1,84 @@
...
+    assert 'E: python-doc-in-site-packages /usr/lib/python3.8/site-packages/docs' in out
+
+
+@pytest.mark.parametrize('package', ['binary/pythoncheck-python-src-in-site-packages'])
+def test_python_src_in_site_packages(tmpdir, package, pythoncheck):
+    output, test = pythoncheck
+    test.check(get_tested_package(package, tmpdir))
+    out = output.print_results(output.results)
+    assert 'E: python-src-in-site-packages /usr/lib/python2.7/site-packages/src' in out
+    assert 'E: python-src-in-site-packages /usr/lib/python3.8/site-packages/src' in out
```
hroncok, line 73: Add some lib64 tests maybe?

test/test_python.py
```diff
@@ -0,0 +1,84 @@
...
+
+@pytest.mark.parametrize('package', ['binary/pythoncheck-python-tests-in-package'])
+def test_python_tests_in_package(tmpdir, package, pythoncheck):
+    output, test = pythoncheck
+    test.check(get_tested_package(package, tmpdir))
+    out = output.print_results(output.results)
+    assert 'W: python-tests-in-package /usr/lib/python2.7/site-packages/python-mypackage/test' in out
+    assert 'W: python-tests-in-package /usr/lib/python2.7/site-packages/python-mypackage/tests' in out
+    assert 'W: python-tests-in-package /usr/lib/python3.8/site-packages/python-mypackage/test' in out
+    assert 'W: python-tests-in-package /usr/lib/python3.8/site-packages/python-mypackage/tests' in out
```
hroncok, line 35: Why is this a problem?

test/test_python.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+@pytest.mark.parametrize('package', ['binary/pythoncheck-python-src-in-package'])
+def test_python_src_in_package(tmpdir, package, pythoncheck):
+    output, test = pythoncheck
+    test.check(get_tested_package(package, tmpdir))
+    out = output.print_results(output.results)
+    assert 'W: python-src-in-package /usr/lib/python2.7/site-packages/python-mypackage/src' in out
+    assert 'W: python-src-in-package /usr/lib/python3.8/site-packages/python-mypackage/src' in out
```
hroncok, line 44: Why is this a problem?

test/test_python.py
```diff
@@ -0,0 +1,84 @@
...
+
+@pytest.mark.parametrize('package', ['binary/pythoncheck-python-doc-in-package'])
+def test_python_doc_in_package(tmpdir, package, pythoncheck):
+    output, test = pythoncheck
+    test.check(get_tested_package(package, tmpdir))
+    out = output.print_results(output.results)
+    assert 'W: python-doc-in-package /usr/lib/python2.7/site-packages/python-mypackage/doc' in out
+    assert 'W: python-doc-in-package /usr/lib/python2.7/site-packages/python-mypackage/docs' in out
+    assert 'W: python-doc-in-package /usr/lib/python3.8/site-packages/python-mypackage/doc' in out
+    assert 'W: python-doc-in-package /usr/lib/python3.8/site-packages/python-mypackage/docs' in out
```
hroncok, line 24: Why is this a problem?

hroncok:
Generally checking if tests, test, src, docs or doc is not packaged **directly in site-packages** is a good check. Thanks. The rest of the checks I don't really understand.

test/test_python.py
scarabeusiv: Waste of space, and you have the %doc macro for this one.
Same applies for the src and tests.

hroncok:
> The problem is with pip, it cries a lot when it encounters non-setuptools egg-info.

Do you have an example?

hroncok:
I have this:

```
$ file /usr/lib64/python3.8/site-packages/rpm-4.15.1-py3.8.egg-info 
/usr/lib64/python3.8/site-packages/rpm-4.15.1-py3.8.egg-info: ASCII text
```

My pip doesn't cry.

scarabeusiv:
If you have distutils egg and try to run i.e. uninstall it will fail with error:

ERROR: Cannot uninstall 'package'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.

It does work fine with devel build packages, but if you have pypi package that lacked setuptools in deps it gets exposed to this.

test/test_python.py
hroncok: Check that it si marked as %doc then instead?

I can imagine situations, where there is a src called submodule. I cannot imagine situations where a src is just a redundant copy of sources -- do you have an example?

Packaging .../site-packages/python-mypackage/tests is totally fine by me. maintainers can decide whether or not to package it into a separate subpackage. At least the warning should not trigger when the package name ends with "test(s)" and it should possibly suggest this when it fins such tests.

hroncok:
Users should not use pip to uninstall packages not installed by pip.
I am not fine with triggering a warning/error on a perfectly working package, such as python3-rpm.

test/test_python.py
jgrassler: At least in my day-to-day packaging practice I've always followed a package-tests-in-a-dedicated-subpackage convention. Just like @scarabeusiv pointed out for documentation, it is a waste of space and should rate a warning since tests are usually not necessary for a package's day-to-day operation. Also, here's a quick informal survey for `test` modules on my machine:

```
% ls -l /usr/lib/python3.8/site-packages | wc -l
461
% find /usr/lib/python3.8/site-packages -type d -name test       
/usr/lib/python3.8/site-packages/unittest2/test
/usr/lib/python3.8/site-packages/terml/test
/usr/lib/python3.8/site-packages/mypy/test
/usr/lib/python3.8/site-packages/pypeg2/test
/usr/lib/python3.8/site-packages/pyftpdlib/test
/usr/lib/python3.8/site-packages/ometa/test
/usr/lib/python3.8/site-packages/cheroot/test
/usr/lib/python3.8/site-packages/hyperlink/test
/usr/lib/python3.8/site-packages/mypyc/test
/usr/lib/python3.8/site-packages/ndg/httpsclient/test
/usr/lib/python3.8/site-packages/cherrypy/test
```

I had a quick look at all of these and they are all tests. Now there may be a package out there which does have a `test` module that is not a test, but vitally important in day-to-day operation. But the data I am looking at says such a package is an exception and not the rule. Hence I retained this warning, albeit with the added `-test` check on the package name you suggested below.

test/test_python.py
jgrassler: I dropped this check now.

test/test_python.py
jgrassler: Standalone documentation (i.e. anything other than inline documentation comments in code) does not belong in a module directory in sitelib, that being reserved for code and possibly data that goes with the code such as rpmlint's `.toml` files. Documentation marked as such with `%doc` does not fit the bill for that, so I'd say this rates a warning.

jgrassler:
Note: I have left the `python-egg-info-distutils-style` error in there for now. While users _should_ not uninstall stuff installed from a package with pip, they _might_ do that anyway, just like nobody _should_ ever log in as root but use `sudo`, yet plenty of people do it nonetheless...

test/test_python.py
hroncok: With the package name test (and suggestion) this makes sense.

test/test_python.py
hroncok: I can imagine a doc submodule that simply handles some doc things. But if this is a warning, I guess I can live with that. Corner cases are everywhere.

hroncok: `/usr/lib64/python3.8/site-packages/numpy/doc` is a real module.

`/usr/lib64/python3.8/site-packages/pygame/docs` is documentation. I'll report that.

hroncok: yet even `/usr/lib64/python3.8/site-packages/pygame/docs` is an importable module.

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
+python-tests-in-package="""
+test/ or tests/ directory in package directory. Tests should not be included in %{python_sitelib}/<pkgname>."""
```
hroncok, line 2:
```suggestion
test/ or tests/ directory in package directory. Tests in %{python_sitelib}/<pkgname> should be packaged in a separate test or tests subpackage."""
```

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
...
+python-doc-in-package="""
+doc/ or docs/ directory in package directory. Documentation should go into %{docdir}, not %{python_sitelib}/<pkgname>"""
+python-egg-info-distutils-style="""
+The package's egg-info is a distutils style file, rather than a setuptools
+style directory. Please convert to setuptools style.
```
hroncok, line 7: If we keep it, this needs more information. This error is not actionable. What does "convert to setuptools style" mean and how do i do it? Also, this does not explain why is this wrong.

test/test_python.py
hroncok: Suggestion: check if `.../site-packages/python-mypackage/docs/__init__.py` exists and only warn if not?

rpmlint/descriptions/PythonCheck.toml
Conan-Kudo: Note there are packages that _intentionally_ use distutils for various reasons. For example, `createrepo_c` and other `rpm-software-management` projects providing Python bindings use `distutils` so that the skeleton needed for a Python package to be represented in pip is there.

test/test_python.py
jgrassler: Now checking whether it's a module is brilliant, thanks :-)

I'll do that for both doc and src, in fact: if `src` is a Python module I'd give it the benefit of the doubt, but if it isn't it's, quite likely to be source for a component not written in Python that ties into the Python module.

jgrassler: Scratch the `src` part, sorry. I'll continue leaving that out entirely. I'm not going to reintroduce that - lost track of what I'd already removed :-)

rpmlint/checks/PythonCheck.py
```diff
@@ -0,0 +1,64 @@
+import os
+import re
+
+from rpmlint.checks.AbstractCheck import AbstractFilesCheck
+
+# Warning messages
+WARNS = {'tests': 'python-tests-in-package',
```
mcepl, line 7: Could we have, please, also a warning against stared `%files` ownership like `%{python_sitelib}/*` or `%{python_sitearch}/*`, which has tendency for package to swallow something which it shouldn't?

bnavigator:
> 2\. Check package's Requires against egg-info requires.txt to make sure the
>     package pulls in everything specified by Upstream.

Can we outsource this and bring it into rpmlint irrespective of the other issues? Importing packages frequently fails with a pkg_resources error because the rpm package maintainer forgot to sync something from the install_requires field into the rpm runtime dependencies, e.g. here: https://build.opensuse.org/request/show/947487

which creates:

```
[   43s] sphinx-build -b html -d _build/doctrees   . _build/html
[   44s] Running Sphinx v4.4.0
[   45s] 
[   45s] Configuration error:
[   45s] There is a programmable error in your configuration file:
[   45s] 
[   45s] Traceback (most recent call last):
[   45s]   File "/usr/lib/python3.8/site-packages/sphinx/config.py", line 340, in eval_config_file
[   45s]     exec(code, namespace)
[   45s]   File "/home/abuild/rpmbuild/BUILD/greenlet-1.1.2/docs/conf.py", line 20, in <module>
[   45s]     import pkg_resources
[   45s]   File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3243, in <module>
[   45s]     def _initialize_master_working_set():
[   45s]   File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3226, in _call_aside
[   45s]     f(*args, **kwargs)
[   45s]   File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3255, in _initialize_master_working_set
[   45s]     working_set = WorkingSet._build_master()
[   45s]   File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 568, in _build_master
[   45s]     ws.require(__requires__)
[   45s]   File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 886, in require
[   45s]     needed = self.resolve(parse_requirements(requirements))
[   45s]   File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 772, in resolve
[   45s]     raise DistributionNotFound(req, requirers)
[   45s] pkg_resources.DistributionNotFound: The 'importlib-metadata>=4.4' distribution was not found and is required by Sphinx
[   45s] 
[   45s] make: *** [Makefile:45: html] Error 2
```

marxin:
Closing as there's not activity for quite some time.

PR #431, opened 2020-05-07
Remove Python entries from Filelist checks.

jgrassler:
With the dedicated PythonCheck in place, we no longer need the
Python entries in the FilelistCheck configuration.

Note: this should only be merged once https://github.com/rpm-software-management/rpmlint/pull/430 has landed.


marxin:
Closing due not no progress with `PythonCheck`.

Issue #157, opened 2018-07-25
Add configurations for rpm distros and test them


scarabeusiv:
Have rpmlint defaults for all major distros Fedora/SUSE/Mageia/etc.
With that make sure we can verify the behaviour of rpmlint stays sane with whatever config we loaded.
Also we can this way have distro specific tests validated when we load the desired config.

This could make ie openSUSE rpmlint-checks and test repository merged here while not adding extra burden for the other distributions.

Conan-Kudo:
So this probably wouldn't be too difficult to do for Fedora and Mageia, since both distributions use the base set of checks and contribute their check extensions upstream, but openSUSE would be very difficult at present.

scarabeusiv:
Yep I will simply add the suse checks and add the tests, there is no actual reason to not have them here as they simply won't be imported, and if anyone decides to import them they can.

Also with that we can be safe in expecting that the distros can use the rpmlint right away without any excessive patching, and I know we have ton of it in openSUSE...

marxin:
@scarabeusiv I would close this issue.

scarabeusiv:
Yep looks complete for openSUSE, other distros can add theirs if they need tweaking from defaults.

PR #432, opened 2020-05-12
Support multiple config files and overriding descriptions

kstreitova:
A needed preparation for porting of our old rpmlint distro configs to the new toml style.

### DIFF
--- a/rpmlint/cli.py
+++ b/rpmlint/cli.py
@@ -81,7 +81,7 @@ def process_lint_args(argv):
                                      description='Check for common problems in rpm packages')
     parser.add_argument('rpmfile', nargs='*', type=Path, help='files to be validated by rpmlint')
     parser.add_argument('-V', '--version', action='version', version=__version__, help='show package version and exit')
-    parser.add_argument('-c', '--config', type=Path, help='load up additional configuration data from specified path')
+    parser.add_argument('-c', '--config', type=_validate_conf_location, help='load up additional configuration data from specified path (file or directory with *.toml files')
     parser.add_argument('-e', '--explain', nargs='+', default='', help='provide detailed explanation for one specific message id')
     parser.add_argument('-r', '--rpmlintrc', type=Path, help='load up specified rpmlintrc file')
     parser.add_argument('-v', '--verbose', '--info', action='store_true', help='provide detailed explanations where available')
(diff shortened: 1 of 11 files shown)


Conan-Kudo:
@kstreitova I'd rather see defaults and distribution/vendor provided configurations shipped in `/usr/share/rpmlint/*.toml` and admin/user overrides in `/etc/rpmlint/*.toml` and `~/.config/rpmlint/*.toml`.

kstreitova:
> I'm confused as to why you're changing the layout. Distribution rpmlint policy configurations should always be managed separately. In general, I like most of the updates in here, but I do not think the path changing for the default configs is warranted.

The only thing I did here regarding paths was to move configdefaults.toml from rpmlint/configdefaults.toml to rpmlint/configs/configdefaults.toml. I believe that it's reasonable to have all configs at the one place and it also helps us to easily test all configuration files (as requested in #157) with test/test_lint.py/test_run_full_specs() and test_run_full_rpm(). 

I planned to have the following structure for the distro configs that we want to storage here:

configs
├── configdefaults.toml
├── Fedora
│   ├── config.toml
├── openSUSE
│   ├── config.toml
│   ├── licenses.toml
│   └── pie.toml

> @kstreitova I'd rather see defaults and distribution/vendor provided configurations shipped in `/usr/share/rpmlint/*.toml` and admin/user overrides in `/etc/rpmlint/*.toml` and `~/.config/rpmlint/*.toml`.

This rather sounds like a job for the final packaging.

kstreitova:
After a discussion with @scarabeusiv I reworked the support for overriding descriptions from configuration file. The previous approach updated it right after the check initialization, which was nice but it was needed to call `update_description()` method for every check that uses "dynamic" descriptions. And that was prone to be omitted for the future checks. Now the descriptions are updated in the end - right before the description string is requested to be printed.

scarabeusiv:
LGTM.

The only thing I would still not do is the default config move. We don't expect to store any further configs in rpmlint/configs/ folder and even if we allow to ship distro specific configuration they will be simply stored in git and not shipped by rpmlint by default (eg spec file should be responsible to install the right one in /etc).

kstreitova:
Ok, I understand that you don't want to have it under rpmlint directory. I moved configdefaults.toml back to its original location. Ready to be merged.

scarabeusiv:
Yep now it looks like we addressed all the comments, merging in.

PR #433, opened 2020-05-13
Update pytest to 4.x or newer

scarabeusiv:
It seems to fail with pytest3 that is on openSUSE Leap and I don't
see any reason why we should fix it -> just bump the requirement.


scarabeusiv:
Ah well, will take bit more https://bugzilla.suse.com/show_bug.cgi?id=1171561

PR #439, opened 2020-06-24
add security allow list lints

maltek:
This took a lot longer than I'd hoped. This PR adds the allow lists from [openSUSE/rpmlint-checks](https://github.com/openSUSE/rpmlint-checks) developed in parallel to this rewrite.

I'm not sure if the links to the openSUSE wiki in the description texts are acceptable here. If not, we'll have to find some way to change these texts in openSUSE. We're doing something pretty rude here (telling a packager they can't do what they want to) - the least we have to do is tell the packager the exact steps they can take to get the allowed lists to be adapted.


scarabeusiv:
Seems there are bunch of flake8 and lgtm complaints.

Generally okay. Only thing we will have to figure out how to override those descriptions pointing to suse, so it could be reused by the other distributions too. (alternatively we would have to keep it in opensuse branch but it seems okay for master).

rpmlint/checks/Allowlisting.py
```diff
@@ -0,0 +1,428 @@
+# vim: sw=4 ts=4 sts=4 et :
```
mimi1vx, line 1: it is really needed to have this header ?

rpmlint/checks/Allowlisting.py
```diff
@@ -0,0 +1,428 @@
...
+import json
+import os
+import os.path
+import sys
+import traceback
+
+from rpmlint.checks.AbstractCheck import AbstractCheck
+
+
+class DigestVerificationResult(object):
```
mimi1vx, line 18: inheritance of `object` is by default. There isn't a difference between `class Test:`  and `class Test(object):`

rpmlint/checks/Allowlisting.py
```diff
@@ -0,0 +1,428 @@
...
+    def digests(self):
+        """Returns a dictionary specifying file paths and their allow listed
+        digests. The digests are suitable for the
+        Python hashlib module. They're of the form '<alg>:<hexdigest>'. As a
+        special case the digest entry can be 'skip:<none>' which indicates
+        that no digest verification should be performed and the file is
+        acceptable regardless of its contents."""
+        return self.m_digests
+
+    def isSkipDigest(self, digest):
```
mimi1vx, line 92: this should be `staticmethod`

rpmlint/checks/Allowlisting.py
```diff
@@ -0,0 +1,428 @@
...
+
+        self.m_bug = bug
+        self._verifyBugNr()
+        self.m_comment = ''
+        self.m_digests = {}
+
+    def bug(self):
+        return self.m_bug
+
+    def setComment(self, comment):
```
mimi1vx, line 70: setters/getters in python ? why?   please use `@property`

rpmlint/checks/Allowlisting.py
```diff
@@ -0,0 +1,428 @@
...
+        pass
+
+    @property
+    @abc.abstractmethod
+    def error_map(self):
+        """ A dictionary with keys 'unauthorized', 'changed', 'ghost' that maps these to the
+        desired rpmlint error identifiers for this check. """
+        pass
+
+    def __init__(self, config, output):
```
mimi1vx, line 315: please move this to the beginning of class

maltek:
Yes, the original author's C++ background really showed through in that code. I've added a commit that should address your comments.

maltek:
>  Only thing we will have to figure out how to override those descriptions pointing to suse, so it could be reused by the other distributions too.

So, how about:

I adapt `Filter.get_description` to append the text of `reason + "-suffix"` if such a key exists. That'd allow us to drop in additional TOML files with openSUSE-specific information, without having to merge changes in the generic text.

scarabeusiv:
> > Only thing we will have to figure out how to override those descriptions pointing to suse, so it could be reused by the other distributions too.
> 
> So, how about:
> 
> I adapt `Filter.get_description` to append the text of `reason + "-suffix"` if such a key exists. That'd allow us to drop in additional TOML files with openSUSE-specific information, without having to merge changes in the generic text.

We have it already in progress I think @kstreitova https://github.com/rpm-software-management/rpmlint/pull/432 it just needs a bit of polishing.

Also whats weird is that one test failing on filtering in travis right now, it should not be failing :)

rpmlint/checks/Allowlisting.py
```diff
@@ -0,0 +1,386 @@
+#############################################################################
+# Author        : Matthias Gerstner
+# Purpose       : reusable code for dealing with security allow lists
+#############################################################################
```
scarabeusiv, line 4: The whole comments are moot, just convert them to module docstring if it makes sense.

rpmlint/checks/CheckDBUSServices.py
```diff
@@ -0,0 +1,15 @@
+from .Allowlisting import AbstractSimpleAllowlistCheck
```
scarabeusiv, line 1: The imports should be by default absolute unless the path grows crazy complex.

rpmlint/checks/SUIDPermissionsCheck.py
```diff
@@ -0,0 +1,224 @@
...
+    'mail-server.paranoid',
+    'postfix',
+    'postfix.paranoid',
+    'sendmail',
+    'sendmail.paranoid',
+    'squid',
+    'texlive',
+    'texlive.texlive',
+    'otrs',  # bsc#1118049
+)
```
scarabeusiv, line 30: I bet this should be in a config file as a toml list, also any insight where this default came from?

rpmlint/checks/SUIDPermissionsCheck.py
```diff
@@ -0,0 +1,224 @@
...
+        need_set_permissions = False
+        found_suseconfig = False
+        # second pass, find permissions violations
+        for f, pkgfile in pkg.files.items():
+
+            if pkgfile.filecaps:
+                self.output.add_info('E', pkg, 'permissions-fscaps', f'{f} has fscaps "{pkgfile.filecaps}"')
+
+            mode = pkgfile.mode
+            owner = pkgfile.user + ':' + pkgfile.group
```
scarabeusiv, line 120: f'{pkgfile.user}:{pkgfile.group}' ?

rpmlint/checks/SUIDPermissionsCheck.py
```diff
@@ -0,0 +1,224 @@
...
+                            break
+
+                if not script or not found:
+                    self.output.add_info('W', pkg, 'permissions-missing-verifyscript',
+                                         f'missing %verify_permissions -e {f}')
+
+        if need_set_permissions:
+            if 'permissions' not in (x[0] for x in pkg.prereq):
+                self.output.add_info('E', pkg, 'permissions-missing-requires',
+                                     "missing 'permissions' in PreReq")
```
scarabeusiv, line 219: This sounds wrong, PreReq is obsolete var, now we use Requires(phase): to state the deps.

rpmlint/checks/SUIDPermissionsCheck.py
```diff
@@ -0,0 +1,224 @@
...
+
+        if found_suseconfig:
+            # TODO: nothing in tumbleweed uses this anymore. can probable be dropped.
+            self.output.add_info('I', pkg, 'permissions-suseconfig-obsolete',
+                                 '%run_permissions is obsolete')
```
scarabeusiv, line 224: I would probably drop this one. Also it is auto-removed by spec-cleaner too.

rpmlint/descriptions/SUIDPermissionsCheck.toml
```diff
@@ -0,0 +1,104 @@
...
+"""
+
+
+permissions-missing-postin="""
+Please add an appropriate %post section
+"""
+
+
+permissions-missing-requires="""
+Please add 'PreReq: permissions'
```
scarabeusiv, line 83: PreReq is obsolete variable.

Conan-Kudo:
I'm going to do a more in-depth review of this a bit later, but I've got some simple feedback:

* I'm okay with the idea of having this capability in rpmlint master
* I'm _not_ okay with the `suse-` prefix, or openSUSE wiki references leaking in
* I'd like for someone from Fedora to also take a look at this (perhaps @spotrh or @sidhpurwala-huzaifa)
* I don't want references to specific packages anywhere (e.g. `permissions`, etc.)

rpmlint/descriptions/SUIDPermissionsCheck.toml
marxin: Should be `Requires` used here?

marxin:
@maltek Can you please provide links to spec files from which you created the `.rpm` files for testing?

rpmlint/descriptions/SUIDPermissionsCheck.toml
scarabeusiv: No Requires(phase): one.

rpmlint/checks/SUIDPermissionsCheck.py
marxin: I would prefer to stay with `pkgfile.user + ':' + pkgfile.group`.

rpmlint/descriptions/SUIDPermissionsCheck.toml
maltek: I think it should be a plain `Requires`, because the `verifyscript` also has that dependency.

maltek:
> @maltek Can you please provide links to spec files from which you created the `.rpm` files for testing?

Should be these:
https://gist.github.com/maltek/a1da0cd5d3460c361c4c9149e4dd6f54

maltek:
I've rebased the PR and updated it based on the comments and #469.

(From #469:)
> Apparently, upstream reworked the patches significantly

Well, kind of. There's a few of new checks there, but they are much less important than the ones here for openSUSE. Waiting another few months won't help with the fork situation.

marxin:
> I've rebased the PR and updated it based on the comments and #469.

Thank you for the port!
One comment I have is if you can start using `TOML` configuration files for the added checks?

maltek:
>  One comment I have is if you can start using TOML configuration files for the added checks?

Done.

test/test_polkit_allowlist.py
```diff
@@ -0,0 +1,40 @@
+import os
+import os.path
```
mimi1vx, line 2: maybe use of `pathlib` will be better option

maltek: I don't personally see how it'd provide a benefit over a single call to `os.path.join`.

But I don't care much either way, so I'll use that if that's what you prefer.

Conan-Kudo:
@dcantrell Could you please take a look over this?

dcantrell:
Looking over this PR now.  I have some initial thoughts, but will have more to post on Monday.  Currently on a long weekend and only have my phone available.  Thanks.

dcantrell:
Various comments:

- Allow lists contain a lot of vendor-specifics.  As an example, allowlisting.py has _verify_bug_nr() and checks that the value can split on '#' and the first part is either 'bsc', 'boo', or 'bnc'.  I don't think this information is necessary for a tool like rpmlint.

- The files described at [https://github.com/openSUSE/rpmlint-security-whitelistings/blob/master/README.md](url) Contain a lot of unnecessary information.  There's the bug number reference again as well as the package name that a file lives in.  Since the ultimate test is to check what the installed file looks like, you don't really care what package delivers that file.

- In that same README, the metadata based whitelist is essentially storing the fields from stat(2), but not in a great way.  There's the "dev" key which contains both the major and minor number, requiring the code to always have to split that value.  Just have two fields for major and minor.  Likewise with the "owner" key containing "root:root".  Just have separate user and group keys.  The format of this file also feels clumsy.  To make it more widely maintainable, it might be easier to just allow "ls -l" listings for entries and parse those lines.

- At least CheckDBUSServices.py, CronJobsCheck.py, PolkitDefaultPrivsAllowlist.py, and PolkitRulesAllowedCheck.py contain hardcoded paths of where to find files in question.  These sorts of things should go in runtime configuration files to avoid having to patch the code when paths change and to allow variations among vendors.

- SUIDPermissionsCheck.py contains SLES-specifics in the code.  It also looks for usage of /usr/bin/chkstat, which is not a program I am familiar with.

- The PR contains built RPMs for the test suite.  I recommend using the rpmfluff Python module and generating test RPMs in the test suite.

Overall this PR feels very SuSE-specific and not general enough to apply to a wide range of RPM-based distributions.  The code itself contains a lot of vendor specifics and hardcoded paths and things unique to SuSE releases.

I am also not a fan of the data format described in the README.  For [https://github.com/rpminspect/rpminspect](rpminspect), I have **fileinfo** lists per product release defined in the vendor's rpminspect-data package.  Here's what the rawhide one looks like:

[https://github.com/rpminspect/rpminspect-data-fedora/blob/master/fileinfo/rawhide](https://github.com/rpminspect/rpminspect-data-fedora/blob/master/fileinfo/rawhide)

Separate from that are the **capabilities** lists per product release:

[https://github.com/rpminspect/rpminspect-data-fedora/blob/master/capabilities/rawhide](https://github.com/rpminspect/rpminspect-data-fedora/blob/master/capabilities/rawhide)

I have tried to keep these files minimal so they are easily maintained by all contributors.  The files in question can move among packages in the distributions, but we can still enforce the same settings on the file that is ultimately put on the filesystem.

marxin:
> Various comments:

Thanks for them, they are useful.

> 
> * Allow lists contain a lot of vendor-specifics.  As an example, allowlisting.py has _verify_bug_nr() and checks that the value can split on '#' and the first part is either 'bsc', 'boo', or 'bnc'.  I don't think this information is necessary for a tool like rpmlint.

I agree. That format should be verified in the project where you have the configuration file.

> * The files described at https://github.com/openSUSE/rpmlint-security-whitelistings/blob/master/README.md Contain a lot of unnecessary information.  There's the bug number reference again as well as the package name that a file lives in.  Since the ultimate test is to check what the installed file looks like, you don't really care what package delivers that file.

It seems to me logical to have it grouped on a package basis and having an optional comment (or a bug link) seems fine to.
What about the following TOML structure:

```toml
[mypackage]

bug_url = 'https://bugzilla.opensuse.org/show_bug.cgi?id=1177680'
comment = 'Chroot duplicates of some uncritical character devices. urandom was historically packaged non-world-writable.'

[mypackage.digests."/tmp/a"]

algorithm = 'sha256'
hash = 'd536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359'

[mypackage.digests."/tmp/b"]

algorithm = 'sha256'
hash = 'asdfasdfa'
```

> * In that same README, the metadata based whitelist is essentially storing the fields from stat(2), but not in a great way.  There's the "dev" key which contains both the major and minor number, requiring the code to always have to split that value.  Just have two fields for major and minor.  Likewise with the "owner" key containing "root:root".  Just have separate user and group keys.  The format of this file also feels clumsy.  To make it more widely maintainable, it might be easier to just allow "ls -l" listings for entries and parse those lines.

I like the idea of matching that against the `ls -l` format. It's simple.

> * At least CheckDBUSServices.py, CronJobsCheck.py, PolkitDefaultPrivsAllowlist.py, and PolkitRulesAllowedCheck.py contain hardcoded paths of where to find files in question.  These sorts of things should go in runtime configuration files to avoid having to patch the code when paths change and to allow variations among vendors.

I guess a TOML configuration for that can be added.

> * SUIDPermissionsCheck.py contains SLES-specifics in the code.  It also looks for usage of /usr/bin/chkstat, which is not a program I am familiar with.

I guess the paths that are searched can be also factored out into a TOML configuration?

> * The PR contains built RPMs for the test suite.  I recommend using the rpmfluff Python module and generating test RPMs in the test suite.

This is a generic problem, so far we've added a bunch of RPM files that are built from:
https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests

I can't find any documentation for the `rpmfluff` project. Is it capable of building an RPM from a SPEC file?

> 
> Overall this PR feels very SuSE-specific and not general enough to apply to a wide range of RPM-based distributions. The code itself contains a lot of vendor specifics and hardcoded paths and things unique to SuSE releases.
> 
> I am also not a fan of the data format described in the README. For [https://github.com/rpminspect/rpminspect](rpminspect), I have **fileinfo** lists per product release defined in the vendor's rpminspect-data package. Here's what the rawhide one looks like:
> 
> https://github.com/rpminspect/rpminspect-data-fedora/blob/master/fileinfo/rawhide
> 
> Separate from that are the **capabilities** lists per product release:
> 
> https://github.com/rpminspect/rpminspect-data-fedora/blob/master/capabilities/rawhide
> 
> I have tried to keep these files minimal so they are easily maintained by all contributors. The files in question can move among packages in the distributions, but we can still enforce the same settings on the file that is ultimately put on the filesystem.

dcantrell:
> > * The files described at https://github.com/openSUSE/rpmlint-security-whitelistings/blob/master/README.md Contain a lot of unnecessary information.  There's the bug number reference again as well as the package name that a file lives in.  Since the ultimate test is to check what the installed file looks like, you don't really care what package delivers that file.
> 
> It seems to me logical to have it grouped on a package basis and having an optional comment (or a bug link) seems fine to.
> What about the following TOML structure:
> 
> ```toml
> [mypackage]
> 
> bug_url = 'https://bugzilla.opensuse.org/show_bug.cgi?id=1177680'
> comment = 'Chroot duplicates of some uncritical character devices. urandom was historically packaged non-world-writable.'
> 
> [mypackage.digests."/tmp/a"]
> 
> algorithm = 'sha256'
> hash = 'd536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359'
> 
> [mypackage.digests."/tmp/b"]
> 
> algorithm = 'sha256'
> hash = 'asdfasdfa'
> ```

This improves on it.  I would suggest bug_url be a list because it is possible that you may want to reference both the distribution bug as well as one or more upstream bugs (or even a CVE).

algorithm feels unnecessary as that can be determined by the digest hash, but only if you require the full digest hash.  My preference is towards requiring unabbreviated digest hashes and doing away with 'algorithm' and just inferring that in the code.

> > * At least CheckDBUSServices.py, CronJobsCheck.py, PolkitDefaultPrivsAllowlist.py, and PolkitRulesAllowedCheck.py contain hardcoded paths of where to find files in question.  These sorts of things should go in runtime configuration files to avoid having to patch the code when paths change and to allow variations among vendors.
> 
> I guess a TOML configuration for that can be added.

Anything that gets those sorts of lists out of code and in to runtime-maintainable files.

> > * SUIDPermissionsCheck.py contains SLES-specifics in the code.  It also looks for usage of /usr/bin/chkstat, which is not a program I am familiar with.
> 
> I guess the paths that are searched can be also factored out into a TOML configuration?

They should be, yes.

> > * The PR contains built RPMs for the test suite.  I recommend using the rpmfluff Python module and generating test RPMs in the test suite.
> 
> This is a generic problem, so far we've added a bunch of RPM files that are built from:
> https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests
> 
> I can't find any documentation for the `rpmfluff` project. Is it capable of building an RPM from a SPEC file?

rpmfluff began life inside the Red Hat rpmdiff project (internal tool) but is now a Python module project.

https://pypi.org/project/rpmfluff/

While there is value in running tests against known RPMs files, being able to construct them directly from the test suite is also useful.  Especially for external contributors.  rpmfluff lets you create RPM files for testing purposes for tools like this.  It does not take a spec file.

rpmfluff is not perfect, but it is a timesaver for test suites.  If you point me at an example spec file, I can show you how I would synthesize that with rpmfluff.

marxin:
Based on the internal discussion with openSUSE security team (btw. @maltek left the company), we will add the 2 specific checks into `opensuse` branch. Let's follow #594 and #600.

PR #440, opened 2020-07-02
Fix alternatives check with old rpm

scarabeusiv:


Conan-Kudo:
@scarabeusiv Looks like it's still failing on Leap 15. I assume this is for fixing it there?

scarabeusiv:
yep, the other fail is exposed by the new specchecks.

scarabeusiv:
it now fails only on 2 tests (actually on one, it crashes on loading one rpm thus the all rpm test fails too)

Before it was:
`==================== 4 failed, 368 passed in 48.35 seconds =====================
1842`

thisisshub:
One of the errors `test_check_invalid_url[spec/SpecCheck2]` failed in the travis build [here](https://travis-ci.org/github/rpm-software-management/rpmlint/jobs/704275990#L1655) is fixed in #441

scarabeusiv:
Lets look on that PR of yours and fix the crasher there and I will merge this one as it is just a small step in the right direction not breaking anything.

PR #455, opened 2020-08-18
add support for ZSTD compressed files

tpgxyz:
Handle CompressExtension with zst. These days distros are moving to Zstandard as default compression form RPM and other files.

### DIFF
--- a/rpmlint/checks/SourceCheck.py
+++ b/rpmlint/checks/SourceCheck.py
@@ -13,6 +13,7 @@ class SourceCheck(AbstractCheck):
         'gz': 'gzip compressed',
         'tgz': 'gzip compressed',
         'bz2': 'bzip2 compressed',
+        'zst': 'ZSTD compressed',
     }
 
     def __init__(self, config, output):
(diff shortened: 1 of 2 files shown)


rpmlint/checks/SourceCheck.py
```diff
@@ -59,7 +60,7 @@ def _check_permissions(self, fname, pkgfile, pkg):
 
     def _check_compressed_source(self, fname, pkg):
         """
-        Check if the Source is compressed if CompressExtension configuration options is used (xz, gz, tgz or bz2).
+        Check if the Source is compressed if CompressExtension configuration options is used (xz, gz, tgz or bz2, zst).
```
scarabeusiv, line 63: it looks weird now, could you just remove the or? even better sort it alphabetically and keep the or prior xz :)

rest of the pr looks great.

Conan-Kudo: it should be `(gz, tgz, bz2, xz, or zstd)`

tpgxyz: @scarabeusiv  Thanks for catching this. Fixed as @Conan-Kudo requested.

PR #456, opened 2020-08-20
Fix the zstd compression detection

scarabeusiv:
As it needs the zstd import we are passing we need the new dep
(not detected by tests as we don't have anything zstd compressed there).

Also fix other flake8 errors.

### DIFF
--- a/rpmlint/pkg.py
+++ b/rpmlint/pkg.py
@@ -22,6 +22,7 @@ except ImportError:
 import rpm
 from rpmlint.helpers import byte_to_string, ENGLISH_ENVIROMENT, print_warning
 from rpmlint.pkgfile import PkgFile
+import zstd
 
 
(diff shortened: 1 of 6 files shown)


scarabeusiv:
@Conan-Kudo how is zstd python module called on Fedora?

Conan-Kudo:
@scarabeusiv Working on adding it to Fedora now: https://bugzilla.redhat.com/show_bug.cgi?id=1870571

Conan-Kudo:
@scarabeusiv This should now be available in Fedora.

Conan-Kudo:
@scarabeusiv This should now be available in Fedora: https://bodhi.fedoraproject.org/updates/FEDORA-2020-7597c40a16

scarabeusiv:
In we go then :)

PR #457, opened 2020-08-21
Refactor TagsCheck.py

thisisshub:
This commit contains,
- Refactor checks to individual methods
    - Convert %s strings to {}.format methods
    - Rename variables to snake-case-naming style
- Modify Docstrings and methods in TagsCheck.py
- Drop checks no-packager, invalid-packager
     - Modify methods and Docstrings

### DIFF
--- a/rpmlint/checks/TagsCheck.py
+++ b/rpmlint/checks/TagsCheck.py
@@ -99,9 +99,15 @@ class TagsCheck(AbstractCheck):
         rpm_license = pkg[rpm.RPMTAG_LICENSE]
         name = pkg.name
         deps = pkg.requires + pkg.prereq
-        devel_depend = False
         is_devel = FilesCheck.devel_regex.search(name)
         is_source = pkg.is_source
+
+        # List of words to ignore in spell check
+        ignored_words = set()
+        for pf in pkg.files:
+            ignored_words.update(pf.split('/'))
+        for tag in ('provides', 'requires', 'conflicts', 'obsoletes'):
+            ignored_words.update((x[0] for x in 'pkg.' + str(tag)))
 
         # Run checks for whole package
         self._check_invalid_packager(pkg)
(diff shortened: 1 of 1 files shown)


rpmlint/checks/TagsCheck.py
```diff
@@ -304,33 +537,50 @@ def check(self, pkg):
...
+                # Check if a package contians a entry in %changelog
+                # with timestamp thats in the future of its writing
                 elif clt > time.time():
                     self.output.add_info('E', pkg, 'changelog-time-in-future',
                                          time.strftime('%Y-%m-%d', time.gmtime(clt)))
 
+    def _check_license(self, pkg, rpm_license):
+        """Trigger check no-license, invalid-license-exception, invalid-license
+
+        Checks are triggered due to the configuration set by the user in the confifdefaults.toml
```
Conan-Kudo, line 575: `confifdefaults.toml` -> `configdefaults.toml`

Conan-Kudo:
@thisisshub I am actually using the packager tag check for linting spec files building for Debian/Ubuntu, where the `Packager` tag is mandatory, so please don't drop it.

thisisshub:
@Conan-Kudo @scarabeusiv 
Is there anything else you'd like me to change? otherwise, I think this PR can be merged ^^

scarabeusiv:
The lgtm tells you there is stored variable thats not used anywhere.

That probably means that the function split is missing some feature.

kstreitova:
> @thisisshub I am actually using the packager tag check for linting spec files building for Debian/Ubuntu, where the `Packager` tag is mandatory, so please don't drop it. Just disable running the check by default in `configdefaults.toml`.

I think that Debian/Ubuntu rules shouldn't be reasons to hold that code. The main purpose of rpmlint is to lint RPMS and we just can't take other non-rpm distros into account. That doesn't make any sense to me and it sounds like a very non-systematic way.

I think we should just drop it as all RPM based distros we take into consideration during the whole rpmlint rewrite (like Fedora, openSUSE, Mageia, ...) don't use Packager tag anymore.

Conan-Kudo:
@thisisshub Can you squash your last two commits and rewrite it to just be about the typo fix?

thisisshub:
> @thisisshub Can you squash your last two commits and rewrite the commit message to just be about the typo fix?

I shall but after the upstream decides whether to keep or drop those checks since @kstreitova makes a valid point. Since Debian is a pkg distro or a non-rpm distro. I too am in favor of dropping these checks since they are no longer needed.

PR #469, opened 2020-10-01
add security allow list lints (finish #439)

marxin:
The pull request finishes what was started in #439.
I tried to address all @scarabeusiv and @Conan-Kudo comments.


marxin:
Apparently, upstream reworked the patches significantly https://github.com/openSUSE/rpmlint-checks.
That said a new port needs to be done.

PR #470, opened 2020-10-05
Remove man, perl and elfutils from Docker files.

marxin:


scarabeusiv:
I wonder if we need groff too.

scarabeusiv:
Also how the ... your change broke spelling on leap :D

marxin:
> Also how the ... your change broke spelling on leap :D

That's strange. The `man` package removal is responsible for that. Debugging right now..

marxin:
> I wonder if we need groff too.

It's needed here:
https://github.com/rpm-software-management/rpmlint/blob/1473fab69dce38bf1c76415ec49c9d430263c54f/rpmlint/checks/FilesCheck.py#L847-L853

marxin:
> That's strange. The `man` package removal is responsible for that. Debugging right now..

So `man` installed `glibc-locale`. Adding the package fixes the problem.

scarabeusiv:
@marxin that files check honestly sounds useless so it might be even killed (or open it as separate ticket plz).

marxin:
> @marxin that files check honestly sounds useless so it might be even killed (or open it as separate ticket plz).

Sure, doing that in #472.

PR #472, opened 2020-10-05
Remove useless 'manual-page-warning' warning.

marxin:


scarabeusiv:
@Conan-Kudo from my pov this check is something upstream should do, on the same level like the warnings of wrong FSF address in license file. Anything against gutting it?

PR #477, opened 2020-10-21
Implement config overwrite and proper merging.

marxin:
As seen in uiri/toml#9 TOML does not support an option to overwrite an existing key in a dictionary.
Moreover, we need also a sensitive merging of lists as TOML does not support something like:

```
MyList += [
  "Add value"
]
```

The similar happens for e.g.

```
[Description]
my-check = """...."""
```

here one does not want to replace all `Descriptions`, but only `my-check`.

### DIFF
--- a/rpmlint/config.py
+++ b/rpmlint/config.py
@@ -84,6 +84,39 @@ class Config(object):
                 else:
                     print_warning(f'(none): W: error locating user requested configuration: {path}')
 
+    def _merge_dictionaries(self, dest, source, override):
+        """
+        Merge in place dest dictionary for values in source in recursive way.
+        If override is set to True, override instead of merging.
+        """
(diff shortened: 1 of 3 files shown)


rpmlint/config.py
```diff
@@ -99,7 +115,11 @@ def load_config(self, config=None):
                     self.conf_files.append(path)
 
         try:
-            cfg = toml.load(self.conf_files)
+            cfg = {}
+            print(self.conf_files)
```
scarabeusiv, line 119: This looks like debug print only?

marxin:
I've also implemented an override mechanism via `.override.` naming convense of the config files.

scarabeusiv:
Now we can append by default and still can override which might be needed later.

PR #483, opened 2020-10-29
Automatically load openSUSE configs.

marxin:


scarabeusiv:
Nope we should rather deploy them :)

marxin:
How to achieve that, can you please explain it?

marxin:
I've done that in spec file where all configs are installed into `/etc/xdg/rpmlint`.
The location is automatically searched by `rpmlint` for configuration files.

PR #487, opened 2020-10-30
Make output less verbose.

marxin:
Decrease verbosity from:

```
3 blocks
============================================================================================================================ rpmlint session starts ============================================================================================================================
rpmlint: 2.0.0
configuration:
    /home/marxin/Programming/rpmlint/rpmlint/configdefaults.toml
checks: 26, packages: 1


...
audit-devel-32bit.x86_64: E: devel-dependency audit-devel

=========================================================================================================== 1 packages and 0 specfiles checked; 1 errors, 9 warnings ===========================================================================================================

```

to

```
============================================================================================================================ rpmlint session starts ============================================================================================================================
rpmlint: 2.0.0
configuration:
    /home/marxin/Programming/rpmlint/rpmlint/configdefaults.toml
checks: 26, packages: 1

...
audit-devel-32bit.x86_64: E: devel-dependency audit-devel
=========================================================================================================== 1 packages and 0 specfiles checked; 1 errors, 9 warnings ===========================================================================================================

```

### DIFF
--- a/rpmlint/pkg.py
+++ b/rpmlint/pkg.py
@@ -383,10 +383,10 @@ class Pkg(AbstractPkg):
 
     _magic_from_compressed_re = re.compile(r'\([^)]+\s+compressed\s+data\b')
 
-    def __init__(self, filename, dirname, header=None, is_source=False, extracted=False):
+    def __init__(self, filename, dirname, header=None, is_source=False, extracted=False, verbose=False):
         self.filename = filename
         self.extracted = extracted
-        self.dirname = self.dir_name(dirname)
+        self.dirname = self.dir_name(dirname, verbose)
         self.current_linenum = None
 
         self._req_names = -1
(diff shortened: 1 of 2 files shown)


rpmlint/pkg.py
```diff
@@ -477,7 +477,8 @@ def _extract(self, dirname):
             command_str = \
                 'rpm2cpio %(f)s | cpio -id -D %(d)s ; chmod -R +rX %(d)s' % \
                 {'f': quote(str(self.filename)), 'd': quote(dirname)}
-            subprocess.run(command_str, shell=True, env=ENGLISH_ENVIROMENT)
+            subprocess.check_output(command_str, shell=True, env=ENGLISH_ENVIROMENT,
+                                    stderr=subprocess.DEVNULL)
```
Conan-Kudo, line 481: Can we have a flag to restore this? It does occasionally come in handy to see that stuff.

Conan-Kudo (changes requested):

rpmlint/pkg.py
marxin: Sure, I've just done that. Old output is seen with `--verbose` argument.

Conan-Kudo (approved):

PR #499, opened 2020-11-02
ldd: undefined references - only for InstalledPkg.

marxin:
Run undefined-non-weak-symbol and unused-direct-shlib-dependency checks
only for installed packages.

Otherwise, we get very many false positives.


scarabeusiv:
Has a conflict now

marxin:
> Has a conflict now

Resolved.

PR #501, opened 2020-11-02
Restore error reporting for a threshold badness.

marxin:
When a threshold is set, align the behaviour to what
rpmlint-1.x branch does.

### DIFF
--- a/rpmlint/filter.py
+++ b/rpmlint/filter.py
@@ -94,6 +94,8 @@ class Filter(object):
                 level = 'E'
             elif level == 'E':
                 level = 'W'
+        else:
+            badness = 1
         # allow strict reporting where we override levels and treat everything
         # as an error
         if self.strict:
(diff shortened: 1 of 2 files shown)


scarabeusiv:
This is not correct.

In a call now we agreed all will be errors and just at the end of the run we check if the threshold has been reached and error/pass the whole package accordingly based on that.

Conan-Kudo:
The only behavioral control we should have is a flag to control at what point we return a non-zero error code, AFAIK.

If I remember the modes correctly, by default we return non-zero with any errors. If badness threshold is enabled, then we don't return non-zero error code until the threshold is exceeded. And that's it.

marxin:
> If badness threshold is enabled, then we don't return non-zero error code until the threshold is exceeded. And that's it.

We do so, but `--permissive` flag must be on. It was added by you in 45b5aacb61e48db05ce2675ca574f0593f9d607d. I'm fine with that.

The only question I have is: Do we want to assign a score equal to `1` for an error? Right now, unless somebody uses a `setBadness`, all errors have badness equal to zero.

scarabeusiv:
Yes we should consider errors as 1 if they are there. So they go up for threshold.

marxin:
> Yes we should consider errors as 1 if they are there. So they go up for threshold.

Done that. Note that `(Badness: N)` is printed only when `N` > 1. Otherwise, we will see a massive test-suite fallout.

PR #504, opened 2020-11-02
Use UTF8 in LC_ALL.

marxin:
It is needed in order to get proper encoded utf8 errors
for e.g. desktop-file-validate.

### DIFF
--- a/rpmlint/helpers.py
+++ b/rpmlint/helpers.py
@@ -7,7 +7,7 @@ import sys
 from rpmlint.color import Color
 
 
-ENGLISH_ENVIROMENT = dict(os.environ, LC_ALL='en_US')
+ENGLISH_ENVIROMENT = dict(os.environ, LC_ALL='en_US.UTF-8')
 
 
 def string_center(message, filler=' '):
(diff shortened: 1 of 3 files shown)


rpmlint/helpers.py
```diff
@@ -7,7 +7,7 @@
 from rpmlint.color import Color
 
 
-ENGLISH_ENVIROMENT = dict(os.environ, LC_ALL='en_US')
+ENGLISH_ENVIROMENT = dict(os.environ, LC_ALL='en_US.utf8')
```
Conan-Kudo, line 10: Isn't this `en_US.UTF-8`?

Conan-Kudo (changes requested):

rpmlint/helpers.py
marxin: Yes.

scarabeusiv:
Does there need to be else block after the try?

I don't think so...

marxin:
> Does there need to be else block after the try?

You are right!

scarabeusiv:
Now just one more, should there also be added toml description for the error?

marxin:
> Now just one more, should there also be added toml description for the error?

The error already exists:
```
non-utf8-desktopfile="""
The .desktop file is not encoded in UTF-8.
"""
```

scarabeusiv:
Aha!

PR #505, opened 2020-11-02
Fix AlternativesCheck.

marxin:
Fix situation like:
binaries = ['javadocdir', 'java', 'jre_openjdk', 'jre_11', 'javac', 'java_sdk_openjdk', 'java_sdk_11']

then the regex for 'java' removes also 'java_sdk_openjdk' and then
we reach a 'list.remove(x): x not in list' error.


scarabeusiv:
How about a test for this one too?

marxin:
> How about a test for this one too?

Done that.

PR #511, opened 2020-11-03
Remove fsf address verification

scarabeusiv:
This check is too noisy and it is something from upstream to fix.
Nothing much for distributions to handle.
When looking on openSUSE Tumbleweed few thousand packages have this
issue out of 14 thousand that are in distro (not all of them GPL of
course).

### DIFF
--- a/configs/openSUSE/opensuse.toml
+++ b/configs/openSUSE/opensuse.toml
@@ -199,7 +199,6 @@ Filters = [
     '(?:dosutils|skelcd|installation-images|yast2-slide-show|instlux|skelcd-.*|patterns-.*)\.\S+: \w: filelist-forbidden-fhs23 /CD1',
 
 # Too noisy, and usually not something downstream packagers can fix
-    ' incorrect-fsf-address ',
     ' no-manual-page-for-binary ',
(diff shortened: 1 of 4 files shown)


Conan-Kudo:
Maybe openSUSE folks don't try to contact upstream or help them update licensing information, but this is part of Fedora policy, so I will NAK this PR entirely.

scarabeusiv:
And fedora cares? because we ship the same darn packages with the same error. If half of your system errors out it is useless.

Conan-Kudo:
Then disable it in openSUSE. We _do_ care in Fedora and have sent patches to fix things upstream in the past. We don't use badness count in Fedora to have thresholds for failure like openSUSE does, though.

scarabeusiv:
I am not going to abandon this pr as long as you won't show me the fedora fixed it on their end.

We have same packages, honestly mostly in same versions, yet we see the failure because we run it on all builds while fedora runs it only during new package inclusion.

So enlighten me how it helps if thousand packages are failing because it has wrong fsf address. We sent the fixes upstream too, but fixing thousand is way beyond any sensible scope of distro integration.

scarabeusiv:
OTOH it is not worth arguing this, since we can just disable it.

PR #514, opened 2020-11-03
Remove incorrect-fsf-address check.

marxin:
@scarabeusiv told it's a bogus check.


marxin:
Heh, it's dup of #511.

PR #519, opened 2020-11-03
Automatically load rpmlintrc files.

marxin:
Load them from /home/abuild/rpmbuild/SOURCES and allow loading
of multiple files (for now).

rpmlint-mini/rpmlint-mini.config contains:
```
    configs += glob.glob("/home/abuild/rpmbuild/SOURCES/*rpmlintrc")
    configs += glob.glob("/usr/src/packages/SOURCES/*rpmlintrc")
```

### DIFF
--- a/rpmlint/lint.py
+++ b/rpmlint/lint.py
@@ -1,4 +1,5 @@
 import importlib
+from pathlib import Path
 from tempfile import gettempdir
 
 from rpmlint.color import Color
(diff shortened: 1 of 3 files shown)


scarabeusiv:
This is just temporary hack to be thrown away later.

rpmlint/lint.py
```diff
@@ -83,30 +84,34 @@ def _load_installed_rpms(self, packages):
...
-            rpmlintrc += sorted(pkg.glob('*-rpmlintrc'))
-            if len(rpmlintrc) > 1:
-                # multiple rpmlintrcs are highly undesirable
-                print_warning('There are multiple items to be loaded for rpmlintrc, ignoring them: {}.'.format(' '.join(map(str, rpmlintrc))))
-            elif len(rpmlintrc) == 1:
-                self.options['rpmlintrc'] = rpmlintrc[0]
-                self.config.load_rpmlintrc(rpmlintrc[0])
+        if not self.options['rpmlintrc']:
+            # first load SUSE-specific location
+            sources = Path('/home/abuild/rpmbuild/SOURCES')
```
scarabeusiv, line 99: You don't need the other location?

marxin: Do you want me to drop the following:
```
-                # load only from the same folder specname.rpmlintrc or specname-rpmlintrc
-                # do this only in a case where there is one folder parameter or one file
-                # to avoid multiple folders handling
-                pkg = self.options['rpmfile'][0]
-                if pkg.is_file():
-                    pkg = pkg.parent
-                self.options['rpmlintrc'] += self._find_rpmlintrc_files(pkg)
```

scarabeusiv: No I thought that you had the glob for 2 locations in description and here you use just one, otherwise it is good enough.

scarabeusiv:
OK fix the failing test and could you mention this is temporary hack until OBS is tweaked in the commitmsg?

marxin:
> OK fix the failing test and could you mention this is temporary hack until OBS is tweaked in the commitmsg?

Done and the test is fixed.

PR #526, opened 2020-11-10
Add skipIf based on architecture.

marxin:
Many tests can run only on x86_64 due to ldd-failed, so that
I am adding skipIfs.

Tested on `x86_64`, `i686`, `ppc64le`, `ppc64`, `aarch64` and `armv7l`.

### DIFF
--- a/test/Testing.py
+++ b/test/Testing.py
@@ -1,5 +1,8 @@
+import glob
 import os
 from pathlib import Path
+import platform
+import re
 
 from rpmlint.config import Config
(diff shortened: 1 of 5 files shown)


scarabeusiv:
Why not use pytest skipif functionality directly since we rely on pytest itself?

marxin:
> Why not use pytest skipif functionality directly since we rely on pytest itself?

Sure, resolved.

scarabeusiv:
LGTM now, just waiting on the travis.

PR #527, opened 2020-11-10
When terminal output is on, description is missing.

marxin:
It is due to colored strings that are not found
in `self.error_details`.

### DIFF
--- a/rpmlint/filter.py
+++ b/rpmlint/filter.py
@@ -149,7 +149,7 @@ class Filter(object):
         last_issue = ''
         for diag in results:
             if self.info:
-                rpmlint_issue = diag.split()[2]
+                rpmlint_issue = diag.split()[2].rstrip(Color.Reset)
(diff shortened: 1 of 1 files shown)


scarabeusiv:
This is wrong, when we are not on terminal it should use the values that are already empty.

scarabeusiv:
Or we have wrongly the colors hardcoded somewhere.

marxin:
> This is wrong, when we are not on terminal it should use the values that are already empty.

No, I speak about when we are outputting into a terminal. Coloring is enabled, but `rpmlint_issue` is wrapped with coloring string and it cannot be found in description dictionary.

scarabeusiv:
Eeh the coloring was added after it does the lookup for the description, or what happened there that now it is adding the coloring first?

marxin:
All right, we can fix it much easily.

PR #529, opened 2020-11-10
Run file checks in parallel.

marxin:
That speeds up BashismsCheck for dialog rpm files from
15s to 1s.


scarabeusiv:
lgtm, waiting on travis

PR #532, opened 2020-11-10
Drop badness of 109 for makefile-junk.

marxin:
The check breaks quite some packages right now as its badness
reaches threshold very quickly.


scarabeusiv:
This should go to master for first as this one is in the generic tests.

PR #535, opened 2020-11-11
Do not load config files in pytests.

marxin:
This prevents building the package on a system where rpmlint-2.0
is already installed.

### DIFF
--- a/rpmlint/config.py
+++ b/rpmlint/config.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 import re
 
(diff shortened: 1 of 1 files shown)


rpmlint/config.py
```diff
@@ -67,13 +68,16 @@ def find_configs(self, config=None):
...
         self.conf_files.append(self.config_defaults)
 
-        # Then load up config directories on system
-        for directory in reversed(xdg_config_dirs):
-            confdir = Path(directory) / 'rpmlint'
-            if confdir.is_dir():
-                # load all configs in the folders
-                confopts = sorted(confdir.glob('*toml'))
-                self.conf_files += confopts
+        print(os.environ)
```
scarabeusiv, line 71: Debug print?

PR #542, opened 2020-11-11
SpecCheck: Revise rpm-buildroot-usage description and raise to error

### LINKED ISSUE #541: Show modern macros in e.g. rpm-buildroot-usage
```
[ 8319s] MozillaFirefox.src:540: W: rpm-buildroot-usage %build mkdir -p %{buildroot}%{progdir}/browser/extensions
[ 8319s] $RPM_BUILD_ROOT should not be touched during %build or %prep stage, as it may
[ 8319s] break short circuit builds.
```

(#540 would also apply here.)

Proposed new wording:

```
%buildroot ($RPM_BUILD_ROOT) should not be touched during %build or %prep stage, as it may break short circuit builds.
```

Conan-Kudo:
This check now ensures package builds are coherent with RPM 4.16+
behavior across `%prep`, `%build`, and `%install` sections.

Additionally, it notes that using either `%{buildroot}` or `$RPM_BUILD_ROOT`
has the same effect.

Fixes #541

### DIFF
--- a/rpmlint/checks/SpecCheck.py
+++ b/rpmlint/checks/SpecCheck.py
@@ -211,7 +211,7 @@ class SpecCheck(AbstractCheck):
 
             if (current_section in Pkg.RPM_SCRIPTLETS + ('prep', 'build') and
                     contains_buildroot(line)):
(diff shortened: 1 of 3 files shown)


scarabeusiv:
Travis seems dead, localy works.

PR #545, opened 2020-11-11
Remove no-%clean-section section check.

### LINKED ISSUE #537: rpmlint 2 complains about absence of %clean
```
[   55s] libHX.spec: W: no-%clean-section
[   55s] The spec file doesn't contain a %clean section to remove the files installed
[   55s] by the %install section.
```

The `%clean` section is optional and implicit in rpm since a long time. rpmlint should not warn about it.

marxin:
Fixes #537.

### DIFF
--- a/rpmlint/checks/SpecCheck.py
+++ b/rpmlint/checks/SpecCheck.py
@@ -506,6 +506,7 @@ class SpecCheck(AbstractCheck):
         self._check_no_cleaning_of_buildroot(pkg, buildroot_clean)
         self._check_no_buildroot_tag(pkg, buildroot)
         self._check_no_s_section(pkg, section)
+        self._check_superfluous_clean_section(pkg, section)
         self._check_more_than_one_changelog_section(pkg, section)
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
@marxin We probably want to flip this check to be throwing errors if `%clean` is in a spec file.

Conan-Kudo (changes requested):
We probably want to flip this check to be throwing errors if `%clean` is in a spec file.

marxin:
> We probably want to flip this check to be throwing errors if `%clean` is in a spec file.

Apparently, we have ~400 packages in `openSUSE:Factory` that contains `%clean` section.
Are you sure we really want to error on that?

Conan-Kudo:
Yes, that section should not be present since RPM has its own `%clean` logic.

scarabeusiv:
Yep Neal is right, spec-cleaner automatically removes the section too.

marxin:
Nice, I've just implemented the new check.

Conan-Kudo (approved):

PR #546, opened 2020-11-11
Remove language from TagCheck errors and warnings.

### LINKED ISSUE #538: name-repeated-in-summary emits stray "C"
(Happens in the rpmlint2-based openSUSE:Factory:Staging:M/AppStream)

```
[  139s] AppStream.x86_64: W: name-repeated-in-summary C AppStream
[  139s] AppStream.src: W: name-repeated-in-summary C AppStream
[  139s] The name of the package is repeated in its summary. Make the summary brief and
[  139s] to the point without including redundant information in it.
```

I never understood why there is a "C" or what it is supposed to mean. Can we remove it, lest it be replaced by something meaningful?

marxin:
Fixes #538.

### DIFF
--- a/rpmlint/checks/TagsCheck.py
+++ b/rpmlint/checks/TagsCheck.py
@@ -156,14 +156,17 @@ class TagsCheck(AbstractCheck):
                 self.output.add_info('E', pkg, 'spelling-error', typo)
         for i in description.splitlines():
             if len(i) > self.max_line_len:
-                self.output.add_info('E', pkg, 'description-line-too-long', lang, i)
+                self.output.add_info('E', pkg, 'description-line-too-long', self._lang_for_error(lang), i)
             res = self.forbidden_words_regex.search(i)
             if res and self.config.configuration['ForbiddenWords']:
-                self.output.add_info('W', pkg, 'description-use-invalid-word', lang,
+                self.output.add_info('W', pkg, 'description-use-invalid-word', self._lang_for_error(lang),
                                      res.group(1))
             res = tag_regex.search(i)
             if res:
-                self.output.add_info('W', pkg, 'tag-in-description', lang, res.group(1))
+                self.output.add_info('W', pkg, 'tag-in-description', self._lang_for_error(lang), res.group(1))
+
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Isn't this supposed to be for multi-language spec files [like this one](https://src.fedoraproject.org/rpms/algobox/blob/master/f/algobox.spec)?

marxin:
> Isn't this supposed to be for multi-language spec files [like this one](https://src.fedoraproject.org/rpms/algobox/blob/master/f/algobox.spec)?

Do you mean printing the language if it's different from the default one (`C`), or?

Conan-Kudo:
@marxin I think it's supposed to check each language, print the language associated with the tag, and print the warning/error associated with that language variant of the tag.

scarabeusiv:
Yeah the lang parsing is correct and should not be removed.

You can have Summary(cs): whatever

We should just probably ommit it in a case where the lang is C == default == most probably english.

marxin:
> We should just probably ommit it in a case where the lang is C == default == most probably english.

Should be fixed now.

rpmlint/checks/TagsCheck.py
```diff
@@ -156,14 +156,17 @@ def check_description(self, pkg, lang, ignored_words):
...
-                self.output.add_info('W', pkg, 'description-use-invalid-word', lang,
+                self.output.add_info('W', pkg, 'description-use-invalid-word', self._lang_for_error(lang),
                                      res.group(1))
             res = tag_regex.search(i)
             if res:
-                self.output.add_info('W', pkg, 'tag-in-description', lang, res.group(1))
+                self.output.add_info('W', pkg, 'tag-in-description', self._lang_for_error(lang), res.group(1))
+
+    def _lang_for_error(self, lang):
+        return lang if lang != 'C' else None
```
Conan-Kudo, line 169: Also `C.UTF-8` too, right?

Conan-Kudo (changes requested):

Conan-Kudo (approved):

PR #560, opened 2020-11-12
Add SLP check #371

marxin:
This pull request contains what Tomas did in #421.


scarabeusiv:
Just for sake of obviousness, I split it in so many commits for easier developing, feel free to squash that stuff down.

rpmlint/configdefaults.toml
```diff
@@ -20,6 +20,7 @@ Checks = [
     "MixedOwnershipCheck",
     "PkgConfigCheck",
     "PostCheck",
+    "SharedLibraryPolicyCheck",
```
scarabeusiv, line 23: This one is not for default to be turned on.

rpmlint/configdefaults.toml
```diff
@@ -288,6 +289,12 @@ ValidLicenseExceptions = []
 # Default white list for PAM modules
 PAMModulesWhiteList = []
 
+# Default list of essential dependencies for SLP policy check
+SLPEssentialDependencies = []
+
+# Default list of legacy exceptions for SLP policy check
+SLPPolicyLegacyExceptions = []
```
scarabeusiv, line 296: No exeptions.

Conan-Kudo:
@marxin @scarabeusiv This policy is _off_ by default, right? If so, then this looks fine to me.

rpmlint/configdefaults.toml
```diff
@@ -288,6 +289,12 @@ ValidLicenseExceptions = []
 # Default white list for PAM modules
 PAMModulesWhiteList = []
 
+# Default list of essential dependencies for SLP policy check
+SLPEssentialDependencies = []
+
```
scarabeusiv, line 294: I killed this in my PR on purpose man.

scarabeusiv (changes requested):

rpmlint/descriptions/BinariesCheck.toml
```diff
@@ -64,7 +64,7 @@ Another solution can be to include a version number in the file names
 themselves.
 """
 incoherent-version-in-name="""
```
Conan-Kudo, line 66: We should probably rename this check. The name doesn't match what it is.

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # Check for non-versioned libs in a std lib package
+        if std_lib_package:
+            for lib in libs.copy():
+                if (not (lib[-1].isdigit() or
+                         self.re_soname_strongly_versioned.search(lib))):
+                    self.output.add_info('W', pkg, 'shlib-unversioned-lib', lib)
+
+        # If this package should be or should be splitted into shlib
+        # package(s)
```
scarabeusiv, line 132: The whole soname check is already in binaries check.

Conan-Kudo (changes requested):

marxin:
> Just for sake of obviousness, I split it in so many commits for easier developing, feel free to squash that stuff down.

Done.

rpmlint/configdefaults.toml
Conan-Kudo: Yeah, let's not have this on by default. Many RPM distros don't use this policy today.

rpmlint/configdefaults.toml
Conan-Kudo: I agree with this. Otherwise this check makes no sense.

rpmlint/configdefaults.toml
Conan-Kudo: What does this even mean?

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # If this is a program package (all libs it provides are
+        # required by itself), bail out
+        if not pkg.name.startswith('lib') and not libs.difference(reqlibs):
+            return
+
+        std_lib_package = False
+        pkg_ends_with_digit = pkg.name[-1].isdigit()
+        if pkg.name.startswith('lib') and pkg_ends_with_digit:
+            std_lib_package = True
```
scarabeusiv, line 106: isn’t this checked by re_soname_pkg?

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,174 @@
...
+                        break
+
+        # Check for non-versioned libs in a std lib package
+        if std_lib_package:
+            for lib in libs.copy():
+                if (not (lib[-1].isdigit() or
+                         self.re_soname_strongly_versioned.search(lib))):
+                    self.output.add_info('W', pkg, 'shlib-unversioned-lib', lib)
+
+        if not pkg.name.startswith('lib') or pkg.name.endswith('-lang'):
```
scarabeusiv, line 126: this return is weird can’t we match the regexp above and just either proceed or not.

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,174 @@
...
+            done = set()
+            for directory in dirs:
+                if directory.startswith(sysdir):
+                    ssdir = directory[len(sysdir):].split('/')[0]
+                    if not ssdir[-1].isdigit():
+                        cdirs.add(sysdir + ssdir)
+                    done.add(directory)
+            dirs = dirs.difference(done)
+        for directory in cdirs:
+            self.output.add_info('E', pkg, 'shlib-policy-nonversioned-dir', directory)
```
scarabeusiv, line 174: this is prone to many mistakes, thats why i didnt finish it up. it should act as per above comment, but now does not.

scarabeusiv (changes requested):

rpmlint/checks/SharedLibraryPolicyCheck.py
mimi1vx: isn't this test too late ?

rpmlint/checks/SharedLibraryPolicyCheck.py
marxin: I guess `re_soname_pkg` is more restrictive than `startswith('lib')`.

marxin: We can likely use it..

marxin: So we can't as the regex expect a trailing digit in package name.
And we need to catch: `E: shlib-policy-nonversioned-dir /usr/lib64/hello` in `libslp-missing-suffix.rpm` package.

rpmlint/checks/SharedLibraryPolicyCheck.py
marxin: We want to catch `shlib-unversioned-lib` also for a package not starting with `lib` if I'm correct.

rpmlint/checks/SharedLibraryPolicyCheck.py
marxin: I debug the section and it works (somehow). What's wrong with that exactly?

scarabeusiv: just search for this error across the whole tumbleweed. It is very commonly wrong not picking up, or picking too many things.

scarabeusiv: As by the example in the comment above.

marxin: I see, do you suggest to remove it for now?

marxin: Ok, I removed the patch and added a `FIXME`. Hope it's fine now.

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+
+    def __init__(self, config, output):
+        super().__init__(config, output)
+        self.re_soname_strongly_versioned = re.compile(r'-[\d\.]+\.so$')
+        # the pkgname is based on soname if ending with number; special option is flavor build
+        self.re_soname_pkg = re.compile(r'^lib\S+(\d+(-(32|64)bit)?)$')
+        self.re_so_files = re.compile(r'\S+.so((.(\d+))+)?$')
+
+    def _libname_from_soname(self, soname):
```
scarabeusiv, line 29: Where is this call used?

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def check(self, pkg):
+        if pkg.is_source:
+            return
+
+        # we don't care about develpkgs
+        if pkg.name.endswith('-devel'):
+            return
+
+        # check the pkg has any libname
```
scarabeusiv, line 49: Could this be split to separate function for just the shlib policy missing?

rpmlint/checks/SharedLibraryPolicyCheck.py
scarabeusiv: but there can't be slpp package without starting by lib*

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,158 @@
...
+                if (dep[1] & (rpm.RPMSENSE_GREATER | rpm.RPMSENSE_EQUAL)) == rpm.RPMSENSE_EQUAL:
+                    self.output.add_info('W', pkg, 'shlib-fixed-dependency', formatRequire(dep[0], dep[1], dep[2]))
+
+        # Verify non-lib stuff does not add dependencies
+        if libs:
+            for dep in pkg_requires:
+                if '.so.' in dep and dep not in libs and dep not in libs_needed:
+                    self.output.add_info('E', pkg, 'shlib-policy-excessive-dependency', dep)
+
+        # FIXME: implement
```
scarabeusiv, line 149: Ticket for 2.1 too?

scarabeusiv (changes requested):

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,146 @@
...
+            for lib in libs.copy():
+                if (not (lib[-1].isdigit() or
+                         self.re_soname_strongly_versioned.search(lib))):
+                    self.output.add_info('W', pkg, 'shlib-unversioned-lib', lib)
+
+        if not pkg.name.startswith('lib') or pkg.name.endswith('-lang'):
+            return
+
+        # Verify no non-lib stuff is in the package
+        dirs = set()
```
scarabeusiv, line 118: where is this dirs variable used? same for the dirs bit above

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,146 @@
...
+        if not libs.difference(reqlibs):
+            return
+
+        std_lib_package = False
+        pkg_ends_with_digit = pkg.name[-1].isdigit()
+        if pkg.name.startswith('lib') and pkg_ends_with_digit:
+            std_lib_package = True
+
+        # ignore libs in a versioned non_std_dir
+        if std_lib_package:
```
scarabeusiv, line 97: if std_lib_pkg is checked 3x and it does not make sense to do it trice? or to have it like 3 functions that do some stuff only if std_lib_package?

rpmlint/checks/SharedLibraryPolicyCheck.py
```diff
@@ -0,0 +1,129 @@
...
+        if not libfiles:
+            # verify if name does not match the slpp and if we still don't have any lib then error out
+            if self.re_soname_pkg.match(pkg.name):
+                self.output.add_info('E', pkg, 'shlib-policy-missing-lib')
+
+    def check(self, pkg):
+        if pkg.is_source:
+            return
+
+        # we don't care about develpkgs
```
scarabeusiv, line 47: The comment now does not make sense.
Really put in comment in this stuff otherwise nobody will know whats happening later.

scarabeusiv (commented):
otherwise getting usable i would say

marxin:
> otherwise getting usable i would say

 Can we please merge it now, I hope it's a reasonable starting point...

scarabeusiv:
I would say good for now, there is room for improvement but that can be handled on later time.

PR #561, opened 2020-11-12
Fix encoding error for readelf parser.

marxin:
This happens e.g. for `containerd.x86_64` package.

### DIFF
--- a/rpmlint/readelfparser.py
+++ b/rpmlint/readelfparser.py
@@ -274,18 +274,21 @@ class ElfSymbolTableInfo:
         self.parse()
 
     def parse(self):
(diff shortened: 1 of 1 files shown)


scarabeusiv:
There is debug print left, also when exactly the unicode error happens we should just wrap that.

marxin:
It comes from 

```
    78: 0000000002395ea0     8 OBJECT  GLOBAL DEFAULT   24 crypto/sha256.NewÂf
```

and we really want to bail out. It's again `go` the produces such broken symbols.

scarabeusiv:
Lovely, lets wait for CI.

PR #564, opened 2020-11-13
Fix exception handling for BashismsCheck and BinariesCheck.

marxin:
I noticed that when ThreadPooll is used, exceptions are not propagated
properly. I fixed that and that requires some changes in BinariesCheck.


scarabeusiv:
Could there be at least one test to trigger it to see if it raises properly? As now nothing it tests triggers it? Or would it be too complex?

marxin:
> Could there be at least one test to trigger it to see if it raises properly? As now nothing it tests triggers it? Or would it be too complex?

Yep, I have a nice test-case.

PR #573, opened 2020-11-13
Short errors for readelf-error.

marxin:
From:

```
dev86.x86_64: E: readelf-failed /usr/lib64/bcc/libdos.a readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
```

to:

```
dev86.x86_64: E: readelf-failed /usr/lib64/bcc/libdos.a readelf: Error: Not an ELF file - it has the wrong magic bytes at the start
```

### DIFF
--- a/rpmlint/readelfparser.py
+++ b/rpmlint/readelfparser.py
@@ -339,6 +339,7 @@ class ReadelfParser:
     in a structured format.
     """
 
+    NOT_ELF_ERROR = 'Error: Not an ELF file - it has the wrong magic bytes at the start'
     so_regex = re.compile(r'/lib(64)?/[^/]+\.so(\.[0-9]+)*$')
 
     def __init__(self, pkgfile_path, path):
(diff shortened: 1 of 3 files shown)


marxin:
In total, we have now `110324` of these extra messages for `openSUSE:Factory`.

scarabeusiv:
i would recommend to add test case for this one

marxin:
> i would recommend to add test case for this one

Done.

PR #579, opened 2020-11-18
Updates for ErlangCheck

matwey:


marxin:
Thank you for the PR. Can you please also add a unit test for it?

scarabeusiv:
Also is there some quick explanation how to enable the compiler info like with the debug? To make the description self contained. But if it is long to explain then it is good as-is.

matwey:
> Also is there some quick explanation how to enable the compiler info like with the debug? To make the description self contained. But if it is long to explain then it is good as-is.

Normally, it should be always present, if it is absent then something is going wrong. Maybe I just need to remove misleading phrase "Please, make sure that you enable it while compiling."

marxin:
I like the patch except I would add a test-case.
Can you please create a file that will miss compilation info and add include it in tests?

matwey:
> I like the patch except I would add a test-case.
> Can you please create a file that will miss compilation info and add include it in tests?

Yes. I remember. I am working on it.

matwey:
@marxin where did the spec file for `erlang-0-0.x86_64.rpm` go? Should I update it in the original place? https://github.com/openSUSE/rpmlint-tests/commits/master/tests/erlang.spec

marxin:
> @marxin where did the spec file for `erlang-0-0.x86_64.rpm` go? Should I update it in the original place? https://github.com/openSUSE/rpmlint-tests/commits/master/tests/erlang.spec

It comes now from:
https://build.opensuse.org/package/show/devel:openSUSE:Factory:rpmlint:tests/erlang

You can either make a submit request, or send me a patch and I will motify the package for you.

matwey:
I think I need your help for fixing this: https://build.opensuse.org/package/show/devel:openSUSE:Factory:rpmlint:tests/erlang#comment-1365355

marxin:
> I think I need your help for fixing this: https://build.opensuse.org/package/show/devel:openSUSE:Factory:rpmlint:tests/erlang#comment-1365355

Yep, resolved by moving that to:
https://build.opensuse.org/package/show/devel:openSUSE:Factory:rpmlint:tests/erlang-test

marxin:
I support the pull request.

scarabeusiv:
Do we need newer erlang lib on leap and fedora? asking because it only works on tumbleweed

scarabeusiv:
also should it be warning because there are valid reasons for not having the info or should it be error the packager should always fix?

matwey:
> also should it be warning because there are valid reasons for not having the info or should it be error the packager should always fix?

I consider this as 'black swan' case, it something that currently never happens. Maybe we will see some new compiler for BEAM files or something else, but it is better to print the warning instead of crashing application with exception.

matwey:
> Do we need newer erlang lib on leap and fedora? asking because it only works on tumbleweed

We need `pybeam 0.7`. Not sure how to overcome it for Leap.

marxin:
> We need `pybeam 0.7`. Not sure how to overcome it for Leap.

I would just skip the test with `@pytest.mark.skipif(...)`

scarabeusiv:
You can also sent a MU for the leap pybeam I think.

For the Fedora we can ask @Conan-Kudo to bump it.

FWIW if it is something that should not really happen then I would consider it an error, as warnings are something like mild suggestions to act upon and this seems like something you as a packager certainly should act and fix.

PR #582, opened 2020-11-20
Fix erlang library deprecation warning:

marxin:
```
  /usr/lib/python3.8/site-packages/pybeam/erlang_types.py:42: DeprecationWarning: x.value is deprecated; use x instead
    warnings.warn("x.value is deprecated; use x instead", DeprecationWarning)
```


scarabeusiv:
Do you really have to use byte_to_string then if you already convert it to str using builtin?

marxin:
Yes, `str(beam.compileinfo['source'])` is `b'/home/abuild/rpmbuild/BUILD/m.erl'`.

PR #584, opened 2020-11-23
checkbashisms: support --early-fail option of the script.

marxin:
The script approved addition of the option. With the option,
it exits right after a first error is seen:
https://salsa.debian.org/debian/devscripts/-/merge_requests/208

It speed up e.g. tar package check from 90s to 1s.

### DIFF
--- a/rpmlint/checks/BashismsCheck.py
+++ b/rpmlint/checks/BashismsCheck.py
@@ -9,6 +9,14 @@ class BashismsCheck(AbstractFilesCheck):
     def __init__(self, config, output):
         super().__init__(config, output, r'.*')
         self.use_threads = True
+        self._detect_early_fail_option()
+
+    def _detect_early_fail_option(self):
+        self.has_early_fail_option = False
+        output = subprocess.check_output(['checkbashisms', '--help'],
+                                         shell=True, encoding='utf8')
+        # FIXME: remove in the future
+        self.use_early_fail = '[-e]' in output
 
     def check_file(self, pkg, filename):
         root = pkg.dirName()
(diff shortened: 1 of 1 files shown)


scarabeusiv:
I wonder if it is good idea to bail out right away or not, also we might just enforce newer check bashism in deps rather than have huge magic to detect it.

Do we print out also the reason why it failed? If not then stopping on first detected error on the file is good, if we list all the errors then we should probably keep doing so.

Conan-Kudo:
> I wonder if it is good idea to bail out right away or not, also we might just enforce newer check bashism in deps rather than have huge magic to detect it.

Considering `devscripts` hasn't made a release with that feature added yet, it is premature to consider making it a requirement.

marxin:
> I wonder if it is good idea to bail out right away or not, also we might just enforce newer check bashism in deps rather than have huge magic to detect it.

Well, without the patch, `BashismsCheck` wastes quite a huge time for various packages.
Moreover, the last `devscripts` release was made 6 months ago...

> 
> Do we print out also the reason why it failed? If not then stopping on first detected error on the file is good, if we list all the errors then we should probably keep doing so.

We do not list any errors, we just check for a return code of `checkbashisms`.

marxin:
> Considering `devscripts` hasn't made a release with that feature added yet, it is premature to consider making it a requirement.

I see. On the other hand, it's an optional run-time detection, not a requirement.

marxin:
Added a comment, hope @scarabeusiv can merge it now..

scarabeusiv:
Yep it does nto require new script, just uses it if it can. lgtm.

marxin:
Thanks, please rebase me `opensuse` branch.

PR #585, opened 2020-11-23
makefile-junk: restore to the original implementation

marxin:
I created issues based on an openSUSE staging project and the new error is not welcomed much:
https://bugzilla.suse.com/show_bug.cgi?id=1179039

@bitstreamout:
> Many of those Makefiles are use to run the examples below the doc directories ... I do not want to remove them as in that case  the users are not able to simply to a make to see how stuff works.
> Who has decided that Makefiles are evil?

@wfeldt:
> I'm not saying the package structure can't be improved or Makefiles are
> vital here - just that you seem to imply that using make is evil.

Pedro Monreal Gonzalez:
> Thanks for the report. These Makefiles are for to build the demo files and are packaged as documentation. I think removing > these Makefiles would not be recommended. Is there a way to skip this check for these files?

I tend to disable it as an error for now.
What do you think @scarabeusiv and @kstreitova?


kstreitova:
This check comes from [CheckCommonFiles.py](https://github.com/openSUSE/rpmlint-checks/blob/master/CheckCommonFiles.py#L53) that implements [this](https://bugzilla.suse.com/show_bug.cgi?id=217472#c8) bug.

Original check looks like this:

```python
if (f.endswith("/Makefile.am") and f[:-3] + ".in" in files and
                    f in pkg.docFiles()):

```
So as it's described in that bug, it should detect a situation  when packager copies the whole documentation directory into the package, including the buildsystem.

However, the check meaning was [changed](https://github.com/rpm-software-management/rpmlint/commit/583ce71fc3fb47bec85c4577c1b531d41dc9cdca) when it was ported for rpmlint 2.0. Now it's pretty generic:

```python
makefile_regex = re.compile(r'(Makefile(\.am|\.in|)|CMakeLists.txt)$')
```
This doesn't look right. As described in the original bug and as maintainers indicated now, sometimes it's ok to have standalone Makefile in the package. (e.g. for building examples and so).

If Tomas doesn't have a reason for keeping it, I would revert it to the original behaviour.

scarabeusiv:
I've put it as least common usecase to catch all makefiles that should not be installed.

Tbh if anything the check does not make sense as it was to reduce size to fit more stuff on media and we ain't saving anything really. But yea feel free to tweak its behaviour rather than just making it warning.

marxin:
> This doesn't look right. As described in the original bug and as maintainers indicated now, sometimes it's ok to have standalone Makefile in the package. (e.g. for building examples and so).
> 
> If Tomas doesn't have a reason for keeping it, I would revert it to the original behaviour.

Thank you for a nice analysis. That said, I'm suggesting to restore the behavior which used to be in the `CheckCommonFiles`.

PR #586, opened 2020-11-24
Fix filtering for errors with a badness.

marxin:
Filtering does not consider '(Badness: 12345)' part:
`nfs-client\.\S+: \w: filelist-forbidden-backup-file /var/lib/nfs/sm.bak`

Should match:
`nfs-client.x86_64: E: filelist-forbidden-backup-file (Badness: 10000) /var/lib/nfs/sm.bak`


Conan-Kudo (approved):

scarabeusiv:
Could you also add test using it? As you already adjusted the test config...

marxin:
> Could you also add test using it? As you already adjusted the test config...

It's already covered by the existing testcase `test_filtered_output`.

PR #587, opened 2020-11-24
Strip one filelist-forbidden-backup-file rule in Filters.

marxin:
There's a bogus trailing space.


Conan-Kudo:
@marxin It seems like a lot of these have trailing spaces. What makes this one bogus?

marxin:
This filter covers the entire error message.
The other, ending with space, typically cover only the name of an error:
`    ' init-script-name-with-dot ',` which is fine.

PR #591, opened 2020-11-26
Consider also rpmlintrc file.

marxin:
For instance ocaml uses the file.

The file can be seen here:
https://build.opensuse.org/package/view_file/openSUSE:Factory:Staging:M/ocaml/rpmlintrc?expand=1

### DIFF
--- a/rpmlint/lint.py
+++ b/rpmlint/lint.py
@@ -159,7 +159,7 @@ class Lint(object):
         return existing_packages
 
     def _find_rpmlintrc_files(self, path):
-        rpmlintrcs = []
+        rpmlintrcs = list(path.glob('rpmlintrc'))
         rpmlintrcs += sorted(path.glob('*.rpmlintrc'))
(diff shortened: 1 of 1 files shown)


scarabeusiv:
Actually surprising this works, it goes against our own opensuse spec that says it is to be sourcename-rpmlintrc (that dot working i considered minor cosmetics, but without the sourcename it should've not worked :D)

marxin:
All right, I'm gonna fix the `ocaml` package rather than `rpmlint`.

Conan-Kudo:
Yeah, un-prefixed rpmlintrc files shouldn't work with rpmlint. Mostly because that can cause headaches if multiple specs exist in the same directory.

marxin:
> Yeah, un-prefixed rpmlintrc files shouldn't work with rpmlint. Mostly because that can cause headaches if multiple specs exist in the same directory.

Works for me, thanks.

PR #593, opened 2020-11-26
Do not print empty specfile-errors.

marxin:
Empty lines comes from e.g.

```
warning: extra tokens at the end of %else directive in line 250:  %else	# suse_version

warning: extra tokens at the end of %endif directive in line 252:  %endif	# suse_version

warning: extra tokens at the end of %else directive in line 293:  %else	# suse_version

warning: extra tokens at the end of %endif directive in line 295:  %endif	# suse_version

warning: extra tokens at the end of %else directive in line 296:  %else	# default_qt
```


scarabeusiv:
In theory you should've stripped the line prior doing the condition. But this will of course work too.

marxin:
> In theory you should've stripped the line prior doing the condition. But this will of course work too.

Yep, I improved that now.

PR #594, opened 2020-12-03
[opensuse] Add FileDigestCheck

marxin:


marxin:
@mgerstner

I've got few notes about File Digest check:
- `'{group}-file-digest-mismatch'` was removed; if you have multiple digest groups, you can't know which one should be the match and isn't
- each `FileDigestLocation` has mandatory attribute `FollowSymlinks`, hope it's acceptable as we have just couple of these
- I split checking of symlinks and ghost files in `FileDigestCheck`; if there's an error, file digest groups are not iterated at all

mgerstner:
I looked at the code and tested the new check. The logic should be okay now. I still have a couple of detailed comments though that I will add individually soon.

configs/openSUSE/security.toml
```diff
@@ -26,15 +26,29 @@ Locations = [
     "/usr/share/polkit-1/rules.d/"
 ]
 
-[FileDigestGroup.atop."boo#12345"]
+[FileDigestGroup.file-signature."boo#12345"]
```
mgerstner, line 29: We now lost the coupling between whitelisting and package again.

configs/openSUSE/security.toml
```diff
@@ -26,15 +26,29 @@ Locations = [
...
 note = """
 Builds statistics based on Squid logfile metadata.
 Include sarg-reports which is SUSE specific and important for privilege dropping.
 """
 
-[FileDigestGroup.atop."boo#12345".digests]
-"/etc/cron.daily/suse.de-sarg" = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
-"/etc/cron.monthly/suse.de-sarg" = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
+[FileDigestGroup.file-signature."boo#12345".digests."/etc/cron.daily/suse.de-sarg"]
```
mgerstner, line 35: The configuration format has become quite verbose and redundant now, because we have to repeat the `FileDigestGroup.file-signature."<bug>".digests.` prefix for each path. Items belonging together aren't structured together any more. This could be solved by adding some "ascii art" like comments or whitespace in the file. Still it feels harder to read now.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -8,24 +8,28 @@
 class FileDigestCheck(AbstractCheck):
     def __init__(self, config, output):
         super().__init__(config, output)
-        self.digest_groups = {}
+        self.digest_configurations = {}
         self.follow_symlinks_in_group = {}
         for group, values in self.config.configuration['FileDigestLocation'].items():
-            self.digest_groups[group] = [Path(p) for p in values['Locations']]
```
mgerstner: Could you also rename the `group` for loop variable? Also in some other for loops that deal with `digests_configurations`.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -34,46 +38,90 @@ def _get_digest_group_from_path(self, path):
                     pass
         return None
 
-    def check_binary(self, pkg):
+    def _in_digest_configuration_group(self, pkgfile):
+        if stat.S_ISDIR(pkgfile.mode):
+            return False
+        return self._get_digest_location_from_path(pkgfile.name)
+
+    def _check_symlinks(self, pkg):
```
mgerstner, line 46: It doesn't only check for symlinks but also for ghost files. Maybe name the function something more generic like `_check_filetypes()` or so.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -34,46 +38,90 @@ def _get_digest_group_from_path(self, path):
...
-                self.output.add_info('W', pkg, f'{group}-file-symlink', filename)
+                self.output.add_info('E', pkg, f'{group}-file-symlink', filename)
+                result = False
                 continue
 
             pkgfile = pkg.readlink(pkgfile)
             if not pkgfile:
-                self.output.add_info('W', pkg, f'{group}-file-symlink', filename, 'broken symlink')
-                continue
+                self.output.add_info('E', pkg, f'{group}-file-symlink', filename, 'broken symlink')
```
mgerstner, line 69: The symlink is not necessarily broken, but probably a symlink outside the package. It might become necessary to support such situations. With rpmlint there is probably no way to check such cross-package symlinks, bet we would at least need to be able to whitelist such a case using the 'skip' algorithm.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -34,46 +38,90 @@ def _get_digest_group_from_path(self, path):
...
+                return False
+        return True
+
+    def check_binary(self, pkg):
+        """
+        Check that all files in secured locations are covered by a file digest group
+        in which all files have valid digest.
+        """
+
+        if not self._check_symlinks(pkg):
```
mgerstner, line 104: There are now two linear passes over all package files, here in `_check_symlinks` and below in `secured_paths = ...`. Could this be an effiency problem in large packages?

rpmlint/checks/FileDigestCheck.py
```diff
@@ -34,46 +38,90 @@ def _get_digest_group_from_path(self, path):
...
+        # First collect all files that are in a digest configuration group
+        secured_paths = {pkgfile.name for pkgfile in pkg.files.values() if self._in_digest_configuration_group(pkgfile)}
+
+        # Iterate all digest groups and find one that covers all secured files
+        # and in which all digests match
+        for digest_group in self.digest_groups:
+            covered_files = set(digest_group.keys())
+            if secured_paths - covered_files:
+                # bail out as some secured paths are not covered
+                continue
```
mgerstner, line 116: There is are now missing precise diagnostic messages towards the packager. For example for the `sarg` package case with the related `/usr/bin/sarg-reports` digests we would get three error messages for files in /etc/cron.daily, /etc/cron.weekly, /etc/cron.monthly but no error message at all for /usr/bin/sarg-reports, if the digest for only /usr/bin/sarg-reports would be mismatched.

In our rpmlint1 check we remembered the verification results of the "most recent" whitelisting group and displayed these results as diagnostic towards the packager. So I would like to be able to show the packager:
- file a/b: wrong digest
- file c/d: missing digest
- etc.

The question is how we can now determine what the "most recent" whitelisting entry is. Starting with the highest boo# or bsc# number would probably be a good match. This would also safe us a couple of digest calculations in the "good" verification case, because we don't need to check outdated digests before we encounter the actually matching ones.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -34,46 +38,90 @@ def _get_digest_group_from_path(self, path):
...
+            elif covered_files - set(pkg.files.keys()):
+                # a group has a digest for a file that is not in the package
+                continue
+            elif self._check_group_digests(digest_group, pkg):
+                # we are done, we found a valid group
+                return
+
+        # Report errors
+        for filename in sorted(secured_paths):
+            group = self._get_digest_location_from_path(filename)
```
mgerstner, line 126: This is about the fourth time that the group for a path is calculated. First in `_check_symlinks()` (for all files of the package) then again in `_in_digest_configuration_group()` (for all files of the package), then again in `secured_paths = ...` (for all files of the package) and then finally here, only for the `secured_paths`. I think this could be streamlined some more.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -34,46 +38,90 @@ def _get_digest_group_from_path(self, path):
                     pass
         return None
 
-    def check_binary(self, pkg):
+    def _in_digest_configuration_group(self, pkgfile):
+        if stat.S_ISDIR(pkgfile.mode):
+            return False
+        return self._get_digest_location_from_path(pkgfile.name)
```
mgerstner, line 44: Since this function seems to want to return a boolean value, adding `is not None` might be a good idea here.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -34,46 +38,90 @@ def _get_digest_group_from_path(self, path):
...
+        Check that all symlinks point to a correct location and that
+        symlinks are allowed in a configuration group.
         """
-        known_digests = self.package_digests.get(pkg.name)
+        result = True
         for filename, pkgfile in pkg.files.items():
-            group = self._get_digest_group_from_path(filename)
-            if not group:
+            group = self._get_digest_location_from_path(filename)
+            if not self._in_digest_configuration_group(pkgfile):
```
mgerstner, line 54: this is a bit strange, we already know whether a group matches this filename (`group != None`), but then in `in_digest_configuration_group` we check that again. Maybe call this rather something like `_is_restricted_file(pkgfile`) and only check the file type in there.

configs/openSUSE/security.toml
marxin: I've used a better TOML syntax, is it still a problem?

configs/openSUSE/security.toml
marxin: Should be much better now. Is it fine now?

rpmlint/checks/FileDigestCheck.py
marxin: Yes.

rpmlint/checks/FileDigestCheck.py
marxin: Yes.

rpmlint/checks/FileDigestCheck.py
marxin: All right, I fixed that.

rpmlint/checks/FileDigestCheck.py
marxin: No, it shouldn't be a problem.

rpmlint/checks/FileDigestCheck.py
marxin: Yep, I've added cache for it and renamed the function.

rpmlint/checks/FileDigestCheck.py
marxin: Hmm, to be honest doing proper diagnostics is quite tricky. I don't think the "most recent" bug is always what you want: in the case of a SLE-produce, I would expect an older bug to be used. All that can be achieved is a hint and I'm curious if worth for the compilation. Moreover, the pass now does not compare digests if it's obvious that a configuration group does not cover all "secured" files.

configs/openSUSE/security.toml
mgerstner: Hmm actually I can't see the difference to before. Where is the package coupling now?

configs/openSUSE/security.toml
mgerstner: Also here I can't see a difference. Am I looking at an oudated version or did you forget to commit something?

rpmlint/checks/FileDigestCheck.py
mgerstner: For SLE products we have our dedicated branches where we only backport individual whitelistings so the assumption should typically hold true.

From my experience with working with packagers and my coworkers exactly these bits of information are important and can save an hour or so of trying to figure out what is going on. And it is especially confusing when e.g. an error like described above is shown for /etc/cron.weekly/sarg-something while actually there is a valid digest for it.

jsegitz: I agree with mgerstner. We have to consider here that this pops up for users that might have never had any experience with rpmlint. So having error message to be as clear as possible is a high priority in my opinion. And as we had this for a while in rpmlint(1) it would be sad to lose it in the rewrite. I understand the strive for clean code but that is something that would cause pain for us and others over a long time.

configs/openSUSE/security.toml
marxin: Now please take a look at the changes for `test.config` where you'll see the new format.

rpmlint/checks/FileDigestCheck.py
marxin: Be my guest. I improved the error handling and now should be reported a group for which we have a minimal number of errors. That should work as the best hint for rpmlint consumers.

mgerstner:
I find it a bit confusing with many things going on in parallel in the review/discussion. Maybe we should first agree on the suitable configuration structure before continuing with the other details. I still don't find the current suggestion very intuitive. I dived a bit deeper into what TOML provides and have come up with this approach:

```
[[FileDigestGroup."sarg"]]
note = "some package with a lot of cron jobs"

    [FileDigestGroup."sarg".audits."bsc#4711"]
    digests = [
	{
		path = "/etc/cron.daily/suse.de-sarg",
		algorithm = "sha256",
		hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
	},
	{
		path = "/etc/cron.weekly/suse.de-sarg",
		algorithm = "sha256",
		hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
	}
    ]

    [FileDigestGroup."sarg".audits."bsc#4712"]
    digests = [
	{
		path = "/etc/cron.weekly/suse.de-sarg",
		algorithm = "sha256",
		hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
	},
	{
		path = "/etc/cron.weekly/suse.de-sarg",
		algorithm = "sha256",
		hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
	}
    ]

[[FileDigestGroup."someotherpackage"]]
note = "some other package"

    [FileDigestGroup."someotherpackage".audits."bsc#815"]
    digests = [
	{
		path = "/some/path",
		algorithm = "skip"
	},
	{
		path = "/root/somewhere",
		algorithm = "sha256",
		hash = "d536dc687798189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
	}
    ]
```

This approach still repeats the package names in the sub-tables, but this is due to a limitation of TOML not allowing newlines in inline-tables, see toml-lang/toml#516. The resulting python data structure looks like this:

```
{'FileDigestGroup': {'sarg': [{'audits': {'bsc#4711': {'digests': [{'algorithm': 'sha256',
                                                                    'hash': 'd536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359',
                                                                    'path': '/etc/cron.daily/suse.de-sarg'},
                                                                   {'algorithm': 'sha256',
                                                                    'hash': 'd536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359',
                                                                    'path': '/etc/cron.weekly/suse.de-sarg'}]},
                                          'bsc#4712': {'digests': [{'algorithm': 'sha256',
                                                                    'hash': 'd536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359',
                                                                    'path': '/etc/cron.weekly/suse.de-sarg'},
                                                                   {'algorithm': 'sha256',
                                                                    'hash': 'd536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359',
                                                                    'path': '/etc/cron.weekly/suse.de-sarg'}]}},
                               'note': 'some package with a lot of cron jobs'}],
                     'someotherpackage': [{'audits': {'bsc#815': {'digests': [{'algorithm': 'skip',
                                                                               'path': '/some/path'},
                                                                              {'algorithm': 'sha256',
                                                                               'hash': 'd536dc687798189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359',
                                                                               'path': '/root/somewhere'}]}},
                                           'note': 'some other package'}]}}
```

What do you think?

marxin:
I like the suggested format with one small nit: replacing `[[FileDigestGroup."sarg"]]` with `[FileDigestGroup."sarg"]`.

I've just pushed a patch that supports your format.

test/configs/test.config
```diff
@@ -51,24 +51,29 @@ Locations = [
     "/usr/share/polkit-1/rules.d/"
 ]
 
-[[FileDigestGroup]]
-package = "file-signature"
-bug = "boo#12345"
+[FileDigestGroup."file-signature"]
```
mgerstner, line 54: Again we are missing the "package".  When you look at my suggested format I had the package encoded here. So how is this supposed to work?

marxin: Sorry, it's maybe confusing, but `file-signature` is the name of the package. What will not work?

mgerstner: Ah, sorry, I got confused by the "file-signature" label, maybe using a more descriptive name in the test would avoid this.

marxin: Fixed that, sorry for the confusion.

marxin:
@mgerstner Do you feel the current shape of `FileDigestCheck` supports what you need?

mgerstner:
Okay the configuration syntax looks good now.

The check still behaves erroneous, though. `_calculate_errors_for_digest_group()` does not consider `self.digest_configurations`, i.e. it applies digest groups to all files, not only the files that start with the expected prefix path.

In the `-file-digest-mismatch` error case I'd like to see the expected and encountered digests.

I also get a KeyError in line 111, because the check tries to access a file in the package that isn't even existing. Is there a possibility to show the full exception backtrace when running rpmlint and a check fails internally? Because this error here was not very helpful:

```
$ rpmlint /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/ruby2.7-rubygem-arel-doc-9.0.0-2.10.x86_64.rpm
(none): E: fatal error while reading /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/ruby2.7-rubygem-arel-doc-9.0.0-2.10.x86_64.rpm: '/usr/sbin/sarg-reports'
```

For contextual files like /usr/sbin/sarg-reports the reported "group" is None, because the prefix of this path is not in a cron directory. It could make sense to assign each FileDigestGroup a type. Now that I think about it, something like this will most certainly become necessary, because we need to be able to keep parallel whitelists for cron, systemd-tmpfiles etc. Say we want to keep separate xml files for them:

```
$ cat /etc/xdg/rpmlint/cron_whitelist.xml

    [FileDigestLocation.cron]
    FollowSymlinks = true
    Locations = [
        "/etc/cron.d/",
        "/etc/cron.hourly/",
        "/etc/cron.daily/",
        "/etc/cron.weekly/",
        "/etc/cron.monthly/"
    ]

    [FileDigestGroup."sarg"]
    note = """
    Builds statistics based on Squid logfile metadata.
    Include sarg-reports which is SUSE specific and important for privilege dropping.
    """
      [FileDigestGroup."sarg".audits."bsc#4712"]
      digests = [
        {
          path = "/etc/cron.daily/suse.de-sarg",
          algorithm = "sha256",
          hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
        },
        {
          path = "/etc/cron.monthly/suse.de-sarg",
          algorithm = "sha256",
          hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
        },
        {
          path = "/etc/cron.weekly/suse.de-sarg",
          algorithm = "sha256",
          hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
        },
        {
          path = "/usr/sbin/sarg-reports",
          algorithm = "sha256",
          hash = "00ad25400adc2031cd09f9b8f9e56c448c93b6b89a702d36dce6a385d79e637c"
        },
      ]
```

```
$cat /etc/xdg/rpmlint/tmpfiles_whitelist.xml

    [FileDigestLocation."systemd-tmpfiles"]
    FollowSymlinks = true
    Locations = [
        "/usr/lib/tmpfiles.d/"
    ]

    [FileDigestGroup."sarg"]
    note = """some sarg tmpfiles.d entry"""

      [FileDigestGroup."sarg".audits."bsc#1234"]
      digests  = [
        {
          path = "/usr/lib/tmpfiles.d/sarg.conf",
          algorithm = "sha256",
          hash = "<some-hash>"
        }
      ]
```

These two entries should be treated seperately. So we need to encode the type
of whitelisting somehow in the FileDigestGroup, something like `[FileDigestGroup.cron."sarg"]` and `[FileDigestGroup."systemd-tmpfiles"."sarg"]`.

marxin:
> Okay the configuration syntax looks good now.
> 
> The check still behaves erroneous, though. `_calculate_errors_for_digest_group()` does not consider `self.digest_configurations`, i.e. it applies digest groups to all files, not only the files that start with the expected prefix path.

Ok, the function `_calculate_errors_for_digest_group()` reports 3 types of errors now:
-  `# report errors for secured files not covered by the digest group`
-  `# report errors for missing files mentioned in the digest group`
-  `# report errors for invalid digests`

Which one is not good, or please show me an example.

> 
> In the `-file-digest-mismatch` error case I'd like to see the expected and encountered digests.

Sure, will fix that.

> 
> I also get a KeyError in line 111, because the check tries to access a file in the package that isn't even existing. Is there a possibility to show the full exception backtrace when running rpmlint and a check fails internally? Because this error here was not very helpful:
> 
> ```
> $ rpmlint /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/ruby2.7-rubygem-arel-doc-9.0.0-2.10.x86_64.rpm
> (none): E: fatal error while reading /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/ruby2.7-rubygem-arel-doc-9.0.0-2.10.x86_64.rpm: '/usr/sbin/sarg-reports'
> ```

Sure, just apply the following patch:
```patch
diff --git a/rpmlint/lint.py b/rpmlint/lint.py
index 1d75625..419a629 100644
--- a/rpmlint/lint.py
+++ b/rpmlint/lint.py
@@ -255,6 +255,7 @@ class Lint(object):
                     self.run_checks(pkg)
         except Exception as e:
             print_warning(f'(none): E: fatal error while reading {pname}: {e}')
+            raise e
             sys.exit(3)
 
     def run_checks(self, pkg):
```

> 
> For contextual files like /usr/sbin/sarg-reports the reported "group" is None, because the prefix of this path is not in a cron directory. It could make sense to assign each FileDigestGroup a type. Now that I think about it, something like this will most certainly become necessary, because we need to be able to keep parallel whitelists for cron, systemd-tmpfiles etc. Say we want to keep separate xml files for them:

Oh yes, that's really a limitation.

> 
> ```
> $ cat /etc/xdg/rpmlint/cron_whitelist.xml
> 
>     [FileDigestLocation.cron]
>     FollowSymlinks = true
>     Locations = [
>         "/etc/cron.d/",
>         "/etc/cron.hourly/",
>         "/etc/cron.daily/",
>         "/etc/cron.weekly/",
>         "/etc/cron.monthly/"
>     ]
> 
>     [FileDigestGroup."sarg"]
>     note = """
>     Builds statistics based on Squid logfile metadata.
>     Include sarg-reports which is SUSE specific and important for privilege dropping.
>     """
>       [FileDigestGroup."sarg".audits."bsc#4712"]
>       digests = [
>         {
>           path = "/etc/cron.daily/suse.de-sarg",
>           algorithm = "sha256",
>           hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
>         },
>         {
>           path = "/etc/cron.monthly/suse.de-sarg",
>           algorithm = "sha256",
>           hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
>         },
>         {
>           path = "/etc/cron.weekly/suse.de-sarg",
>           algorithm = "sha256",
>           hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
>         },
>         {
>           path = "/usr/sbin/sarg-reports",
>           algorithm = "sha256",
>           hash = "00ad25400adc2031cd09f9b8f9e56c448c93b6b89a702d36dce6a385d79e637c"
>         },
>       ]
> ```
> 
> ```
> $cat /etc/xdg/rpmlint/tmpfiles_whitelist.xml
> 
>     [FileDigestLocation."systemd-tmpfiles"]
>     FollowSymlinks = true
>     Locations = [
>         "/usr/lib/tmpfiles.d/"
>     ]
> 
>     [FileDigestGroup."sarg"]
>     note = """some sarg tmpfiles.d entry"""
> 
>       [FileDigestGroup."sarg".audits."bsc#1234"]
>       digests  = [
>         {
>           path = "/usr/lib/tmpfiles.d/sarg.conf",
>           algorithm = "sha256",
>           hash = "<some-hash>"
>         }
>       ]
> ```
> 
> These two entries should be treated seperately. So we need to encode the type
> of whitelisting somehow in the FileDigestGroup, something like `[FileDigestGroup.cron."sarg"]` and `[FileDigestGroup."systemd-tmpfiles"."sarg"]`.

What about:
```
[FileDigestGroup."squid"]
type = "cron"
note = """
Builds statistics based on Squid logfile metadata.
Include sarg-reports which is SUSE specific and important for privilege dropping.
"""
```

where when checking the `FileDigestGroup."squid"` we can report all errors with the prefix taken from `type` attribute.
That can be applied to all files mentioned in the group, including e.g. `/usr/sbin/sarg-reports`. Is it a viable approach?
It can simplify the check quite well.

mgerstner:
> > The check still behaves erroneous, though. `_calculate_errors_for_digest_group()` does not consider `self.digest_configurations`, i.e. it applies digest groups to all files, not only the files that start with the expected prefix path.
> 
> Ok, the function `_calculate_errors_for_digest_group()` reports 3 types of errors now:
> 
> * `# report errors for secured files not covered by the digest group`
> * `# report errors for missing files mentioned in the digest group`
> * `# report errors for invalid digests`
> 
> Which one is not good, or please show me an example.

Maybe I didn't express the problem good enough. The problem is that even if a package doesn't
contain the files listed in the whitelisting, an error will be displayed. Consider this example:

```
# zypper download ruby2.7-rubygem-arel-doc
# rpmlint /var/cache/zypp/packages/^Cwnload.opensuse.org-oss_2/x86_64/ruby2.7-rubygem-arel-doc-9.0.0-2.10.x86_64.rpm
[...]
ruby2.7-rubygem-arel-doc.x86_64: E: unknown-file-digest-unauthorized /etc/cron.daily/suse.de-sar
ruby2.7-rubygem-arel-doc.x86_64: E: unknown-file-digest-unauthorized /etc/cron.monthly/suse.de-sarg
ruby2.7-rubygem-arel-doc.x86_64: E: unknown-file-digest-unauthorized /etc/cron.weekly/suse.de-sarg
ruby2.7-rubygem-arel-doc.x86_64: E: unknown-file-digest-unauthorized /usr/sbin/sarg-reports
[...]
```

> What about:
> 
> ```
> [FileDigestGroup."squid"]
> type = "cron"
> note = """
> Builds statistics based on Squid logfile metadata.
> Include sarg-reports which is SUSE specific and important for privilege dropping.
> """
> ```
> 
> where when checking the `FileDigestGroup."squid"` we can report all errors with the prefix taken from `type` attribute.
> That can be applied to all files mentioned in the group, including e.g. `/usr/sbin/sarg-reports`. Is it a viable approach?
> It can simplify the check quite well.

Yes it sounds good.

marxin:
> Maybe I didn't express the problem good enough. The problem is that even if a package doesn't
> contain the files listed in the whitelisting, an error will be displayed. Consider this example:

You are right, that were bogus errors. I've just pushed a version with simplified logic that respects the newly added `type` property.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -102,20 +103,24 @@ def _calculate_errors_for_digest_group(self, pkg, digest_group, secured_paths):
...
             filename = digest['path']
             if filename in pkg_files:
                 group = self._get_digest_configuration_group(pkg.files[filename])
-                if not self._is_valid_digest(pkg.files[filename], digest, pkg):
-                    errors.append((f'{group}-file-digest-mismatch', filename))
+                valid_digest, file_digest = self._is_valid_digest(pkg.files[filename], digest, pkg)
+                if not valid_digest:
+                    error_detail = None
+                    if file_digest:
+                        error_detail = f'expected:{digest["hash"]}, has:{file_digest}'
```
mgerstner, line 113: Could you please also add the digest algorithm used to the output? So it is clear to everybody which kind of digest this is.

mgerstner:
Okay we're getting close, however there is still a problem with the `type` property. When I want to add two different types of restrictions to the same package:

```
[FileDigestGroup."sarg"]                                                          
type = "cron"                                                                     
note = """                                                                        
Builds statistics based on Squid logfile metadata.                                
Include sarg-reports which is SUSE specific and important for privilege dropping.
"""                                                                               
  [FileDigestGroup."sarg".audits."bsc#4711"]                                      
  digests = [                                                                     
    {                                                                             
      path = "/etc/cron.daily/suse.de-sarg",                                      
      algorithm = "sha256",                                                       
      hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
    }                                                                             
  ]

[FileDigestGroup."sarg"]                                                          
type = "apache"                                                                   
note = """Indians are around"""                                                
  [FileDigestGroup."sarg".audits."bsc#815"]                                    
  digests = [                                                                  
    {                                                                          
      path = "/etc/apache3/something",                                         
      algorithm = "sha256",                                                    
      hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
    }                                                                          
  ]
```

Then I receive an error:

```
rpmlint /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/sarg-2.4.0-3.3.x86_64.rpm
(none): W: error parsing configuration files: What? sarg already exists?
```

So we would need to change the data structure to allow for this.

marxin:
> Then I receive an error:
> 
> ```
> rpmlint /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/sarg-2.4.0-3.3.x86_64.rpm
> (none): W: error parsing configuration files: What? sarg already exists?
> ```
> 
> So we would need to change the data structure to allow for this.

That's really problem. It will need something like:
https://github.com/rpm-software-management/rpmlint/blob/f6f767173ed42d1734ab2dfde9e8703372ebb2f6/test/configs/test.config#L60

However, I'm considering flattening of the `[[FileDigestGroup]]`. What do you think about it?
It seems easier to edit and read the rules.

mgerstner:
> > Then I receive an error:
> > ```
> > rpmlint /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/sarg-2.4.0-3.3.x86_64.rpm
> > (none): W: error parsing configuration files: What? sarg already exists?
> > ```
> > 
> > 
> > So we would need to change the data structure to allow for this.
> 
> That's really problem. It will need something like:
> https://github.com/rpm-software-management/rpmlint/blob/f6f767173ed42d1734ab2dfde9e8703372ebb2f6/test/configs/test.config#L60
> 
> However, I'm considering flattening of the `[[FileDigestGroup]]`. What do you think about it?
> It seems easier to edit and read the rules.

Hmm, I don't see big advantages or disadvantages. The hierarchical approach before made it clearer that the different digest groups are related IMO but I can also live with this new approach.

It looks like there is a regression now, though: The package field is not evaluated, so the digests are considered for all packages. Can you maybe add this to your tests that the config only is evaluated for matching package names?

Also I noticed the check doesn't work when there are not `[[FileDigestGroup]]` entries at all, it fails with a `KeyError`.

marxin:
> Hmm, I don't see big advantages or disadvantages. The hierarchical approach before made it clearer that the different digest groups are related IMO but I can also live with this new approach.

It will help as the implementation can be simplified a bit.

> 
> It looks like there is a regression now, though: The package field is not evaluated, so the digests are considered for all packages. Can you maybe add this to your tests that the config only is evaluated for matching package names?

Fixed that and it's now newly tests.

> 
> Also I noticed the check doesn't work when there are not `[[FileDigestGroup]]` entries at all, it fails with a `KeyError`.

I added `[FileDigestGroup]` to `configdefaults.toml` which provides empty default. Note that also other config values must have a default value.

mgerstner:
The package is now correctly take in to account.

I'm sorry to say it still doesn't work as expected though. Having two separate check types doesn't work out, the problem must be somewhere in line 131 `secured_paths = {...}`, where paths from different check types are mixed together. I guess each check type must be considered separately. Example:

```
[FileDigestLocation.cron]                                                         
FollowSymlinks = true                                                             
Locations = [                                                                     
    "/etc/cron.d/",                                                               
    "/etc/cron.hourly/",                                                          
    "/etc/cron.daily/",                                                           
    "/etc/cron.weekly/",                                                          
    "/etc/cron.monthly/"                                                          
]                                                                                 
                                                                                  
[FileDigestLocation.apache]                                                       
FollowSymlinks = true                                                             
Locations = [                                                                     
    "/etc/apache2/conf.d",                                                        
] 

[[FileDigestGroup]]                                                            
package = "sarg"                                                               
type = "cron"                                                                  
note = """                                                                     
Builds statistics based on Squid logfile metadata.                             
Include sarg-reports which is SUSE specific and important for privilege dropping.
"""                                                                            
bug = "bsc#4712"                                                               
digests = [                                                                    
{                                                                              
        path = "/etc/cron.daily/suse.de-sarg",                                 
             algorithm = "sha256",                                             
             hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
},                                                                             
{                                                                              
        path = "/etc/cron.monthly/suse.de-sarg",                               
        algorithm = "sha256",                                                  
        hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
},                                                                             
{                                                                              
        path = "/etc/cron.weekly/suse.de-sarg",                                
        algorithm = "sha256",                                                  
        hash = "d536dc68e198189149048a907ea6d56a7ee9fc732ae8fec5a4072ad06640e359"
},                                                                             
{                                                                              
        path = "/usr/sbin/sarg-reports",                                       
        algorithm = "sha256",                                                  
        hash = "00ad25400bdc2031cd09f9b8f9e56c448c93b6b89a702d36dce6a385d79e637c"
}                                                                             
]

[[FileDigestGroup]]                                                            
package = "sarg"                                                               
type = "apache"                                                                
bug = "bsc#4713"                                                               
digests = [                                                                    
{                                                                              
     path = "/etc/apache2/conf.d/sarg-apache.conf",                            
     algorithm = "skip"
}                                                                              
]
```

This results in the following error:

```
root# rpmlint /var/cache/zypp/packages/download.opensuse.org-oss_2/x86_64/sarg-2.4.0-3.3.x86_64.rpm
sarg.x86_64: E: cron-file-digest-unauthorized /etc/apache2/conf.d/sarg-apache.conf
```

So the apache config file is treated as being part of the cron type check and the apache type whitelisting isn't used.

marxin:
> So the apache config file is treated as being part of the cron type check and the apache type whitelisting isn't used.

I'm sorry that it still does not work as expected :/

It's being caused by fact that we require that all package files (in a secured file digest location) must be covered by **exactly one** `[[FileDigestGroup]]`. How can I find out that multiple groups (with equal `package` and a different `type`) successfully cover files from a package?

mgerstner:
> > So the apache config file is treated as being part of the cron type check and the apache type whitelisting isn't used.
> 
> I'm sorry that it still does not work as expected :/
> 
> It's being caused by fact that we require that all package files (in a secured file digest location) must be covered by **exactly one** `[[FileDigestGroup]]`. How can I find out that multiple groups (with equal `package` and a different `type`) successfully cover files from a package?

Groups of different type should not have any influence on each other. For each restriction type exactly one group of the same type should fully match.

marxin:
> Groups of different type should not have any influence on each other. For each restriction type exactly one group of the same type should fully match.

I see that makes sense. I fixed that and tested it on the package. We'll definitely need a better test coverage I guess.

mgerstner:
> > Groups of different type should not have any influence on each other. For each restriction type exactly one group of the same type should fully match.
> 
> I see that makes sense. I fixed that and tested it on the package. We'll definitely need a better test coverage I guess.

Yes a more complex test case that checks all the corner cases would surely be helpful.

Okay the current state seems to finally work for the FileDigestCheck. Now there remains the question about the "error details". In rpmlint-1 we have `Filter.addDetails()` that we used to provide detailed instructions to packagers what to do about whitelisting issues. I implemented placeholders to avoid repeating the same paragraphs over and over again, see [here](https://github.com/openSUSE/rpmlint-checks/blob/master/Whitelisting.py#L26). We provide a wiki URL to packagers here so they can always refer to up to date information how the review process works etc. It looks like rpmlint-1 only prints those details when invoked with the `-I` / `--explain` switch, which is obivously what happens when rpmlint runs in the OBS context.

How can we achieve this in rpmlint-2?

marxin:
> How can we achieve this in rpmlint-2?

This can be easily achieved by adding a description to `rpmlint/descriptionsFileDigestCheck`. rpmlint-2 will be run in `-v` mode, so the explanation will be printed.

Similarly one can see:
```
$ ./lint.py  /home/marxin/BIG/rpm/sarg/binaries/sarg-2.4.0-3.3.x86_64.rpm -v
...

sarg.x86_64: W: manpage-not-compressed bz2 /usr/share/man/man1/sarg.1.gz
sarg.x86_64: W: manpage-not-compressed bz2 /usr/share/man/man8/sarg-reports.1.gz
This manual page is not compressed with the bz2 compression method
(does not have the bz2 extension). If the compression does not happen
automatically when the package is rebuilt, make sure that you have the
appropriate rpm helper and/or config packages for your target distribution
installed and try rebuilding again; if it still does not happen automatically,
you can compress this file in the %install section of the spec file.

sarg.x86_64: W: invalid-license GPL-2.0-or-later
The value of the License tag was not recognized.  Known values are:
''.

sarg.x86_64: E: incorrect-fsf-address /srv/www/htdocs/sarg-php/COPYING
sarg.x86_64: E: incorrect-fsf-address /usr/sbin/sarg-reports
sarg.x86_64: E: incorrect-fsf-address /usr/share/licenses/sarg/COPYING
sarg.x86_64: E: incorrect-fsf-address /usr/share/licenses/sarg/LICENSE
The Free Software Foundation address in this file seems to be outdated or
misspelled. Ask upstream to update the address, or if this is a license file,
possibly the entire file with a new copy available from the FSF.

```

mgerstner:
> > How can we achieve this in rpmlint-2?
> 
> This can be easily achieved by adding a description to `rpmlint/descriptionsFileDigestCheck.toml`. rpmlint-2 will be run in `-v` mode, so the explanation will be printed.

Okay so the mechanism is clear. However we would like to add the openSUSE wiki URL to the explanation and we also would like to avoid redundancy. Currently we do this by using some [generic text snippets](https://github.com/openSUSE/rpmlint-checks/blob/master/Whitelisting.py#L14) that are `format`ted during runtime.

Since this is now a TOML file where the descriptions reside, doing it during runtime is not so easy. We could generate the description file from the repository via a script or something but that would complicate the installation somewhat. Any ideas?

marxin:
> Okay so the mechanism is clear. However we would like to add the openSUSE wiki URL to the explanation and we also would like to avoid redundancy. Currently we do this by using some [generic text snippets](https://github.com/openSUSE/rpmlint-checks/blob/master/Whitelisting.py#L14) that are `format`ted during runtime.
> 
> Since this is now a TOML file where the descriptions reside, doing it during runtime is not so easy. We could generate the description file from the repository via a script or something but that would complicate the installation somewhat. Any ideas?

I think it's a good idea to have snippets (variables). Should be addressed with 2c78ab1a455d98adb4591aae5576ffd8271ab1b7 that can be upstreamed later.

Hope it will work for your use case?

mgerstner:
Okay snippets in error descriptions look good this should allow us to do what we want.

Testing this I found yet another regression, however: When there are no digest groups present for a package then no error is raised at all (`return` in FileDigestCheck.py:412).

marxin:
> Okay snippets in error descriptions look good this should allow us to do what we want.

Good. Will you please prepare a commit that will add the descriptions?

> 
> Testing this I found yet another regression, however: When there are no digest groups present for a package then no error is raised at all (`return` in FileDigestCheck.py:412).

Fixed that.

I'm suggesting to move the unreviewed `FileMetadataCheck` check into a different pull request. What do you think?
If you are fine, I can squash the changes and make a real pull request for `FileDigestCheck`?

mgerstner:
> > Okay snippets in error descriptions look good this should allow us to do what we want.
> 
> Good. Will you please prepare a commit that will add the descriptions?

For this to be complete I would need to add configuration for all whitelisting restrictions we currently have in place, also add the actual whitelists as we already have them in rpmlint1 and then the error descriptions for each type of error we have. Would that be okay?
 
> > Testing this I found yet another regression, however: When there are no digest groups present for a package then no error is raised at all (`return` in FileDigestCheck.py:412).
> 
> Fixed that.

We really need some actual test coverage before we can bring that into production. We should implement various test cases:

- an RPM that does not contain files from any of the restricted locations should not trigger any errors
- an RPM that contains files in a restricted location but has no whitelisting entry should trigger an appropriate error
- an RPM that contains files in a restricted location but has one matching (among a choice of multiple groups of) whitelisting entry should not trigger an error.
- an RPM that contains files in a restricted location and a whitelisting entry contains matches but only for 2 of 3 files should trigger an error, and display the offending / missing file.
- an RPM that contains files in two different types of restricted locations and has matching whitelisting entries for both locations should not trigger an error.
- the symlink handling (follow, nofollow) should be covered by a testcase
- an RPM of a different (package) name should not be covered by a whitelisting entry, even if the digests match.

Probably I missed a couple of situations there.
 
> I'm suggesting to move the unreviewed `FileMetadataCheck` check into a different pull request. What do you think?
> If you are fine, I can squash the changes and make a real pull request for `FileDigestCheck`?

Yes that is a good idea, because this took quite a lot of time and we should "save" the progress we have made. Before we do that we should cover the points above, however, and I also want to involve Johannes to test the final variant of the checker, because I'm starting to get blind myself about this and we need a final fresh view on things before we move it into production.

marxin:
> For this to be complete I would need to add configuration for all whitelisting restrictions we currently have in place, also add the actual whitelists as we already have them in rpmlint1 and then the error descriptions for each type of error we have. Would that be okay?
> 

Yes. We can likely make it a separate pull request to `opensuse` branch once this is merged.
Note we still have some time before `rpmlint2` will be used in `openSUSE:Factory`.

marxin:
> Probably I missed a couple of situations there.
> 

Sure, please submit such packages to https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests and then add the created `rpm` files to `tests`. Leaving that to your team.

marxin:
> Yes that is a good idea, because this took quite a lot of time and we should "save" the progress we have made.

Good, I've just done that in #600.

mgerstner:
> > For this to be complete I would need to add configuration for all whitelisting restrictions we currently have in place, also add the actual whitelists as we already have them in rpmlint1 and then the error descriptions for each type of error we have. Would that be okay?
> 
> Yes. We can likely make it a separate pull request to `opensuse` branch once this is merged.
> Note we still have some time before `rpmlint2` will be used in `openSUSE:Factory`.

Maybe we should at the moment just add some example configuration (commented out) so that users can quickly understand how the configuration os the checker is supposed to work. This can be part of the standard rpmlint, while the actual production configuration we need are then added to the opensuse branch.

mgerstner:
> > Probably I missed a couple of situations there.
> 
> Sure, please submit such packages to https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests and then add the created `rpm` files to `tests`. Leaving that to your team.

Can you explain to us how the connection between these test packages and the tests within the `test` sub-directory of rpmlint works? So should we only add the packages into OBS or also write the actual test?

marxin:
> Can you explain to us how the connection between these test packages and the tests within the `test` sub-directory of rpmlint works? So should we only add the packages into OBS or also write the actual test?

In order to create a rpm test-case, one needs something like OBS. So that's why we track .spec files in the `devel:openSUSE:Factory:rpmlint:tests `. And the built rpm packages are taken and copied into `test` folder in rpmlint.
And yes, you are supposed to write Python tests in `test/test_xyz.py`.

marxin:
> Maybe we should at the moment just add some example configuration (commented out) so that users can quickly understand how the configuration os the checker is supposed to work. This can be part of the standard rpmlint, while the actual production configuration we need are then added to the opensuse branch.

To be honest, based on the experience with #439, I don't think upstream is going to accept the pass.
It's quite openSUSE-specific and thus I recommend putting this entire pull request into `opensuse` branch.

mgerstner:
Okay I have informed @jsegitz about the current state of the check and asked him to look at it from his point of view if usability, feature set and robustness are all right. I am waiting for this statement before I continue.

Which process should be follow to complete the integration? I can create commits to add the error descriptions and port the actual production whitelisting entries from rpmlint1. Should I do this as a separate PR# once this PR# is merged?

marxin:
> Which process should be follow to complete the integration?

Currently, we do face problems with extensive dependencies of the patch:
https://jira.suse.com/browse/OBS-88

I hope @kstreitova can make a progress with taht.

> I can create commits to add the error descriptions and port the actual production whitelisting entries from rpmlint1. Should I do this as a separate PR# once this PR# is merged?

I guess so.

May I please ask @kstreitova for a review of this pull request?

mgerstner:
@jsegitz also pointed out that rpmlint is behaving badly when there are syntax errors in TOML configuration files like `/etc/xdg/rpmlint/security.toml`. Errors will look like:

```
(none): W: error parsing configuration files: Invalid inline table encountered (line 46 column 1 char 886)
```

So its not clear in which file exactly the error occured. And even worse, the rpmlint processing continues, bit without the security related checks. In such a case we need rpmlint to abort with an error such that package with potential dangerous content don't build successfully in protected projects on OBS.

Another things we noted is that at least in the packaging that you have in `home:marxin:rpmlint-ring0/rpmlint`, the configuration file in `/etc/xdg/rpmlint` all carry an executable bit.

marxin:
> @jsegitz also pointed out that rpmlint is behaving badly when there are syntax errors in TOML configuration files like `/etc/xdg/rpmlint/security.toml`. Errors will look like:

Very good point, I've just made #603 for it:

> 
> ```
> (none): W: error parsing configuration files: Invalid inline table encountered (line 46 column 1 char 886)
> ```

This is replaced with:
`(none): E: fatal error while parsing configuration file configs/openSUSE/opensuse.toml: Unbalanced quotes (line 122 column 29 char 4640)`

> 
> So its not clear in which file exactly the error occured. And even worse, the rpmlint processing continues, bit without the security related checks. In such a case we need rpmlint to abort with an error such that package with potential dangerous content don't build successfully in protected projects on OBS.

... and rpmlint prints a fatal error and ends.

> 
> Another things we noted is that at least in the packaging that you have in `home:marxin:rpmlint-ring0/rpmlint`, the configuration file in `/etc/xdg/rpmlint` all carry an executable bit.

Fixed here: `home:scarabeus_iv:rpmlint-testbed/rpmlint`.

marxin:
> We really need some actual test coverage before we can bring that into production. We should implement various test cases:
> 
> * an RPM that does not contain files from any of the restricted locations should not trigger any errors
> * an RPM that contains files in a restricted location but has no whitelisting entry should trigger an appropriate error
> * an RPM that contains files in a restricted location but has one matching (among a choice of multiple groups of) whitelisting entry should not trigger an error.
> * an RPM that contains files in a restricted location and a whitelisting entry contains matches but only for 2 of 3 files should trigger an error, and display the offending / missing file.
> * an RPM that contains files in two different types of restricted locations and has matching whitelisting entries for both locations should not trigger an error.
> * the symlink handling (follow, nofollow) should be covered by a testcase
> * an RPM of a different (package) name should not be covered by a whitelisting entry, even if the digests match.

Btw. I've just extended `FakePkg` syntax and one can cover all these situations with tests that don't need an RPM file. Please see:
https://github.com/rpm-software-management/rpmlint/blob/a7f460d6d8fce2d725066137ce1f5aa9d16287bc/test/test_file_digest.py#L49-L59

PR #597, opened 2021-01-10
Introduce GitHub workflow.

marxin:


.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
+name: Build and Test
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
scarabeusiv, line 7: Shouldn't it monitor also the openSUSE branch itself?

marxin: Good point!

scarabeusiv:
LGTM

PR #599, opened 2021-01-11
Detect Go binaries and skip some checks for them.

### LINKED ISSUE #598: E: shared-lib-without-dependency-information /usr/bin/xxx for GO binaries
I use Fedora 33 and rpmlint from the distro (rpmlint-1.11-12.fc33.noarch)

Fragment  of .spec file how I build my GO sources:
```sh
%build
export GOPATH=/usr/share/gocode
export GOBUILDFLAGS="%gobuildflags"
export CGO_ENABLED=0
mkdir bin
export GOBIN=$(pwd)/bin
eval "go install $GOBUILDFLAGS ./mysources/..."
```

Example of the binary:
[bug.zip](https://github.com/rpm-software-management/rpmlint/files/5794106/bug.zip)

marxin:
Fixes #598.


Conan-Kudo:
Does this make sense when Go binaries are built with standard Go macros? @jcajka, could you verify?

jcajka:
I don't think outright disabling the checks makes sense(as they from quick look on the python check script make sense to me). IIRC we use in fedora pie build mode of Go. That will generate pie binaries. In the linked example it should AFAIK in default mode produce static binaries(in C sense). I don't remember from top of my head how "static" pie (elf) C binaries looks like but I would assume that Go ones will look the same. I will double check.

jcajka:
> Does this make sense when Go binaries are built with standard Go macros? @jcajka, could you verify?

@Conan-Kudo  BTW do you remember to what and where are the "gobuildflags" set?

jcajka:
Other issue might be that the CGO got disabled for the build of the binaries, IMO the rpmlint can't realistically cater for such manual fine tuning, on other hand should probably handle it more gracefully.

marxin:
@Conan-Kudo What do you think about this pull request?

Conan-Kudo:
I'm not sure it makes sense to incorporate this into rpmlint, given @jcajka's feedback. Actually, I wonder if we might be doing something wrong in openSUSE that we're not getting PIE and debuginfo+build-ids enabled on Go builds there...

marxin:
Ok, so taking the provided example from the bug and my analysis:
https://github.com/rpm-software-management/rpmlint/issues/598#issuecomment-899292028

We report the following:
```
$ ./lint.py /tmp/example-1.0-1.utm.x86_64.rpm
...
example.x86_64: E: shared-library-without-dependency-information /usr/bin/example1
example.x86_64: E: shared-library-without-dependency-information /usr/bin/example2
...
```

So either the provided Go shared library is self-contained (does not link to any shared library) and it's fine or the error message is correct and we can close the reported issue?

What should I choose @jcajka ?

PR #600, opened 2021-01-12
[opensuse] Add FileMetadataCheck check.

marxin:
Second check separated from #594.

### DIFF
--- a/rpmlint/checks/WorldWritableCheck.py
+++ b/rpmlint/checks/WorldWritableCheck.py
@@ -10,6 +10,6 @@ class WorldWritableCheck(FileMetadataCheck):
 
     def check_binary(self, pkg):
         files = [FileMeta(pfile) for pfile in pkg.files.values()]
-        # select only block and character devices, ignore symlinks
-        files = {f.path: f for f in files if not f.mode[0] in 'bcs' and f.mode[-1] == 't'}
+        # select any world-writable files except device files and symlinks
+        files = {f.path: f for f in files if not f.mode[0] in 'bcl' and f.mode[-2] == 'w'}
         self.check_files(pkg, files)
--- a/rpmlint/checks/WorldWritableCheck.py
+++ b/rpmlint/checks/WorldWritableCheck.py
@@ -1,0 +1,15 @@
+from rpmlint.checks.FileMetadataCheck import FileMeta, FileMetadataCheck
+
+
+class WorldWritableCheck(FileMetadataCheck):
+    def __init__(self, config, output):
+        super().__init__(config, output)
+        self.whitelists = self.config.configuration['WorldWritableWhitelist']
+        self.verify_whitelists(['path', 'owner', 'group', 'mode'])
+        self.prefix = 'world-writable'
+
+    def check_binary(self, pkg):
+        files = [FileMeta(pfile) for pfile in pkg.files.values()]
+        # select only block and character devices, ignore symlinks
+        files = {f.path: f for f in files if not f.mode[0] in 'bcs' and f.mode[-1] == 't'}
+        self.check_files(pkg, files)
--- a/rpmlint/checks/DeviceFilesCheck.py
+++ b/rpmlint/checks/DeviceFilesCheck.py
@@ -1,0 +1,15 @@
+from rpmlint.checks.FileMetadataCheck import FileMeta, FileMetadataCheck
+
+
+class DeviceFilesCheck(FileMetadataCheck):
+    def __init__(self, config, output):
+        super().__init__(config, output)
+        self.whitelists = self.config.configuration['DeviceFilesWhitelist']
+        self.verify_whitelists(['path', 'owner', 'group', 'mode', 'device_minor', 'device_major'])
+        self.prefix = 'device'
+
+    def check_binary(self, pkg):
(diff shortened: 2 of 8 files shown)


marxin:
May I please ping this @mgerstner?

mgerstner:
> May I please ping this @mgerstner?

Sure ... I first wanted to add the tests for the other check but didn't find time yet. I can prioritize this review here instead.

mgerstner:
@marxin where can one get current rpmlint2 builds in OBS these days?

marxin:
It's here: https://build.opensuse.org/package/show/home:scarabeus_iv:rpmlint-testbed/rpmlint

mgerstner:
Well this check does not do at all yet what we (security team) intended. This check only triggers when a whitelisting entry *exists*, but it needs to trigger for any packaged file that matches certain criteria.

So currently we have `CheckDeviceFiles` and `CheckWorldWritable`. For device files any file that is a character or block device needs to trigger the check. The check should then only pass if a matching whitelisting entry for a file exists. Otherwise an error with badness should be raised. For world writable files any file that has a world writeable bit set should be handled equally. But symlinks and device files should be excluded for that check.

Similar to the file digest check we need to tie whitelisting entries to package names.

marxin:
> Well this check does not do at all yet what we (security team) intended. This check only triggers when a whitelisting entry _exists_, but it needs to trigger for any packaged file that matches certain criteria.

I see.

> 
> So currently we have `CheckDeviceFiles` and `CheckWorldWritable`. For device files any file that is a character or block device needs to trigger the check. The check should then only pass if a matching whitelisting entry for a file exists. Otherwise an error with badness should be raised. For world writable files any file that has a world writeable bit set should be handled equally. But symlinks and device files should be excluded for that check.

Can we do these 2 checks in one? I mean what about listing package files that are either a device (block or character) or have a writable file. For these, we should find a whitelisting entry that will list all the files and their metadata match?

Am I right that you expect a different set of flags of a file for each of the 2 checks you have?

```
					"/tmp/.ICE-unix": {
						"type": "d",
						"mode": "1777",
						"owner": "root:root"
					}
...
					"/var/lib/named/dev/random": {
						"type": "c",
						"mode": "0666",
						"dev": "1,8",
						"owner": "root:root"
					},

```

> 
> Similar to the file digest check we need to tie whitelisting entries to package names.

Yep, we pretty much know how to define a configuration for it.

mgerstner:
> Can we do these 2 checks in one? I mean what about listing package files that are either a device (block or character) or have a writable file. For these, we should find a whitelisting entry that will list all the files and their metadata match?

I am not quite sure yet what you mean. Is your question whether we should keep a single whitelist for both, device files and world writable files? I don't think this is so important. Currently in rpmlint1 we have two differently configured checks and thus two separate whitelists.

I don't think it is easy to come up with a generic metadata checker that catches both cases. What I currently do in rpmlint-checks is already a bit unclean but I managed to reuse the same checker code with different initialization data for device restrictions and for world-writable restrictions.

It would be good to be able to easily reuse the code to e.g. also implement the setuid-root binary restrictions and so on. But I believe still each type of restriction needs to have its own checker instantiation.
 
> Am I right that you expect a different set of flags of a file for each of the 2 checks you have?
> 
> ```
> 					"/tmp/.ICE-unix": {
> 						"type": "d",
> 						"mode": "1777",
> 						"owner": "root:root"
> 					}
> ...
> 					"/var/lib/named/dev/random": {
> 						"type": "c",
> 						"mode": "0666",
> 						"dev": "1,8",
> 						"owner": "root:root"
> 					},
> ```

Well for the device file we naturally need the minor and major IDs otherwise the settings can be shared.

marxin:
All right, I implemented the 2 checks. It's tested very lightly, feel free to add more tests, please.
And please make a pull request with a description of the checks.

rpmlint/checks/WorldWritableCheck.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self, config, output):
+        super().__init__(config, output)
+        self.whitelists = self.config.configuration['WorldWritableWhitelist']
+        self.verify_whitelists(['path', 'owner', 'group', 'mode'])
+        self.prefix = 'world-writable'
+
+    def check_binary(self, pkg):
+        files = [FileMeta(pfile) for pfile in pkg.files.values()]
+        # select only block and character devices, ignore symlinks
+        files = {f.path: f for f in files if not f.mode[0] in 'bcs' and f.mode[-1] == 't'}
```
mgerstner, line 14: This is not quite what we need here. This only considers sticky bit directories. What we need are any world-writable files except device files and symlinks. So something like this:

```
files = {f.path: f for f in files if not f.mode[0] in 'bcl' and f.mode[-2] == 'w'}
```

Also you should update the comment line above.

mgerstner (commented):
The device file check looks good. The world writable check requires some tuning but after that should also work out.

rpmlint/checks/WorldWritableCheck.py
marxin: Thank you for the fix.

mgerstner:
With this fix the basic check should work as expected. I can then add more tests and the description in a follow-up PR.

mgerstner (approved):

marxin:
Great, thanks.

PR #606, opened 2021-02-04
Fix and improve FakePkg

mgerstner:
While trying to implement tests based on FakePkg I found an error and some shortcomings. This addresses them.


mgerstner:
@marxin Nice idea with the `FakePkg` but it needs some polishing. Can you please review this?

marxin:
LGTM

PR #607, opened 2021-02-05
tests: add tests with better coverage of FileDigestsCheck

mgerstner:


marxin:
Please install all the `python3-flake8*` packages mentioned in `.github/workflows/main.yml`.
Then you should see the failing flake8:

```
$ flake8
./test/test_file_digest.py:2:1: I100 Import statements are in the wrong order. 'import os' should be before 'import pytest' and in a different group.
./test/test_file_digest.py:2:1: I201 Missing newline between import groups. 'import os' is identified as Stdlib and 'import pytest' is identified as Third Party.
./test/test_file_digest.py:3:1: I201 Missing newline between import groups. 'from rpmlint.checks.FileDigestCheck import FileDigestCheck' is identified as Third Party and 'import os' is identified as Stdlib.
./test/test_file_digest.py:60:36: Q000 Remove bad quotes
./test/test_file_digest.py:72:36: Q000 Remove bad quotes
./test/test_file_digest.py:87:36: Q000 Remove bad quotes
./test/test_file_digest.py:103:36: Q000 Remove bad quotes
./test/test_file_digest.py:120:36: Q000 Remove bad quotes
./test/test_file_digest.py:129:16: Q000 Remove bad quotes
./test/test_file_digest.py:130:16: Q000 Remove bad quotes
./test/test_file_digest.py:136:36: Q000 Remove bad quotes
```

and you will also see it as part of `pytest`.

mgerstner:
yeah I already saw that, should be better now

test/test_file_digest.py
```diff
@@ -46,14 +57,85 @@ def test_description_message(tmpdir, digestcheck):
...
+    output, test = get_digestcheck('digests.config')
+    with FakePkg('testpkg') as pkg:
+        pkg.add_file_with_content('/restricted/1/dangerous', 'really dangerous stuff')
+        pkg.add_symlink_to('/alsorestricted/2/suspicious', '/other/place/suspicious.txt')
+        pkg.add_file_with_content('/other/place/suspicious.txt', 'really suspicious stuff')
+        pkg.add_file_with_content('/related/and/also/sensitive', 'related fine stuff')
+        test.check(pkg)
+        assert len(output.results) == 1
+        if output.results:
+            error = output.results[0]
```
marxin, line 97: Likewise here.

test/test_file_digest.py
```diff
@@ -46,14 +57,85 @@ def test_description_message(tmpdir, digestcheck):
...
+    with FakePkg('testpkg') as pkg:
+        pkg.add_file_with_content('/restricted/1/dangerous', 'really dangerous stuff')
+        pkg.add_symlink_to('/alsorestricted/2/suspicious', '/other/place/suspicious.txt')
+        pkg.add_file_with_content('/other/place/suspicious.txt', 'really suspicious stuff')
+        pkg.add_file_with_content('/related/and/also/sensitive', 'related sensitive stuff')
+        pkg.add_file_with_content('/restricted/1/evil', 'evil stuff')
+        test.check(pkg)
+        assert len(output.results) == 1
+        if output.results:
+            error = output.results[0]
```
marxin, line 114: Likewise here.

test/test_file_digest.py
```diff
@@ -46,14 +57,85 @@ def test_description_message(tmpdir, digestcheck):
...
+    output, test = get_digestcheck('digests.config')
+    with FakePkg('testpkg') as pkg:
+        pkg.add_file_with_content('/restricted/1/dangerous', 'really dangerous stuff')
+        pkg.add_symlink_to('/alsorestricted/2/suspicious', '/other/place/suspicious.txt')
+        pkg.add_file_with_content('/other/place/suspicious.txt', 'really good stuff')
+        pkg.add_file_with_content('/related/and/also/sensitive', 'related sensitive stuff')
+        test.check(pkg)
+        assert len(output.results) == 1
+        if output.results:
+            error = output.results[0]
```
marxin, line 82: You can make this check unconditional.

marxin (changes requested):

test/test_file_digest.py
mgerstner: I actually deliberately kept the `if` condition, because if the prior assertion fails and output.results is empty then an unhelpful follow-up error (IndexError) will occur. If you think it is better without that then I can adjust, of course.

mgerstner: I thought that I saw in my test runs that the initial assertion does not abort execution of the test. But maybe I am wrong about that.

marxin: That should not happen, I tested that locally :)

mgerstner:
Okay, I remove the if statements.

PR #622, opened 2021-03-23
add initial Fedora config

xsuchy:
I used openSUSE config as a start. Taken the config form rpmlint 1.11 from Fedora and
converted it to toml. Plus removed stuff which no longer exist or is too old.

Stuff which is commented should be discussed with Fedora community and either enabled
or removed.

Originally, I wanted to start with mostly empty placeholders and then start a discussion on Fedora's mailing list. But I got quite far.... :)

Anyway - once this will be merged I start a discussion on the mailing list. You can expect future changes on this config.


configs/Fedora/scoring.toml
```diff
@@ -0,0 +1,6 @@
+[Scoring]
+# This can set how bad each error is.
+# But we set BadnessThreshold to -1 so anything above 0
+# is actually fatal.
```
marxin, line 4: This can be changed (and that's what we'll use in openSUSE):
`  -P, --permissive      treat individual errors as non-fatal`

configs/Fedora/users-groups.toml
```diff
@@ -0,0 +1,29 @@
+# generated using vvvv on 2021-03-23
+#import re
```
marxin, line 2: I would consider extracting the script to a `scripts` folder in the project?

marxin: We already have such a folder and it contains `tools/generate-isocodes.py`.

PR #644, opened 2021-05-17
Opensuse polkit checks

acipm:
Ports PolkitCheck from rpmlint-checks/CheckPolkitPrivs.py to rpmlint/checks/PolkitCheck.py

Moves digests of rules.d to be checked by FileDigestCheck.

PR #646, opened 2021-05-18
Polkit check

acipm:
As discussed in #644


rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def _collect_privs(self):
+        for filename in self.polkit_privs_files:
+            if os.path.exists(filename):
+                self._parse_privs_file(filename)
+
+    def _parse_privs_file(self, filename):
+        with open(filename) as inputfile:
+            for line in inputfile:
+                line = line.split('#')[0].split('\n')[0]
```
marxin, line 24: What about using `.rstrip()`?

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,128 @@
...
+                if len(line):
```
marxin, line 25: You can use `if line` only, which is `False` for an empty string.

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,128 @@
...
+                if f.startswith(prefix):
+                    if f in pkg.ghost_files:
+                        self.output.add_info('E', pkg, 'polkit-ghost-file', f)
+                        continue
+
+                    xml = parse(pkg.dirName() + f)
+                    for a in xml.getElementsByTagName('action'):
+                        self.check_action(pkg, a)
+            except Exception as x:
+                self.output.add_info('E', pkg, 'rpmlint-exception', f'{f:s} raised an exception: {x}')
```
marxin, line 79: It's too much generic error. I would not catch an expection. It's caught in `lint.py`.

marxin (changes requested):

rpmlint/checks/PolkitCheck.py
acipm: I think it needs to catch the error to continue if the xml is broken. I would instead use `xml.parsers.expat.ExpatError`, would that be okay for you?

marxin: Then, please rename the error to something like `polkit-xml-exception`?

marxin: Here I meant reporting `self.output.add_info('E', pkg, 'polkit-xml-exception'...`

marxin (changes requested):

rpmlint/checks/PolkitCheck.py
acipm: Ah, sorry I misunderstood. It should be fixed now.

marxin (changes requested):
LGTM, one last change: can you please add a description for the new warnings/errors?
You can make them SUSE-agnostic and please extend the description in the follow up pull request.

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,129 @@
...
+                line = line.split('#')[0].rstrip()
+                if line:
+                    line = split(r'\s+', line)
+                    priv = line[0]
+                    value = line[1]
+                    self.privs[priv] = value
+
+    def check_perm_files(self, pkg):
+        """Checks files in polkit-default-privs.d."""
+        prefix = '/etc/polkit-default-privs.d/'
```
Conan-Kudo, line 34: What is this directory and where does it come from?

Conan-Kudo (changes requested):

marxin:
@mgerstner

rpmlint/checks/PolkitCheck.py
acipm: This directory can be used by applications to write their own polkit profile to `/etc/polkit-default-privs.d/` (such as for example https://github.com/openSUSE/polkit-default-privs/blob/master/profiles/polkit-default-privs.standard). These profile paths have to be whitelisted in the rpmlint config.

The check and the path come from here: https://github.com/openSUSE/rpmlint-checks/blob/master/CheckPolkitPrivs.py#L75

This is currently only used like that by opensuse but other users could theoretically also do it like that.

Conan-Kudo: How does this even work? It doesn't seem integrated into polkit itself? Has anyone discussed it with them?

Conan-Kudo:
So I have a general comment here: I thought that this check leveraged infrastructure that existed in polkit. That doesn't appear to be the case. I also am not very sure how this whole system works for the host system. For integrating this into mainline rpmlint, I'd like some kind of understanding of how this would work first and whether this check is even a reasonable thing to implement _period_. And yes, that means I want to understand what benefit it gives for openSUSE users (especially since I know that this is an aspect that people generally hate about openSUSE).

acipm:
> So I have a general comment here: I thought that this check leveraged infrastructure that existed in polkit. That doesn't appear to be the case. I also am not very sure how this whole system works for the host system. For integrating this into mainline rpmlint, I'd like some kind of understanding of how this would work first and whether this check is even a reasonable thing to implement _period_. And yes, that means I want to understand what benefit it gives for openSUSE users (especially since I know that this is an aspect that people generally hate about openSUSE).

Thats why I was surprised, why you would want to get this into `main` instead of into `opensuse` as described in #644  :) I thought this was very suse-specific but I did not want to start a discussion. Probably @mgerstner could elaborate further, but I would also be okay to create another PR to `opensuse` and close this.

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,128 @@
...
+
+                if i.nodeName in allow_types:
+                    settings[i.nodeName] = i.firstChild.data
+        except KeyError:
+            foundunauthorized = True
+
+        for i in allow_types:
+            if i not in settings:
+                foundundef = True
+                settings[i] = '??'
```
mgerstner, line 109: This old logic is not quite correct. There are no "undefined" action settings, because there is a default applied in polkit if it is not explicitly stated. This default is not well documented, however from my notes I gather the following:

```
If a setting is missing from the XML policy file then by default `no` is
implied. This is not well documented and needs to be tracked down in the
polkit code, where the `ParserData` is zero initialized and a zero implicit
authorization means `no`.
```

This means that instead of `'??'`, `'no'` should be used here.

mgerstner:
> Thats why I was surprised, why you would want to get this into `main` instead of into `opensuse` as described in #644 :) I thought this was very suse-specific but I did not want to start a discussion. Probably @mgerstner could elaborate further, but I would also be okay to create another PR to `opensuse` and close this.

Yes this check definitely belongs into the opensuse branch. It is SUSE specific, because it only works together with the polkit-default-privs package that is also SUSE specific.

Conan-Kudo:
> > So I have a general comment here: I thought that this check leveraged infrastructure that existed in polkit. That doesn't appear to be the case. I also am not very sure how this whole system works for the host system. For integrating this into mainline rpmlint, I'd like some kind of understanding of how this would work first and whether this check is even a reasonable thing to implement _period_. And yes, that means I want to understand what benefit it gives for openSUSE users (especially since I know that this is an aspect that people generally hate about openSUSE).
> 
> Thats why I was surprised, why you would want to get this into `main` instead of into `opensuse` as described in #644 :) I thought this was very suse-specific but I did not want to start a discussion. Probably @mgerstner could elaborate further, but I would also be okay to create another PR to `opensuse` and close this.

It's based on me misunderstanding this change. Sorry for wasting your time! This is good for `opensuse` only.

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,128 @@
...
+from re import split
+from xml.dom.minidom import parse
+
+from rpmlint.checks.AbstractCheck import AbstractCheck
+
+
+class PolkitCheck(AbstractCheck):
+    def __init__(self, config, output):
+        super().__init__(config, output)
+        self.polkit_privs_whitelist = config.configuration.get('PolkitPrivsWhitelist', ())
```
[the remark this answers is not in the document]
acipm: it is needed for `check_perm_files` because it contains whitelisted paths from `/etc/polkit-default-privs.d/` . if we remove the part with `/etc/polkit-default-privs.d/`, this could be removed too.

The whitelisting variable was named `POLKIT_RULES_WHITELIST` and is already removed with all the filedigests stuff.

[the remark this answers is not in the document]
acipm: I have removed the perm file feature in the new PR #649

PR #649, opened 2021-05-18
Opensuse polkit check

acipm:
as discussed in #646, this needs the opensuse polkit-default-privs package and thus is opensuse specific


acipm:
@mgerstner

marxin (approved):

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,95 @@
+import os
+from re import split
+from xml.dom.minidom import parse
+from xml.parsers.expat import ExpatError
+
+from rpmlint.checks.AbstractCheck import AbstractCheck
+
+
+class PolkitCheck(AbstractCheck):
+    """Note: This expects the package polkit-default-privs to be installed to work."""
```
mgerstner, line 10: I still would like to see more elaborate documentation about the purpose of this check:

```
This check enforces a whitelisting for polkit policies installed by packages. openSUSE uses the package
`polkit-default-privs` to apply one of a selection of default system profiles for polkit actions. A package's polkit
action is considered whitelisted if an entry for it is existing in polkit-default-priv's standard profile.

Apart from this this check also performs minor consistency checks and gives suggestions to packages regarding 
installed polkit policies.
```

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,95 @@
...
+        """Inspect a single polkit action used by an application."""
+        action_id = action.getAttribute('id')
+        if action_id in self.privs:
+            # the action is explicitly whitelisted, nothing else to do
+            return
+
+        allow_types = ('allow_any', 'allow_inactive', 'allow_active')
+        foundunauthorized = False
+        foundno = False
+        foundundef = False
```
mgerstner, line 60: Can you please add underscores here to match the style of the rest of the code like `found_unauthorized` etc.

rpmlint/checks/PolkitCheck.py
```diff
@@ -0,0 +1,95 @@
...
+                    continue
+
+                if i.nodeName in allow_types:
+                    settings[i.nodeName] = i.firstChild.data
+        except KeyError:
+            foundunauthorized = True
+
+        for i in allow_types:
+            if i not in settings:
+                foundundef = True
```
mgerstner, line 75: I think with the change to assume `'no'` by default we can drop the `foundundef` differentiation and simply reuse the `foundno` variable to keep track of this state.

rpmlint/descriptions/PolkitCheck.toml
```diff
@@ -0,0 +1,20 @@
+polkit-unauthorized-privilege="""
+The package allows unprivileged users to carry out privileged
+operations without authentication. This could cause security
+problems if not done carefully. #REVIEW_NEEDED_TEXT#
+"""
+polkit-untracked-privilege="""
+The privilege is not listed in /etc/polkit-default-privs.*
+which makes it harder for admins to find. Furthermore polkit
+authorization checks can easily introduce security issues. #REVIEW_NEEDED_TEXT#
```
mgerstner, line 9: Maybe make this into `Furthermore *improper* polkit authorization checks ...`

mgerstner (changes requested):

rpmlint/descriptions/PolkitCheck.toml
```diff
@@ -0,0 +1,20 @@
...
+"""
+polkit-cant-acquire-privilege="""
+Usability can be improved by allowing users to acquire privileges
+via authentication. Use e.g. 'auth_admin' instead of 'no' and make
+sure to define 'allow_any'. This is an issue only if the privilege
+is not listed in /etc/polkit-default-privs.
+"""
```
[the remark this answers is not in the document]
jsegitz: I agree. There are cases where setting it to no makes sense and then this would be confusing

rpmlint/descriptions/PolkitCheck.toml
```diff
@@ -0,0 +1,16 @@
...
+The package allows unprivileged users to carry out privileged
+operations. This could cause security
+problems if not done carefully. #REVIEW_NEEDED_TEXT#
+"""
+polkit-untracked-privilege="""
+The privilege is not listed in /etc/polkit-default-privs.*
+which makes it harder for admins to find. Furthermore improper polkit
+authorization checks can easily introduce security issues. #REVIEW_NEEDED_TEXT#
+"""
+polkit-cant-acquire-privilege="""
```
mgerstner, line 11: Instead of providing an empty text here I think it is better to completely remove the `add_info` invocation above. You can instead keep the `if` branch with a comment like:

```
if found_no:
    # in the past this triggered 'polkit-cant-acquire-privilege', but this is not considered very helpful for
    # packagers since there are legitimate use cases for this.
    pass
```

We can then drop this `found_no` logic in a follow-up commit completely, just so that it is documented in the check's git history.

mgerstner (approved):
looks good to me now

PR #651, opened 2021-05-25
Remove deprecated entries from security whitelistings

acipm:


acipm:
@mgerstner

test/test_cli.py
```diff
@@ -25,7 +25,7 @@ def test_parsing_config_file(test_arguments):
 def test_parsing_opensuse_conf(test_arguments):
     parsed = process_lint_args(test_arguments)
 
-    assert len(parsed['config']) == 13
+    assert len(parsed['config']) == 12
```
mgerstner, line 28: Is it really a good idea to test for the exact number of entries of the production config? This test will surely fail every time somebody changes something in the configuration and cause a lot of noise and trouble.

acipm: @marxin

marxin: Yes, please remote it, it's pretty fragile as it seems.

marxin:
Please rebase the branch so that CI is re-run.

marxin:
@mgerstner Are you fine with the PR now?

mgerstner:
> Please don't delete the device files whitelist, even if it is empty. We want to be able to quickly find the right file should we ever need to add a whitelisting there. Instead add a comment in there stating that currently no device files are whitelisted.

The file still is deleted, can you please adjust that?

marxin:
> The file still is deleted, can you please adjust that?

The file is not deleted and there's a comment you requested.

mgerstner:
> > The file still is deleted, can you please adjust that?
> 
> The file is not deleted and there's a comment you requested.

Actually is **gets** deleted but it is re-added in the follow-up commit. Well I can live with that.

marxin:
Got it. @acipm Can you please squash the 2 commits?

PR #652, opened 2021-05-25
Move opensuse specific checks into opensuse.toml

acipm:
There is an inconsistency how checks are enabled. opensuse-specific WorldWritableCheck, DeviceFilesCheck.... are enabled in "configdefaults.toml", while other also opensuse-specific checks are enabled in "opensuse.toml". 

This moves every opensuse specific check from configdefaults.toml to opensuse.toml. If there is a reason to keep it as is or do it otherwise, I am open to it.

@mgerstner @marxin


marxin:
> There is an inconsistency how checks are enabled. opensuse-specific WorldWritableCheck, DeviceFilesCheck.... are enabled in "configdefaults.toml", while other also opensuse-specific checks are enabled in "opensuse.toml".

Well, I think all checks that are in `main` branch (and not enabled by default) should be enabled in `opensuse.toml` file as we do it now. The reason is that it increases test coverage as most pull requests are based on `main` branch.

marxin:
Oh, but you speak about checks that are present only on `opensuse` branch. That's correct!

rpmlint/configdefaults.toml
```diff
@@ -331,14 +327,6 @@ binaries = false
...
 [DanglingSymlinkExceptions]
 
-[FileDigestLocation]
-
-[FileDigestGroup]
-
-[DeviceFilesWhitelist]
-
-[WorldWritableWhitelist]
-
```
mgerstner, line 341: Do we need to move these empty config entries to `opensuse.toml`? Otherwise the checks might fail when there are no explicit whitelistings like with devicefileswhitelist.

acipm: Yes, you are right, I missed that. I will fix it

PR #655, opened 2021-05-30
Rewrite parts of FileDigestCheck

acipm:
file-digest-mismatch is not triggered when a whitelist entry exists but the digest is wrong. This rewrites parts of the FileDIgestCheck to fix the bug.

Please **don't** merge before @mgerstner has reviewed.


marxin:
Please rebase it.
@mgerstner : Can you please review it?

rpmlint/checks/FileDigestCheck.py
```diff
@@ -90,47 +90,60 @@ def _is_valid_digest(self, pkgfile, digest, pkg):
...
+            for digest in digests_of_path:
+                filename = digest['path']
+                # Check if digest whitelist path has a matching file in our package
+                if not pkg.files.get(filename):
+                    # This digest entry is not needed anymore and could be dropped
+                    continue
+                valid_digest, file_digest = self._is_valid_digest(pkg.files[filename], digest, pkg)
+                if valid_digest:
+                    # Valid digest found, no mismatch error will be printed
+                    error = False
```
mgerstner, line 139: to make the code a bit more compact you could consider dropping the `error` variable and simply reset `error_digests = []`. Or do you think it hurts readability of the code?

mgerstner:
Apart from the one idea I commented about the change looks good to me. Also nice that you commented the code to make it more easily understandable, since this logic is non-obvious for people outside of our review team.

mgerstner:
@acipm: And one more thing: Can you please adjust the unit tests so that this aspect is actually covered in the future?

acipm:
> @acipm: And one more thing: Can you please adjust the unit tests so that this aspect is actually covered in the future?

done

mgerstner (approved):
LGTM now.

marxin (changes requested):
Can you please submit the newly created test-case to `https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests`?

PR #656, opened 2021-05-31
Re-add permissions.d whitelisting

acipm:
There is an issue when the "whitelisting" is not in the main permissions profiles, but only in the per-package permission profiles, as they are not considered at all.

This re-adds the first pass of the original check without the original whitelist. Instead we use the FileDIgestLocation "permission" in the opensuse config to ensure files in the permissions.d path to be whitelisted.

Please **don't** merge until @mgerstner reviewed this.


marxin:
Please rebase it.
@mgerstner : Can you please review it?

mgerstner:
Change looks good to me, but can also here please add test coverage so that this issue will be detected in the future?

acipm:
> Change looks good to me, but can also here please add test coverage so that this issue will be detected in the future?

done

mgerstner:
Looks good to me now. @marxin can merge it if he it also fine with the change.

marxin (changes requested):
Again, can you please submit a test-case to the OBS project?

PR #661, opened 2021-06-11
Implement filter exceptions.

### LINKED ISSUE #660: equivalent of setFilterException
SUSE's rpmlint 1.x had a downstream patch to implement `setFilterException` 

this was used in the config packages to make certain checks non-filterable by the package included rpmlintrcs:

`./rpmlint-Factory/config:setFilterException('suse-dbus-unauthorized-service')`

how is that now working within rpmlint 2.x?

marxin:
The feature is handy for hardening checks that should not be filterable
by a rpmlintrc file.

Fixes #660.

### DIFF
--- a/rpmlint/configdefaults.toml
+++ b/rpmlint/configdefaults.toml
@@ -29,6 +29,8 @@ Checks = [
 ]
 # Various output filters, list of regexp strings eg. "E: .* no-signature"
 Filters = []
+# List of errors that can't be filtered
+BlockedFilters = []
 # Treshold where we should error out, by default single error is enough
 BadnessThreshold = -1
 # When checking that various files that should be compressed are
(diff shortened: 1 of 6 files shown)


rpmlint/configdefaults.toml
```diff
@@ -29,6 +29,8 @@ Checks = [
 ]
 # Various output filters, list of regexp strings eg. "E: .* no-signature"
 Filters = []
+# List of errors that can't be filtered
+FilterExceptions = []
```
Conan-Kudo, line 33: I don't think this is an intuitive name for this. The idea here is that check rules listed here must pass, right?

Conan-Kudo (changes requested):

rpmlint/configdefaults.toml
marxin: Hm, you are right. What about `SeriousErrors`, `SecurityErrors` or `FatalErrors`. Or do you have a more suitable name?

Conan-Kudo: `FatalErrors` sounds better. This could also turn a warning into an error, right?

dirkmueller: it is not supposed to change the warning level at all. it is about suppressing filtering. 

we could also call it "FilterDenylist" or the like

Conan-Kudo: `BlockedFilters` maybe?

Conan-Kudo (approved):

PR #664, opened 2021-06-16
Fix failure when checking rust library files

mlehtima:
These are seen as archives in run_elf_checks which returns before
initializing readelf_parser.

Fixes error when checking rpms which contain rlib files:
(none): E: fatal error while reading some_file.rpm: 'BinariesCheck' object has no attribute 'readelf_parser'

### DIFF
--- a/rpmlint/checks/BinariesCheck.py
+++ b/rpmlint/checks/BinariesCheck.py
@@ -626,10 +626,13 @@ class BinariesCheck(AbstractCheck):
             self.run_elf_checks(pkg, pkgfile.path, fname)
 
             # inspect binary file
-            is_shlib = self.readelf_parser.is_shlib
+            try:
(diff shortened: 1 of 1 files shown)


marxin:
Can you please upload somewhere such a RPM file?
Or is it possible to create a simple reproducer?

mlehtima:
Hi, sorry for the delayed answer. Here is a link to a rpm file which can be used to reproduce the issue https://releases.jolla.com/releases/4.1.0.24/jolla/armv7hl/oss/armv7hl/rust-std-static-armv7-unknown-linux-gnueabihf-1.0+git3-1.2.1.jolla.armv7hl.rpm

marxin:
Thanks for the test-case. I've just created a pull request that fixes that: #665.

PR #665, opened 2021-06-23
Fix non-standard archives (rust library).

marxin:


mlehtima:
I tested this fix with the test rpm linked in the other pull request and this also fixes the issue.

marxin:
Thanks for testing of the patch.

PR #670, opened 2021-06-30
OpenSUSE: Add compact skip format for filedigestcheck

acipm:
Adds a `nodigest` option as a compact way to add many secured paths without digests.

Changes the `dbus-services.toml` config to be covered by FileDigestCheck and update config. Please note that I could not find all entries from the old config. Thus those I did not found are commented out.
Removes DBusSevicesCheck.

Please don't merge until @mgerstner has reviewed.

@mgerstner This does **not** include the globbing for the pam modules, I will open a seperate PR next week for that


rpmlint/checks/FileDigestCheck.py
```diff
@@ -20,6 +20,17 @@ def __init__(self, config, output):
             if dg_type not in self.digest_configurations:
                 raise KeyError(f'FileDigestGroup type "{dg_type}" is not '
                                f'supported, known values: {list(self.digest_configurations)}')
+            # expand skip digests
+            for skip_digest in digest_group.get('skips', []):
```
mgerstner: Didn't we want to use a more intuitive name than `skips` here? Like `nodigests` or something alike.

mgerstner (changes requested):

rpmlint/checks/FileDigestCheck.py
mgerstner: Ah only now I see that you renamed it it a follow-up commit. Can you sqash these two commits so we get a cleaner history?

configs/openSUSE/dbus-services.toml
```diff
@@ -1,270 +1,750 @@
...
+    "/usr/lib/systemd/system/dbus-org.freedesktop.timedate1.service",
+    "/usr/share/dbus-1/system-services/org.freedesktop.timedate1.service",
+    "/usr/share/dbus-1/system.d/org.freedesktop.locale1.conf",
+    "/usr/lib/systemd/system/dbus-org.freedesktop.locale1.service",
+    "/usr/share/dbus-1/system-services/org.freedesktop.locale1.service",
+]
+
+# gconf2
+#"org.gnome.GConf.Defaults.service",
+#"org.gnome.GConf.Defaults.conf",
```
mgerstner, line 87: This is still existing in `gconf-polkit` from `gconf2`.

configs/openSUSE/dbus-services.toml
```diff
@@ -1,270 +1,750 @@
...
+
+# system-config-printer (bnc#694640)
+#"com.redhat.NewPrinterNotification.conf",
+#"com.redhat.PrinterDriversInstaller.conf",
```
mgerstner, line 91: this is also still existing in `system-config-printer-applet` from `system-config-printer`

configs/openSUSE/dbus-services.toml
```diff
@@ -1,270 +1,750 @@
...
+skips = [
+    "/usr/share/dbus-1/system-services/org.kde.kcontrol.kcmclock.service",
+    "/usr/share/dbus-1/system.d/org.kde.kcontrol.kcmclock.conf",
+]
+
+# kdebase4-workspace
+#"org.kde.fontinst.service",
+#"org.kde.fontinst.conf",
+#"org.kde.ksysguard.processlisthelper.service",
+#"org.kde.ksysguard.processlisthelper.conf",
```
mgerstner, line 135: kde4 is gone from Factory so this should be fine

mgerstner: Actually, no, these still exist in `plasma5-workspace` etc. Please check all commented out candidates using the `query_archives_db` tool. For erxample like this: `query_archives_db -p "%org.kde.ksysguard%" -f rpm,path`.

mgerstner (changes requested):

mgerstner:
Looks good to me now. @marxin, you may merge now.

PR #673, opened 2021-07-06
OpenSUSE filedigest globbing

acipm:
This needs to be merged **after** #670, as it is branched from there and includes its commits.

These are the new commits:

b143327ed9fa12fd695b5ffff09b68e3ae8b412d adds globbing for the `path` in `FileDigestLocation` so that `/lib64` and `/lib` can be caught using one entry `/lib*` for the pam modules. Additionally this adds an `IncludeOnly` keyword that can be used to "filter" `.so` files without having to glob the `Locations` in `FileDigestLocation` (because imo, globbing `Locations` could lead to some weird side-effects). Details and example see commit message.

494274576f98a100e43e6bea98d679ab710daed9 migrates `pam-modules.toml` to use FileDigestCheck with globbing.

a75b93b860bb2f1553274ef1f73c2947081b9995 adds tests.

Please **don't** merge before @mgerstner has reviewed.


rpmlint/checks/FileDigestCheck.py
```diff
@@ -49,7 +52,12 @@ def _get_digest_configuration_group(self, pkgfile):
             for location in locations:
                 try:
                     if path.relative_to(location):
-                        return group
+                        if not self.include_only_in_group[group]:
+                            return group
+                        else:
+                            for glob in self.include_only_in_group[group]:
+                                if fnmatch(path, glob):
```
mgerstner, line 59: should we maybe only match against the basename of the `path`? Otherwise false positives could occur. Say we have an `IncludeOnly="*lib*.so"` and encounter a path like `/usr/lib/security/plugin.so` then it would still match against the complete path, while the basename would not match.

configs/openSUSE/pam-modules.toml
```diff
@@ -1,143 +1,458 @@
...
+    "/usr/lib*/security/pam_ecryptfs.so",
+]
+
+[[FileDigestGroup]]
+package = "gnome-keyring-pam"
+type = "pam"
+note = "legacy: not audited"
+bug = ""
+nodigests = [
+    "/lib*/security/pam_gnome_keyring.so",
```
mgerstner, line 26: We currently seem to be in a migration from /lib/security to /usr/lib/security. To make this migration smooth it would make sense to whitelist both locations like:

`*/lib*/security/pam_gnome_keyring.so`

Since the check only triggers for paths are located either in /usr/lib/security or /lib/security we could even just whitelist "*/pam_gnome_keyring.so", am I right?

Due to the coupling of packages to whitelisting entries there is also no danger that other packages hijack a PAM module name.

acipm: I renamed them now in 3f7d5cafb2254ea6e890c1ce92ad26eda9e09ccd

mgerstner:
Okay looks fine from my side now. Ready for @marxin to review.

PR #674, opened 2021-07-08
look for man pages in regular files, not just symlinks

spotaws:
Currently, the code only looks for man pages in the stat.S_ISLNK(mode) conditional. This adds the same man page search to the regular file case. This is useful because most man pages are not symlinks and without this change, rpmlint reports missing man pages when they are present (unless they happen to be symlinks).

Reported here: https://bugzilla.redhat.com/show_bug.cgi?id=1980400


marxin:
Thank you for the pull request.

PR #678, opened 2021-07-22
Come up with --mini-mode option.

marxin:
The option will be used by rpmlint-lint package wrapper.

### DIFF
--- a/rpmlint/checks/SpecCheck.py
+++ b/rpmlint/checks/SpecCheck.py
@@ -107,6 +107,7 @@ class SpecCheck(AbstractCheck):
         self._spec_file = None
         self._spec_name = None
         self.valid_groups = config.configuration['ValidGroups']
+        self.mini_mode = config.mini_mode
         self.output.error_details.update({'non-standard-group':
                                          """The value of the Group tag in the package is not valid.  Valid groups are:
                                          '%s'.""" % ', '.join(self.valid_groups)})
@@ -518,8 +519,10 @@ class SpecCheck(AbstractCheck):
         # Checks below require a real spec file
         if not self._spec_file:
             return
-        self._check_specfile_error(pkg)
-        self._check_invalid_url(pkg, rpm)
+
+        if not self.mini_mode:
+            self._check_specfile_error(pkg)
+            self._check_invalid_url(pkg, rpm)
 
     def _check_no_spec_file(self, pkg):
         """Check if no spec file is found in RPM meta data."""
--- a/rpmlint/checks/TagsCheck.py
+++ b/rpmlint/checks/TagsCheck.py
@@ -47,7 +47,7 @@ class TagsCheck(AbstractCheck):
         self.valid_buildhost_regex = re.compile(config.configuration['ValidBuildHost'])
         self.use_epoch = config.configuration['UseEpoch']
         self.max_line_len = config.configuration['MaxLineLength']
-        self.spellcheck = config.configuration['UseEnchant']
+        self.spellcheck = config.configuration['UseEnchant'] and not self.config.mini_mode
         self.valid_license_exceptions = config.configuration['ValidLicenseExceptions']
         if self.spellcheck:
             self.spellchecker = Spellcheck()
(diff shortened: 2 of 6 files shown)


Conan-Kudo:
What is this for?

marxin:
In order to keep a limited number of bootstrap cycle packages, we don't want to depend on `pyenchant` and `python39`.
The pull request goes only to `opensuse` branch..

marxin:
Instead of `python39`, we create a `virtualenv`.

Conan-Kudo:
That doesn't solve the problem for you. You'd still need the Python interpreter because virtualenvs do not include the interpreter. Moreover, I don't know why you need an extra wrapper or mode when you could just make those those things weak dependencies instead. The `-mini` convention isn't really for minimized builds, it's for producing hacksaw versions for the OBS build environment. (Personally, I think any package that is called `-mini` in openSUSE needs a better name, because it's clearly wrong.).

It also doesn't actually make sense to have rpmlint as a bootstrap package. It's not _required_ for building packages, and trying to force it in that early is just asking for pain.

marxin:
> That doesn't solve the problem for you. You'd still need the Python interpreter because virtualenvs do not include the interpreter. 

We do have it, but does not work due to: Because it runs rpm, which expands every macro in the spec file, which includes things like %pythons and so on, and so it ends up running python3.9 which then complains bitterly since most of it is removed.

> Moreover, I don't know why you need an extra wrapper or mode when you could just make those those things weak dependencies instead.

Right now, the spellchecker is a weak dependency but generates `(none): W: unable to init enchant, spellchecking disabled.`.

> The `-mini` convention isn't really for minimized builds, it's for producing hacksaw versions for the OBS build environment. 

Yes, we are aware of the hackish approach.

(Personally, I think any package that is called `-mini` in openSUSE needs a better name, because it's clearly wrong.).
> 
> It also doesn't actually make sense to have rpmlint as a bootstrap package. It's not _required_ for building packages, and trying to force it in that early is just asking for pain.

Well, that's decision made by `openSUSE:Factory` people and they want to have rpmlint 2.0 as a mandatory requirement for now.

Conan-Kudo:
> Well, that's decision made by openSUSE:Factory people and they want to have rpmlint 2.0 as a mandatory requirement for now.

As an rpmlint upstream maintainer _and_ an openSUSE person, I would strongly advise _against_ it. When OBS does bootstrap cycles, it goes through and builds them and then rebuilds them again in the main project itself. Because of that, there's no particular reason to make the bootstrap phase include rpmlint. Since hacksaw stuff is already happening there anyway, it's not particularly useful.

marxin:
> As an rpmlint upstream maintainer _and_ an openSUSE person, I would strongly advise _against_ it. When OBS does bootstrap cycles, it goes through and builds them and then rebuilds them again in the main project itself. Because of that, there's no particular reason to make the bootstrap phase include rpmlint. Since hacksaw stuff is already happening there anyway, it's not particularly useful.

I fully got that and support it!
I've just added you to a Jira ticket where the problem was discussed. Feel free to express the need to make `rpmlint` only an optional dependency.

Conan-Kudo:
> > As an rpmlint upstream maintainer _and_ an openSUSE person, I would strongly advise _against_ it. When OBS does bootstrap cycles, it goes through and builds them and then rebuilds them again in the main project itself. Because of that, there's no particular reason to make the bootstrap phase include rpmlint. Since hacksaw stuff is already happening there anyway, it's not particularly useful.
> 
> I fully got that and support it!
> I've just added you to a Jira ticket where the problem was discussed. Feel free to express the need to make `rpmlint` only an optional dependency.

I don't think I can comment in that ticket, as I'm not a SUSE employee. I didn't get a Jira notification about it. 😅

marxin:
> I don't think I can comment in that ticket, as I'm not a SUSE employee. I didn't get a Jira notification about it.

Hm, I found your account and added you to https://jira.suse.com/browse/OBS-88.
Your username seems to be `Pharaoh_Atem`, am I right?

Conan-Kudo:
It is, but I only have access to the `OPENSUSE-` namespace.

marxin:
Ah, that's it :/ So are fine for now with this pull request, that's purely `opensuse`-specific?

Conan-Kudo:
Yeah, sure. I think we need to be careful with how much we add there, because every commit that's openSUSE-specific is tech debt that we _know_ is unsuitable for general use. I originally resisted the creation of the openSUSE vendor branch because I was worried about stuff like this happening, and I wanted it to be more painful to do stuff like this. But here we are...

marxin:
Thanks. Anyway, I would appreciate it if you would discuss the mandatory use of rpmlint in Factory with Dimstar.

PR #682, opened 2021-08-10
Enable Fedora rawhide + Use python3 -m pytest on the CI to have $PWD in sys.path

hroncok:


hroncok:
There are some actual failures, but the tests at least work here.

No idea why `pytest` worked before. It shouldn't have.

marxin:
That seems good, the rest fails due to:
https://bugzilla.redhat.com/show_bug.cgi?id=1990817

Please unmark it as Draft, I'm going to merge it. Thanks for help.

PR #685, opened 2021-08-15
Shorten Action workflow.

marxin:
One can easily distinguish builds in:
https://github.com/rpm-software-management/rpmlint/actions


Conan-Kudo (approved):
🥳

PR #687, opened 2021-08-15
Add launch.json.

marxin:
By default debug all .rpm files we have in test/binary folder.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug",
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
Conan-Kudo, line 4: This is invalid.

Conan-Kudo (changes requested):

.vscode/launch.json
marxin: This is a default template. What's wrong with it?

Conan-Kudo: Comments aren't allowed in JSON.

marxin: I see, fixed in the main branch.

PR #688, opened 2021-08-15
Fix detection of RPM database.

marxin:


marxin:
@carlsmedstad

marxin:
Good, now any tests are skipped (we had 4 tests before).

PR #690, opened 2021-08-19
Skip test requiring rpmdb

carlsmedstad:
Hey again! :wave:

A test was added without the proper predicate in the 2.1.0 release which causes it to fail on Arch Linux.

The added CI job should make sure this does not happen anymore in the future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ main, opensuse ]
+    branches: [main, opensuse]
   pull_request:
-    branches: [ main, opensuse ]
+    branches: [main, opensuse]
 
 jobs:
   CI:
@@ -13,20 +13,20 @@ jobs:
     strategy:
       matrix:
         build-type: ['normal']
-        docker:
+        container:
           - 'registry.fedoraproject.org/fedora:latest'
           - 'registry.fedoraproject.org/fedora:rawhide'
           - 'registry.opensuse.org/opensuse/leap-dnf:15.3'
           - 'registry.opensuse.org/opensuse/tumbleweed-dnf:latest'
         include:
-          - docker: 'registry.fedoraproject.org/fedora:latest'
+          - container: 'registry.fedoraproject.org/fedora:latest'
             build-type: 'no-optional-deps'
-          - docker: 'registry.opensuse.org/opensuse/tumbleweed-dnf:latest'
+          - container: 'registry.opensuse.org/opensuse/tumbleweed-dnf:latest'
             build-type: 'no-optional-deps'
       fail-fast: false
 
     container:
-      image: ${{ matrix.docker }}
+      image: ${{ matrix.container }}
       options: --security-opt seccomp=unconfined
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -7,7 +7,45 @@ on:
...
+              python3-pytest-xdist
+              python3-flake8
+              python3-flake8-import-order
+              python3-pyxdg
+              python3-toml
+              python3-zstd
+              python3-pip
+              rpm-build
+              git
+      - run: rm -rf $(rpm --eval %_dbpath)
```
Conan-Kudo, line 44: Please quote `%_dbpath` here so that the shell doesn't evaluate it.

Conan-Kudo (changes requested):

marxin (changes requested):
As mentioned, I prefer using matrix with `container` and `build-type` properties, where the later can be one of `normal`, `optional-deps`.

.github/workflows/main.yml
carlsmedstad: Fixed!

carlsmedstad:
> As mentioned, I prefer using matrix with `container` and `build-type` properties, where the later can be one of `normal`, `optional-deps`.

Something like in https://github.com/rpm-software-management/rpmlint/pull/690/commits/3cfe85ba58f2cd077c2a7111149b2f2df9326845?

.github/workflows/main.yml
```diff
@@ -17,6 +17,9 @@ jobs:
           - 'registry.fedoraproject.org/fedora:rawhide'
           - 'registry.opensuse.org/opensuse/leap-dnf:15.3'
           - 'registry.opensuse.org/opensuse/tumbleweed-dnf:latest'
+        build-type:
+          - 'normal'
+          - 'no-optional-deps'
```
marxin, line 22: Nice! I would not run all 2x4 combinations, but I would rather use `include` and run only 2 `no-optional-deps` (one fedora and one opensuse).

.github/workflows/main.yml
```diff
@@ -52,26 +52,26 @@ jobs:
               rpm-build
               git
         if: ${{ contains(matrix.docker, 'opensuse') }}
+      - run: dnf --assumeyes install
+              checkbashisms dash
+              desktop-file-utils
+              appstream-glib
+              myspell-en_US myspell-cs_CZ
+              python3-pyenchant
+        if: ${{ contains(matrix.docker, 'opensuse') && contains(matrix.build-type, 'normal') }}
```
marxin, line 61: Use rather `matrix.build-type == 'normal'`, please.

.github/workflows/main.yml
```diff
@@ -88,6 +88,17 @@ jobs:
...
         if: ${{ contains(matrix.docker, 'fedora') }}
+      - run: dnf --nogpgcheck --assumeyes install
+              /usr/bin/appstream-util
+              /usr/bin/desktop-file-validate
+              dash
+              devscripts-checkbashisms
+              hunspell-en
+              hunspell-cs
+              python3-enchant
+        if: ${{ contains(matrix.docker, 'fedora') && contains(matrix.build-type, 'normal') }}
```
marxin, line 99: Likewise here.

marxin (changes requested):
Very nice improvement, please fix the remaining comments I have.

carlsmedstad:
I believe I've addressed all the comments. Also, I've squashed any redundant commits.

OpenSUSE's docker registry seems to be down thought, so those tests are failing.

marxin:
> I believe I've addressed all the comments. Also, I've squashed any redundant commits.

Great, thank you for the contribution.

PR #705, opened 2021-09-17
Keep having self.options['rpmlintrc'] an array.

marxin:
The array is used in opensuse branch where auto-loading mechanism
can load multiple files.

Fixes #704.


marxin:
We can do even better by checking that the given `rpmlintrc` exists when arguments are parsed.

PR #708, opened 2021-09-17
dbus-services: whitelisting for low-memory-monitor (bsc#1189899)

mgerstner:


jsegitz (approved):
Why is this included as nodigest? I will approve this for now and we can discuss it once you're back

mgerstner:
> Why is this included as nodigest? I will approve this for now and we can discuss it once you're back

Simply because this feature wasn't available before and we don't check digests for D-Bus service files yet at all. To make this practical at least a filter of XML comments and whitespace needs to be implemented IMO.

PR #709, opened 2021-09-21
Correct fallout from stricter checking

jsegitz:
These locations were already whitelisted in rpmlint1, but now that the package name is tied to the whitelist we need additional entries


jsegitz:
overlap with #711, closing this one

PR #710, opened 2021-09-21
Ghost file exceptions

jsegitz:
We need a way to whitelist %ghost'ed files that are present in whitelist targets. We currently see this in https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:L/polkit-default-privs/standard/x86_64


jsegitz:
As this is blocking a current version of rpmlint getting into Factory I'll accept this for now. If this is suboptimal we can still change it easily as there's only one package relying on this for now

PR #712, opened 2021-09-21
Remove deprecated entries

acipm:
Please don't merge until @jsegitz or @mgerstner have reviewed it.


jsegitz:
The changes themselves are okay, but the CI checker errors need to be fixed first

acipm:
I removed some more entries in the filesystem package.

I think the Ci was just flaky, because suddenly it runs through fine.

PR #713, opened 2021-09-23
whitelist pam_systemd_home.so for systemd-experimental

jsegitz:
audited in 1185285

### DIFF
--- a/configs/openSUSE/pam-modules.toml
+++ b/configs/openSUSE/pam-modules.toml
@@ -211,9 +211,9 @@ nodigests = [
 ]
 
 [[FileDigestGroup]]
-package = "systemd"
+package = "systemd-experimental"
 type = "pam"
-note = "Module in systemd-experimental."
+note = "Imported as part of the rpmlint2 migration"
 bug = "bsc#1185285"
 nodigests = [
     "*/security/pam_systemd_home.so",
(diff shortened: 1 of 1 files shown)


configs/openSUSE/pam-modules.toml
```diff
@@ -210,6 +210,14 @@ nodigests = [
     "*/security/pam_systemd.so",
 ]
 
+[[FileDigestGroup]]
+package = "systemd"
+type = "pam"
+note = "Module in systemd-experimental. This was not audited, if it's moved to a different binary rpm it needs to be audited first"
```
wfrisch, line 216: It appears to have been audited in the meantime:
https://bugzilla.opensuse.org/show_bug.cgi?id=1185285

thesp0nge:
Hi @jsegitz, this PAM module was audited some months ago: https://bugzilla.opensuse.org/show_bug.cgi?id=1185285

configs/openSUSE/pam-modules.toml
wfrisch: We can change it to
`note = "Module in systemd-experimental."`

thesp0nge: I changed the "note" field and added the bugzilla ticket number!

wfrisch (approved):

PR #716, opened 2021-09-30
skip %ghost files in SUIDPermissionsCheck. 

jsegitz:
This allows packages to %ghost files that are managed by e.g. tmpfiles.d and rpm -v to work properly

### DIFF
--- a/rpmlint/checks/SUIDPermissionsCheck.py
+++ b/rpmlint/checks/SUIDPermissionsCheck.py
@@ -146,6 +146,15 @@ class SUIDPermissionsCheck(AbstractCheck):
         need_set_permissions = False
 
         for f, pkgfile in pkg.files.items():
+            if pkgfile.is_ghost:
+                # if a file is ghost we want to skip the files here. It's not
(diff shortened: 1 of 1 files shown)


jsegitz:
One example is systemd:
%ghost %dir %attr(2755, root, systemd-journal) %{_localstatedir}/log/journal/
This is now handled by tmpfiles.d. With 2755 in there rpmlint will complain, without it rpm -v will complain. The alternative is to whitelist these entries, but we don't want to duplicate the entries

marxin:
Please squash the 2 commits. Apart from that, I'm fine with that.

mgerstner:
> One example is systemd: %ghost %dir %attr(2755, root, systemd-journal) %{_localstatedir}/log/journal/ This is now handled by tmpfiles.d. With 2755 in there rpmlint will complain, without it rpm -v will complain. The alternative is to whitelist these entries, but we don't want to duplicate the entries

Since we have a task pending to restrict tmpfiles.d configuration files we could maybe only exclude %ghost files that are in tmpfiles.d and keep the rest? How many would that be?

PR #730, opened 2021-10-11
Enable shlib-policy-name-error error.

marxin:
Set badness to 10000 in strict configuration.

Fixes https://bugzilla.opensuse.org/show_bug.cgi?id=1191323.


darix:
Yeah having the badness back would be good. We just had a case where a library was missing and broke things.

PR #735, opened 2021-10-14
Trigger shlib-policy-name-error only for lib* packages.

marxin:


marxin:
Does not make sense to trigger the error for e.g.
```
amarok.x86_64: E: shlib-policy-name-error (Badness: 10000) SONAME: libamarok-sqlcollection.so.1, expected package suffix: 1
amarok.x86_64: E: shlib-policy-name-error (Badness: 10000) SONAME: libamarok-transcoding.so.1, expected package suffix: 1
```

marxin:
> That should probably trigger an error saying the library needs to be subpackaged out, no?

Dunno, KDE packages tend to use many shared libraries.
@Vogtinator What do you think about it?

PR #752, opened 2021-11-05
Fixed bug where a valid symlink is reported as invalid.

archiebold811:
Current bug reports `invalid-ldconfig-symlink` for a library that has proper symlinks.

This appears to be a regression from rpmlint 1.  Comparing the two reveals that the comparison of `link` to `path.parent` doesn't make sense, and comparing `link` to `path.name` would resemble rpmlint 1's behavior.


marxin:
Thank you for the pull request.
Can you please submit a spec file for the added .rpm file to https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests ?

archiebold811:
> Thank you for the pull request. Can you please submit a spec file for the added .rpm file to https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests ?

I'm trying to figure out how (I've never added one before), maybe I don't have permission?

marxin:
> I'm trying to figure out how (I've never added one before), maybe I don't have permission?

Ok, then you don't have to do it.

marxin (approved):

PR #753, opened 2021-11-05
Fix backward compatibility with rpmlintrc files.

archiebold811:
While upgrading from rpmlint v1 to v2, we found that some of our `addFilter()`s were being ignored due to spaces between the parenthesis and the filter text.  I've updated the regex that matched these filters to allow spaces as well and added tests.

### DIFF
--- a/rpmlint/config.py
+++ b/rpmlint/config.py
@@ -21,7 +21,7 @@ class Config(object):
     existing one.
     """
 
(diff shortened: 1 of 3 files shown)


marxin (approved):

marxin:
Thank you for the pull request!

marxin:
Btw. may I ask you for what distro do you update rpmlint to 2.x version?

archiebold811:
Fedora 35, found it when migrating packages from our Fedora 34 repository.

PR #760, opened 2021-11-17
duplicate postfix entry for postfix-bdb

jsegitz:
whitelistings are tied to pacakges in rpmlint2 and need to be there for every package


wfrisch (approved):
OK. Hashes identical to `package = "postfix"`.

PR #766, opened 2021-11-29
Sort input .rpm files so that the output is stable.

marxin:
Fixes: https://bugzilla.opensuse.org/show_bug.cgi?id=1193189

### DIFF
--- a/rpmlint/lint.py
+++ b/rpmlint/lint.py
@@ -224,6 +224,9 @@ class Lint(object):
         # check all elements if they are a folder or a file with proper suffix
         # and expand everything
         packages = self._expand_filelist(files)
+
+        # Sort the files so that the output is stable
+        packages = sorted(packages)
         for pkg in packages:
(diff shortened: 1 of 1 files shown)


rpmlint/lint.py
```diff
@@ -224,6 +224,9 @@ def validate_files(self, files):
         # check all elements if they are a folder or a file with proper suffix
         # and expand everything
         packages = self._expand_filelist(files)
+
+        # Sort the files so that the output is stable
+        packages = sorted(packages)
```
Conan-Kudo, line 229: We probably want this forced to C locale for sorting right?

Conan-Kudo (changes requested):

rpmlint/lint.py
marxin: Well, we need any stable sort and we have other places where we use just `sorted` without any specific locale.

PR #768, opened 2021-12-07
configs/Fedora: Sync Fedora rpmlint policies back upstream

Conan-Kudo:


marxin:
Have you used `tools/generate-fedora-users-groups.py` script for generating the users and groups?
I would recommend splitting lines in `configs/Fedora/users-groups.toml` so that diffs are more readable.

Conan-Kudo:
I had not. I just synced it from [the Fedora Dist-Git repo](https://src.fedoraproject.org/rpms/rpmlint), which @spotrh principally maintains. He might be interested in that tool, though.

Conan-Kudo:
The file is here: https://src.fedoraproject.org/rpms/rpmlint/blob/rawhide/f/users-groups.toml

Conan-Kudo:
But yes, we can split the lines in the config. I just needed this synced to make the rpmlint 2.2.0 release.

marxin:
Sure.

So I would split the lines in the config (and update the header comment) and please think about the usage of the scripts in `tools` folder. Or should we remove it?

Conan-Kudo:
does the script generate these in the correct format (split in newlines)? If so, we can use it going forward, for sure.

marxin:
> does the script generate these in the correct format (split in newlines)?

Dunno, but changing the code would be very easy ;)

marxin:
Note I can't run it as `/usr/share/doc/setup/uidgid` does not exist on openSUSE system.

PR #773, opened 2021-12-14
Make missing-call-to-setgroups-before-setuid only warning.

### LINKED ISSUE #772: rpmlint errors about setuid/setgid calls in binaries on packages without setuid set on them.
When checking several QT apps I get the following error:
```
qt5-qtapplicationmanager.x86_64: E: missing-call-to-setgroups-before-setuid /usr/lib64/qt5/bin/appman
qt5-qtapplicationmanager.x86_64: E: missing-call-to-setgroups-before-setuid /usr/lib64/qt5/bin/appman-qmltestrunner

```
None of these applications are setuid and I did not find an explicite call in the code for it (however because the code is C++ with a lot of templating my grep is probably missing something obvious :laughing:

marxin:
If the application is setuid, emit warning, otherwise report
an error.

Fixes #772.


Conan-Kudo (approved):

marxin:
@smooge: Can you please test the suggested change?

PR #778, opened 2022-01-03
Use different Python package for ZStandard

carlsmedstad:
The Python package [zstd](https://github.com/sergey-dryabzhinsky/python-zstd) does not have an `open()` function which means the current implementation is broken. I suspect the intention was to use the package [zstandard](https://github.com/indygreg/python-zstandard) all along.

To test this, try building an RPM of the following SPEC and running rpmlint on it:

```spec
Name:       zstd-file
Version:    1
Release:    1
Summary:    RPM containing ZStandard compressed file
License:    FIXME
BuildRequires: zstd

%description
RPM containing ZStandard compressed file.

%build
cat > hello-world <<EOF
HELLO WORLD
EOF
zstd -f hello-world

%install
install -Dm 644 hello-world.zst %{buildroot}/usr/share/doc/zstd-file/hello-world.zst

%files
%doc %attr(0644, root, root) "/usr/share/doc/zstd-file/hello-world.zst"
```

This will produce an exception:

```sh
$ ./lint.py zstd-file-1-1.x86_64.rpm
(none): E: fatal error while reading zstd-file-1-1.x86_64.rpm: module 'zstd' has no attribute 'open'
```

With the dependency replaced this does not happen.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
               python3-flake8-import-order
               python3-flake8-quotes
               python3-pyxdg
-              python3-zstd
+              python3-zstandard
               python3-toml
               python3-pip
               rpm-build
@@ -86,7 +86,7 @@ jobs:
               python3-flake8-import-order
               python3-pyxdg
               python3-toml
-              python3-zstd
+              python3-zstandard
               python3-pip
               rpm-build
               git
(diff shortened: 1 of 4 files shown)


marxin:
Oh, you are correct, thanks!

marxin:
Just a note about the maintenance of the packages. Last week, zstd `1.5.4` [^1] was released, `python-zstd` [^2] was updated a day after that, but we're still waiting for a `zstandard` [^3] release.

[^1]: https://github.com/facebook/zstd/releases/tag/v1.5.4
[^2]: https://github.com/sergey-dryabzhinsky/python-zstd/releases/tag/v1.5.4.0
[^3]: https://github.com/indygreg/python-zstandard/issues/190

PR #784, opened 2022-01-06
Add icingaweb2 group

lrupp:
Add icingaweb2 to the list of allowed groups. This group is used by the package icingaweb2, which already made it into Tumbleweed beside the missing group:
 https://build.opensuse.org/package/live_build_log/openSUSE:Factory/icingaweb2/standard/x86_64


marxin:
Thanks for the pull request.

PR #796, opened 2022-01-27
dbus-services: whitelist NetworkManager nm-priv-helper (bsc#1194799)

mgerstner:


marxin:
I've just created SR to Factory: https://build.opensuse.org/request/show/949508.

PR #797, opened 2022-02-02
pam-modules: remove now outdated entry for modules that moved into pa…

mgerstner:
…m_unix


jsegitz (approved):

marxin:
I've just made a SR: https://build.opensuse.org/request/show/950753.

PR #800, opened 2022-02-08
FileDigestCheck: implement support for file digest filtering

mgerstner:
We want to become more robust against trivial changes in whitelisted
files. Currently we are not using file digests for things like D-Bus
service XML files, because a lot of non-functional changes might be
disrupting our processes.

With these filter we will be able to avoid this and become aware of more
sensitive upstream changes in the future.

This commit only contains the infrastucture, no production whitelists
use the feature yet, they will rely on the DefaultDigester until we
decide how to continue from here.

### DIFF
--- a/rpmlint/checks/FileDigestCheck.py
+++ b/rpmlint/checks/FileDigestCheck.py
@@ -2,11 +2,123 @@ from fnmatch import fnmatch
 import hashlib
 from pathlib import Path
 import stat
+import xml.etree.ElementTree as ET
 
 from rpmlint.checks.AbstractCheck import AbstractCheck
 
 
 DEFAULT_DIGEST_ALG = 'sha256'
+
+
+# We support different "filters" for calculating file digests for whitelisting
+# purposes. The reasoning behind this is that we want to detect *functional*
+# changes to certain files but we don't want to be bothered by any other
+# changes like changes in comments, copyright headers, whitespace and so on.
+#
+# To avoid such bogus changes showing up in the checker we can filter the
+# target files depending on their file type. Currently we have two common
+# special cases:
+#
+# - shell like files (configuration files, shell scripts, Python
+# scripts, Perl scripts etc) that can contain empty lines or comments
+# introduced by '#'.
+# - XML files for things like D-Bus configuration files or polkit policies.
+# Here also XML style comments and whitespace can occur.
+#
+# The filter classes help in calculating a file digest for a filtered version
+# of the target file that is normalized and therefore stable against the
+# outlined kinds of changes. Since this makes getting the right whitelisting
+# digests hard, a small companion tool in tools/get_whitelisting_digest.py
+# exists.
+
+
+class DefaultDigester:
+    """This class performs the default digest calculation of arbitrary file
+    contents as is."""
+
+    def __init__(self, path, halg):
+        self.path = path
+        self.halg = halg
+
+    def parse_content(self):
+        # simply generate chunks on binary level
+        with open(self.path, 'rb') as fd:
+            while True:
+                chunk = fd.read(4096)
+                if not chunk:
+                    break
+                yield chunk
+
+    def get_digest(self):
+        """This returns the hash digest of the *un*filtered input file."""
+        hasher = hashlib.new(self.halg)
+
+        for chunk in self.parse_content():
+            hasher.update(chunk)
+
+        return hasher.hexdigest()
+
+
+class ShellDigester(DefaultDigester):
+    """This class performs digest calculation of shell style configuration
+    files or scripts. Certain aspects like comments and whitespace will be
+    filtered out of the digest calculation."""
+
+    def parse_content(self):
+        # generate filtered lines
+        with open(self.path) as fd:
+            for line_nr, line in enumerate(fd):
+                stripped = line.strip()
+                if not stripped:
+                    # skip empty lines
+                    continue
+                elif line_nr == 0 and stripped.startswith('#!'):
+                    # keep shebang lines instact
+                    pass
+                elif stripped.startswith('#'):
+                    # skip comments
+                    # NOTE: we don't strip trailing comments like in
+                    # 'if [ 5 -eq $NUM ]; then # compare to 5'
+                    # because the danger would be too high to remove actual
+                    # content instead of just comments
+                    continue
+
+                # don't use the completely stripped version, in Python for
+                # example the indentation is part of the syntax and changes
+                # here could change the meaning
+                yield (line.rstrip() + '\n').encode()
+
+
+class XmlDigester(DefaultDigester):
+    """This class performs digest calculation of XML configuration files.
+    Certain aspects like comments and whitespace will be filtered out of the
+    digest calculation."""
+
+    def parse_content(self):
+        # NOTE: the ElementTree is not robust against malicious input. Rpmlint
+        # processes a lot of untrusted input. In the OBS context this only
+        # happens within the virtual machines where packagers can manipulate a
+        # lot of things anyway so it should be okay.
+
+        # this returns a canonicalized form of the XML without comments or
+        # anything. It is missing the XML preamble and DOCTYPE declaration,
+        # though. It's as good as I could get it to be without parsing XML on
+        # foot or relying on third party XML modules.
+
+        # chunked / line wise processing is likely impossible for XML so
+        # return the whole bunch
+        yield ET.canonicalize(from_file=self.path, strip_text=True).encode()
+
+
+# These values can be used in the FileDigest configuration for the "Digester"
+# key. A digester is always the same for one type of FileDigest whitelistings
+# and thus not tied to individual whitelisting entries but to the type of a
+# FileDigestGroup.
+DIGESTERS = {
+    'default': DefaultDigester,
+    'shell': ShellDigester,
+    'xml': XmlDigester,
+}
 
 
 class FileDigestCheck(AbstractCheck):
@@ -15,10 +127,16 @@ class FileDigestCheck(AbstractCheck):
         self.digest_configurations = {}
         self.follow_symlinks_in_group = {}
         self.name_patterns_in_group = {}
+        self.digester_for_group = {}
         for group, values in self.config.configuration['FileDigestLocation'].items():
             self.digest_configurations[group] = [Path(p) for p in values['Locations']]
             self.follow_symlinks_in_group[group] = values['FollowSymlinks']
             self.name_patterns_in_group[group] = values.get('NamePatterns')
+            digester_name = values.get('Digester', 'default')
+            digester = DIGESTERS.get(digester_name, None)
+            if not digester:
+                raise Exception(f'Invalid digester {digester_name} encountered for group {group}')
+            self.digester_for_group[group] = digester
         self.ghost_file_exceptions = self.config.configuration.get('GhostFilesExceptions', [])
 
         self.digest_groups = self.config.configuration.get('FileDigestGroup', [])
@@ -99,7 +217,7 @@ class FileDigestCheck(AbstractCheck):
                     pass
         return None
 
-    def _is_valid_digest(self, path, digest, pkg):
+    def _is_valid_digest(self, group_type, path, digest, pkg):
         algorithm = digest['algorithm']
         if algorithm == 'skip':
             return (True, None)
@@ -108,7 +226,7 @@ class FileDigestCheck(AbstractCheck):
         if pkgfile is None:
             return (False, None)
 
-        file_digest = self._calc_digest(pkgfile, algorithm)
+        file_digest = self._calc_digest(group_type, pkgfile, algorithm)
         return (file_digest == digest['hash'], file_digest)
 
     def _resolve_links(self, pkg, path):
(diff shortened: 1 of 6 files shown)


rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
+                    # keep shebang lines instact
+                    pass
+                elif stripped.startswith('#'):
+                    # skip comments
+                    continue
+
+                # don't use the completely stripped version, in Python for
+                # example the indentation is part of the syntax and changes
+                # here could change the meaning
+                line_cb(line.rstrip() + "\n")
```
wfrisch, line 63: This efficient functional style saves memory. It may be harder to read for novices, but they're not the target audience. I like it.

wfrisch (approved):
Looks good, and also extendible in case of future whitelisting noise that needs to be filtered.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
+    """This class performs digest calculation of shell style configuration
+    files or scripts. Certain aspects like comments and whitespace will be
+    filtered out of the digest calculation."""
+
+    def __init__(self, path, halg):
+        self.path = path
+        self.halg = halg
+
+    def _parse_content(self, line_cb):
+        with open(self.path, 'r') as fd:
```
marxin, line 44: 'r' is default, please remove it.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
+            line_nr = 0
+            for line in fd:
```
marxin, line 46: Use enumerate if you need both value and index.

rpmlint/checks/FileDigestCheck.py
marxin, line 63: I would prefer returning byte content that can be later used for hash calculation.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
+
+                # don't use the completely stripped version, in Python for
+                # example the indentation is part of the syntax and changes
+                # here could change the meaning
+                line_cb(line.rstrip() + "\n")
+
+    def get_filtered(self):
+        """This just returns the filtered version of the input file."""
+        ret = []
+        self._parse_content(lambda line: ret.append(line))
```
marxin, line 68: Please just return bytes or string content here.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
+
+        return ''.join(ret)
+
+    def get_digest(self):
```
marxin, line 72: Please do this just once as get_filtered should return bytes (or string) and then you just calculate hash of it.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
+        """This returns the hash digest for the filtered version of the input
+        file."""
+        hasher = hashlib.new(self.halg)
+        normalized = self._parse_content()
+        hasher.update(normalized.encode())
+
+        return hasher.hexdigest()
+
+
+class DefaultDigester:
```
marxin, line 117: I think the 2 previous can inherit this class.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
         self.follow_symlinks_in_group = {}
         self.name_patterns_in_group = {}
+        self.digester_for_group = {}
         for group, values in self.config.configuration['FileDigestLocation'].items():
             self.digest_configurations[group] = [Path(p) for p in values['Locations']]
             self.follow_symlinks_in_group[group] = values['FollowSymlinks']
             self.name_patterns_in_group[group] = values.get('NamePatterns')
+            digester_name = values.get('Digester', 'default')
+            digester = DIGESTERS.get(digester_name, None)
+            if digester is None:
```
marxin, line 168: Just `if digester`.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,165 @@
...
+                raise Exception(f"Invalid digester {digester_name} encountered for group {group}")
```
marxin, line 169: Please use everywhere apostrophes, not quotes.

test/test_file_digest.py
```diff
@@ -221,6 +221,119 @@ def test_multiple_packages():
...
+# let's iterate over 1 to 10
+for i in `seq 10`; do
+    echo "$i"
+done
+
+fun() {
+    # what fun?!
+    echo "have fun!"  
+}'''  # noqa: W291
+
```
marxin, line 234: Please move this to separate files so that you don't have to use `noqa`. Similarly for expected output.

tools/get_whitelisting_digest.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/python3
+import argparse
+import os
+import sys
+
+# make rpmlint modules available
+sys.path.append(os.path.dirname(__file__) + "/../")
```
marxin, line 7: Use `os.path.join`.

tools/get_whitelisting_digest.py
```diff
@@ -0,0 +1,33 @@
...
+
+import rpmlint.checks.FileDigestCheck as fdc  # noqa: E402
+
+parser = argparse.ArgumentParser(
+    description="This tool helps calculating whitelisting digests for the FileDigestCheck"
+)
+parser.add_argument(
```
marxin, line 14: I don't like too many lines for this, please split it among less lines.

rpmlint/checks/FileDigestCheck.py
mgerstner: This class processes explicitly text based files i.e. we would need to encode it and then if a client needs text after all then it will need to be decoded again. And we do need the text version for the digest tool for debugging purposes.

Furthermore the `get_digest()` method is the canonical way to get digest from this and it does the encode() so I wouldn't change this.

rpmlint/checks/FileDigestCheck.py
mgerstner: If it is about the multiple `get_digest()` methods then switching to generator/`yield` approach might be the better way to avoid having to read in all data at once.

rpmlint/checks/FileDigestCheck.py
marxin: I fully understand the motivation, but I doubt we would reach memory consumption issues. Thus please make it simple and return the string from `get_filtered`.

rpmlint/checks/FileDigestCheck.py
mgerstner: Using a generator like it is now is likely clearer for any audience

mgerstner:
@marxin 

I now changed the digester interface to use a generator and return bytes() always. Transformation to text and error handling is now handled in the tool script only. Still we only process chunks or lines when possible to do so. I think this should address both concerns, clarity and efficiency.

mgerstner:
Good, thanks for the review and helping to improve the code. I've rebased the changes. Do you think we are ready to merge now?

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,135 @@
...
+class ShellDigester(DefaultDigester):
+    """This class performs digest calculation of shell style configuration
+    files or scripts. Certain aspects like comments and whitespace will be
+    filtered out of the digest calculation."""
+
+    def parse_content(self):
+        # generate filtered lines
+        with open(self.path) as fd:
+            for line_nr, line in enumerate(fd):
+                line_nr += 1
```
marxin, line 69: This is incremented automatically by `enumerate` function.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -9,16 +9,135 @@
...
+                yield (line.rstrip() + '\n').encode()
+
+
+class XmlDigester(DefaultDigester):
+    """This class performs digest calculation of XML configuration files.
+    Certain aspects like comments and whitespace will be filtered out of the
+    digest calculation."""
+
+    def parse_content(self):
+        import xml.etree.ElementTree as ET
```
marxin, line 98: Please import it in module context, this seems to me like not necessary optimization.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -120,23 +239,19 @@ def _resolve_links(self, pkg, path):
...
             return digest
 
-        h = hashlib.new(algorithm)
-        with open(pkgfile.path, 'rb') as fd:
-            while True:
-                chunk = fd.read(4096)
-                if not chunk:
-                    break
-                h.update(chunk)
+        Digester = self.digester_for_group[group_type]
```
marxin, line 251: Likely use `digester` here, similarly to previous `digester = DIGESTERS.get(digester_name, None)`.

tools/get_whitelisting_digest.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/python3
+import argparse
+import os
+import sys
+
+# make rpmlint modules available
+sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
```
marxin, line 7: Use apostrophes in this file instead of quotes.

marxin (changes requested):
Now it's almost fine, I have some minor comments.

marxin:
And please take a look at the FLAKE8 errors reported by CI.

rpmlint/checks/FileDigestCheck.py
mgerstner: Good catch. There was also still another bug lingering in the DefaultDigester that didn't actually loop until EOF.

tools/get_whitelisting_digest.py
mgerstner: I thought I catched all of them double quotes. Why aren't the flake8 warnings showing up in my local test run? Is there something special I need to do?

marxin: Note flake8 uses plugins, so it's very likely you miss one of these:
https://github.com/rpm-software-management/rpmlint/blob/8cc9efa1065835753cdb99e15a8016391ca9d9d9/.github/workflows/main.yml#L44-L65

Anyway, I'm going to add a test that will want about missing plug-in when pytest is run. That can simplify the development of the tool.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -2,23 +2,141 @@
...
+        # lot of things anyway so it should be okay.
+
+        # this returns a canonicalized form of the XML without comments or
+        # anything. It is missing the XML preamble and DOCTYPE declaration,
+        # though. It's as good as I could get it to be without parsing XML on
+        # foot or relying on third party XML modules.
+
+        # chunked / line wise processing is likely impossible for XML so
+        # return the whole bunch
+        yield ET.canonicalize(from_file=self.path, strip_text=True).encode()
```
marxin, line 110: Oh, CI shows that this won't work on Leap:
`E       AttributeError: module 'xml.etree.ElementTree' has no attribute 'canonicalize'`

It's pretty new (`3.8`):
https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.canonicalize

marxin (changes requested):
Right now, the minimal supported version is Python `3.6`, so what do you suggest about the XML digester? Do we care about Leap?

rpmlint/checks/FileDigestCheck.py
mgerstner: Yes I saw that before ... does it need to work on Leap?

marxin: > Are there any plans currently to backport rpmlint2 to Leap?

No, I actually refused to update rpmlint to version 2 in 15.4. That said, I think we can request version `3.8` and later. I'm going to discuss that in another pull request. For now, we can merge your pull request as it's only for the `opensuse` branch.

Conan-Kudo:
If we're adding extended functionality to rpmlint without adjusting defaults, please send it into the main codebase.

The compromise I struck originally for maintaining the opensuse vendor branch was that functionality goes into `main`, only policy goes into `opensuse`.

mgerstner:
> If we're adding extended functionality to rpmlint without adjusting defaults, please send it into the main codebase.
> 
> The compromise I struck originally for maintaining the opensuse vendor branch was that functionality goes into `main`, only policy goes into `opensuse`.

I'm not sure about all existing agreements about the opensuse branch (Martin surely knows better), but there have been longer discussions about these whitelisting checkers during rpmlint 2.0 development and the conclusion was that the way SUSE manages its security whitelistings is specific to SUSE workflows (RedHat uses its own workflows independent of rpmlint).

Conan-Kudo:
That is true, on the Fedora side, we have [rpminspect](https://github.com/rpminspect/rpminspect) for a lot of this. That said, I want it made clear that the `opensuse` branch is not an upstream branch, but a vendor one. So for the purposes of SUSE's upstream-first policy, merging into `opensuse` ***does not count*** for upstreaming.

marxin:
Yes, as @mgerstner tells. I feel id that all SUSE-specific checks should be in `opensuse` branch. While anything more generic (either enabled or disabled by default), should live in the `main` branch.

PR #801, opened 2022-02-11
Require at least Python 3.8.

marxin:
Right now, we would like to use some xml.etree functionality that
is there since 3.8: #800.

Thus I'm suggesting using Python 3.8 that's out for quite some time.

What do you think @Conan-Kudo about this pull request?


Conan-Kudo:
The only thing it does is break support for RHEL 8 and SLE 15, but if you don't have a problem with that, then neither do I.

Conan-Kudo (approved):

PR #804, opened 2022-02-20
Speed up FileDigestCheck.

marxin:
Use string trie for fast path lookup.

Speed up kernel-source.rpm with 63663 files from 7s to 0.5s.

### DIFF
--- a/rpmlint/checks/FileDigestCheck.py
+++ b/rpmlint/checks/FileDigestCheck.py
@@ -125,6 +125,8 @@ class FileDigestCheck(AbstractCheck):
     def __init__(self, config, output):
         super().__init__(config, output)
         self.digest_configurations = {}
+        # Build trie for fast lookup
+        self.digest_configuration_locations = {}
         self.follow_symlinks_in_group = {}
         self.name_patterns_in_group = {}
         self.digester_for_group = {}
@@ -137,6 +139,16 @@ class FileDigestCheck(AbstractCheck):
             if not digester:
                 raise Exception(f'Invalid digester {digester_name} encountered for group {group}')
             self.digester_for_group[group] = digester
+
+        # Build trie of Locations that are present in FileDigestLocation
+        for config_locations in self.digest_configurations.values():
+            for location in config_locations:
+                parts = Path(location).parts
+                node = self.digest_configuration_locations
+                for part in parts:
+                    node = node.setdefault(part, {})
+                node[None] = None
+
         self.ghost_file_exceptions = self.config.configuration.get('GhostFilesExceptions', [])
 
         self.digest_groups = self.config.configuration.get('FileDigestGroup', [])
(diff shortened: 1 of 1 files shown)


rpmlint/checks/FileDigestCheck.py
```diff
@@ -198,11 +210,30 @@ def _matches_pkg(self, digest_group, pkg):
...
+        """
+        Return true if there is a digest configuration that covers
+        provided file at given path.
+        """
+        parts = list(path.parts) + [None]
+        node = self.digest_configuration_locations
+        for part in parts:
+            if None in node:
+                return True
+            if part not in node:
```
mgerstner, line 223: This could be a tad more efficient by doing:

```
try:
    node = node[part]
except KeyError:
    return False
```

That avoids two identical dictionary lookups

rpmlint/checks/FileDigestCheck.py
```diff
@@ -198,11 +210,30 @@ def _matches_pkg(self, digest_group, pkg):
...
+                return False
+            node = node[part]
+        return True
```
mgerstner, line 226: I don't think this final return is correct. Consider we have a restriction for "/etc/cron.daily" and a package ships the file or dir "/etc". In that case the function would return `True`, although that is not really correct.

mgerstner (changes requested):

mgerstner:
Looks like I've been eight minutes too late :-P

marxin:
> Looks like I've been eight minutes too late :-P

Sorry about it, I merged that accidentally. Anyway, I'm going to address your comments.

marxin:
I've just reverted the pull request as what I merged was quite broken.

rpmlint/checks/FileDigestCheck.py
jsegitz: also just from a security POV is would like to see the logic to return False unless something is found, which makes later modifications more reliable

rpmlint/checks/FileDigestCheck.py
```diff
@@ -137,6 +139,16 @@ def __init__(self, config, output):
...
             self.digester_for_group[group] = digester
+
+        # Build trie of Locations that are present in FileDigestLocation
+        for config_locations in self.digest_configurations.values():
+            for location in config_locations:
+                parts = Path(location).parts
+                node = self.digest_configuration_locations
+                for part in parts:
+                    node = node.setdefault(part, {})
+                node[None] = None
```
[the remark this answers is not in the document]
marxin: I'm all for not using `[None] = None` as termination and I've also added the suggested hunk that ensures that one can's have 2 locations (in configuration) where one is a subfolder of the first one.

rpmlint/checks/FileDigestCheck.py
marxin: Sure.

rpmlint/checks/FileDigestCheck.py
marxin: The code is different now, but I wanted to not include exception machinery that tends to be slower than a normal dictionary lookup.

PR #810, opened 2022-02-21
Speed up FileDigestCheck.

marxin:
Use string trie for fast path lookup.
Speed up kernel-source.rpm with 63663 files from 7s to 0.5s.

This is follow up of reverted #804.


mgerstner (approved):
Look good now, thanks!

PR #812, opened 2022-02-22
Modem manager dbus adjusted path

mgerstner:


configs/openSUSE/dbus-services.toml
```diff
@@ -476,9 +476,10 @@ nodigests = [
 package = "ModemManager"
 type = "dbus"
 note = "imported from rpmlint1 DBUSServices.WhiteList"
-bug = "bsc#798273"
+bug = ["bsc#798273", "bsc#1196170"]
+note = "Mobile broadband modem control via D-Bus service, vast and complex interface"
```
marxin, line 480: This introduces one another `note` property, about which TOML complains:

```
(none): E: fatal error while parsing configuration file configs/openSUSE/dbus-services.toml: Duplicate keys! (line 480 column 1 char 13275)
```

marxin (changes requested):

configs/openSUSE/dbus-services.toml
wfrisch: Apart from that it looks good to me. Feel free to merge after the duplicate note is removed.

mgerstner: Completely missed that. Fixed now.

PR #823, opened 2022-03-03
D-Bus services whitelist: add kpmcore (bsc#1178848)

mgerstner:
The upstream code is still a bit shaky but we are whitelisting
nonetheless, because the existing kpmcore package in Factory is even
worse (running completely as root, currently broken build).


thesp0nge (approved):
Ok

PR #824, opened 2022-03-03
missing alpha_3 hye added for Armenian

aekoroglu:
I was updating a xfce package and rpmlint gave an error as "E: invalid-lc-messages-dir /usr/share/locale/hye/LC_MESSAGES/xfce4-taskmanager.mo" then I tested for some other xfce packages from fedora35 and got the same error. I also checked xfce packages on Ubuntu + Archlinux + Pardus and they've the locale.


marxin:
Thanks for the pull request.

Issue #847, opened 2022-04-25
Check for broken R(UN)PATH


Vogtinator:
Currently rpmlint doesn't heavily complain if a binary has a broken runpath e.g. pointing inside `rpmbuild`.

I found this in an older audacity RPM (audacity-3.0.5-4.2.x86_64.rpm) from Tumbleweed:

`(RUNPATH)            Library runpath: [/home/abuild/rpmbuild/BUILD/audacity-Audacity-3.0.5/build/bin/RelWithDebInfo/lib64/audacity:/home/abuild/rpmbuild/BUILD/audacity-Audacity-3.0.5/build/shared/RelWithDebInfo]`

marxin:
Hmm, the current check is pretty broken and also disabled in our openSUSE config.
Let me fix that.

marxin:
With the suggested pull request I see:

```
...
audacity.x86_64: E: binary-or-shlib-defines-rpath /usr/lib64/audacity/lib-string-utils.so /home/abuild/rpmbuild/BUILD/audacity-Audacity-3.0.5/build/bin/RelWithDebInfo/lib64/audacity
audacity.x86_64: E: binary-or-shlib-defines-rpath /usr/lib64/audacity/lib-strings.so /home/abuild/rpmbuild/BUILD/audacity-Audacity-3.0.5/build/bin/RelWithDebInfo/lib64/audacity
audacity.x86_64: E: binary-or-shlib-defines-rpath /usr/lib64/audacity/lib-utility.so /home/abuild/rpmbuild/BUILD/audacity-Audacity-3.0.5/build/bin/RelWithDebInfo/lib64/audacity
audacity.x86_64: E: binary-or-shlib-defines-rpath /usr/lib64/audacity/lib-uuid.so /home/abuild/rpmbuild/BUILD/audacity-Audacity-3.0.5/build/bin/RelWithDebInfo/lib64/audacity
audacity.x86_64: E: binary-or-shlib-defines-rpath /usr/lib64/audacity/modules/mod-script-pipe.so /home/abuild/rpmbuild/BUILD/audacity-Audacity-3.0.5/build/bin/RelWithDebInfo/lib64/audacity
The binary or shared library defines `RPATH' (or `RPATH').
```

PR #848, opened 2022-04-25
Fix up binary-or-shlib-defines-rpath error.

marxin:
The error should be emitted if it does not point to
a system library folder or /usr/lib*.

Enable the check in openSUSE config.

Fixes: #847.

### DIFF
--- a/rpmlint/descriptions/BinariesCheck.toml
+++ b/rpmlint/descriptions/BinariesCheck.toml
@@ -46,7 +46,7 @@ the libtool wrapper file run
 to install the relinked file.
 """
 binary-or-shlib-defines-rpath="""
-The binary or shared library defines `RPATH'.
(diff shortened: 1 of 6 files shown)


rpmlint/descriptions/BinariesCheck.toml
```diff
@@ -46,7 +46,7 @@ the libtool wrapper file run
 to install the relinked file.
 """
 binary-or-shlib-defines-rpath="""
-The binary or shared library defines `RPATH'.
+The binary or shared library defines `RPATH' (or `RPATH').
```
Vogtinator, line 49: `RUNPATH`?

marxin:
@Vogtinator: Do you think we should increase the badness of the error?

marxin:
> IMO yes. Previously it was a fatal error.

Agree, done.

PR #849, opened 2022-04-25
Update shlib-policy-name-error score for openSUSE

jengelh:
In openSUSE Leap 15.3, rpmlint 1.x reported shlib-policy-name-error
instances with a 10000 point penalty. Restore this for rpmlint 2.x.

### DIFF
--- a/configs/openSUSE/scoring.toml
+++ b/configs/openSUSE/scoring.toml
@@ -25,6 +25,7 @@ percent-in-conflicts = 10000
 percent-in-dependency = 10000
 percent-in-obsoletes = 10000
 percent-in-provides = 10000
+shlib-policy-name-error = 10000
 spurious-executable-perm = 50
 summary-ended-with-dot = 20
 summary-not-capitalized = 20


marxin:
As mentioned in #730, there's still quite some fallout of the `shlib-policy-name-error` as seen here:
https://bugzilla.opensuse.org/show_bug.cgi?id=1191323 (still 15 packages needs to be resolved).

Moreover, the `shlib-policy-name-error` was quite reworked in `2.x` and I'm not sure it's comparable to `1.x`. The # of affected packages proves that.

So can we first identify & somehow fix the remaining packages?

jengelh:
Yes, it was reworked in #746, the claims in [1191323](https://bugzilla.opensuse.org/show_bug.cgi?id=1191323) should be resolved. Current cases are at https://rpmlint.opensuse.org/openSUSE:Factory/x86_64/standard?rule=shlib-policy-name-error , and most of them point in the direction of human error (too low error score, build “succeeded” when it shouldn't).

```
[  139s] libKF5AkonadiSearch.x86_64: E: shlib-policy-name-error SONAME: libKF5AkonadiSearchCore.so.5, expected package suffix: 5
[  174s] libAnalitza5.x86_64: E: shlib-policy-name-error SONAME: libAnalitza.so.8, expected package suffix: 8
[   51s] libatlascpp-0_6-1.x86_64: E: shlib-policy-name-error SONAME: libAtlas-0.6.so.3, expected package suffix: 0_6-3
[ 3898s] libaugeas0.x86_64: E: shlib-policy-name-error SONAME: libfa.so.1, expected package suffix: 1
[  268s] libcapnp-0_9.x86_64: E: shlib-policy-name-error SONAME: libcapnp-0.9.1.so, expected package suffix: 0_9_1
[   61s] libcdk5.x86_64: E: shlib-policy-name-error SONAME: libcdk.so.6.2.4, expected package suffix: 6_2_4
...
```

marxin:
> Yes, it was reworked in #746, the claims in [1191323](https://bugzilla.opensuse.org/show_bug.cgi?id=1191323) should be resolved. Current cases are at https://rpmlint.opensuse.org/openSUSE:Factory/x86_64/standard?rule=shlib-policy-name-error , and most of them point in the direction of human error (too low error score, build “succeeded” when it shouldn't).

Are you willing to resolve all the remaining issues of this error? It's unpleasant that a badness score increment will eventually block rpmlint from being accepted to Factory. Apparently, package maintainers haven't resolved all the reported bugs..

marxin:
All right, so I'm ready to merge this as soon as we'll have Staging packages resolved.

marxin:
@jengelh Any progress about the 8 packages that still violate this check?

e4t:
> @jengelh Any progress about the 8 packages that still violate this check?

There are more than 8 packages that fail. Most notably the HPC packages. For HPC this check does not make any sense:

- shared libraries are never installed in a system directory nor any directory mentioned in /etc/ld.so.conf or /etc/ld.so.conf.d/*
- multiple package versions of the same .so major version may have to be installed in parallel.

To accomplish this, packages use a package version string (with '.' replaced by '_') instead of .SO versions - in addition to name elements denoting the toolchain and MPI version used.
To select the desired library stack, the well known [module system](https://hpc-wiki.info/hpc/Modules) is utilized. With this a user  is able to select a library stack to use per session.

I believe, the tests should be improved to not blindly grab all shared libraries but only those residing in system paths or paths provided on /etc/ld.so.conf* files. This may also help to exclude other types of plugins which are only loaded on request of an application or library.

marxin:
> There are more than 8 packages that fail. Most notably the HPC packages. For HPC this check does not make any sense:

Yes, speaking out of Ring0+1, there are only a few. But in Factory there are almost 150 such errors:
https://rpmlint.opensuse.org/openSUSE:Factory/x86_64/standard?rule=shlib-policy-name-error

> To accomplish this, packages use a package version string (with '.' replaced by '_') instead of .SO versions - in addition to name elements denoting the toolchain and MPI version used. To select the desired library stack, the well known [module system](https://hpc-wiki.info/hpc/Modules) is utilized. With this a user is able to select a library stack to use per session.

If I see correctly, the gsl libraries sit in:
```
/usr/lib/hpc/gnu12/gsl/2.6/lib64/libgsl.so.25
/usr/lib/hpc/gnu12/gsl/2.6/lib64/libgsl.so.25.0.0
```

> I believe, the tests should be improved to not blindly grab all shared libraries but only those residing in system paths or paths provided on /etc/ld.so.conf* files. This may also help to exclude other types of plugins which are only loaded on request of an application or library.

How exactly should we identify that? By ignoring all sub-folders of the standard system paths?

marxin:
Adding Factory folks @fvogt and @Dimstar77

jengelh:
>tests should be improved to not blindly grab all shared libraries but only those residing in system paths

I do not think rpmlint can reliably determine that. /etc/ld.so.conf.d/* files are a _runtime_ thing and may potentially be absent in the _build_ chroot. With a filename like `libgsl.so.25`, the chance is pretty high it is _not dlopened_ but loaded on program start, and so would fall under SLPP no matter the location. Some programs may even be started with shell wrappers that set `LD_LIBRARY_PATH` rather than using a ld.so.conf.d fragment, and that loops back that SLPP is desirable irrespective of paths.

jengelh:
All packages with multibuild in them should just ignore the error for the time being. Since filters are regexes, it's possible to not apply the filter to the base package but only the variants. E.g.

```
addFilter("libvtk1-\S+: E: shlib-policy-name-error SONAME:.*")
```

e4t:
> If I see correctly, the gsl libraries sit in:
> 
> ```
> /usr/lib/hpc/gnu12/gsl/2.6/lib64/libgsl.so.25
> /usr/lib/hpc/gnu12/gsl/2.6/lib64/libgsl.so.25.0.0
> ```
> 
> > I believe, the tests should be improved to not blindly grab all shared libraries but only those residing in system paths or paths provided on /etc/ld.so.conf* files. This may also help to exclude other types of plugins which are only loaded on request of an application or library.
> 
> How exactly should we identify that? By ignoring all sub-folders of the standard system paths?

Well,  I'd say /usr/lib/hpc/gnu12/gsl is quite far away from any standard system path, isn't it?

> > tests should be improved to not blindly grab all shared libraries but only those residing in system paths
> 
> I do not think rpmlint can reliably determine that. /etc/ld.so.conf.d/* files are a _runtime_ thing and may potentially be absent in the _build_ chroot. 

When rpmlint runs the packages have been created and installed on the system, so /etc/ld.so.conf.d/* files should be part of them, shouldn't they?

> With a filename like `libgsl.so.25`, the chance is pretty high it is _not dlopened_ but loaded on program start, and so would fall under SLPP no matter the location. Some programs may even be started with shell wrappers that set `LD_LIBRARY_PATH` rather than using a ld.so.conf.d fragment, and that loops back that SLPP is desirable irrespective of paths.

Where did I say dlopen() is used? 
Environment modules in HPC do exactly that: set LD_LIBRARY_PATH and numerous other paths exactly the way the user has requested modules to be loaded. Programs that use shell wrappers to pluck *their* shared from a specific non-standard path do so generally because they need a custom version of this library which is not meant for public consumption. These libraries are exactly those which would qualify for *not* having the SLPP applied. Generally, packages requiring such libraries have means to ensure the correct libraries are installed and it is unlikely, these are packaged in separate SLPP-compliant shared library packages.
What is the shared library packaging policy for? To help those programs that use *standard* shared libraries to find the correct library without having to worry about multiple so version being around.

marxin:
I think it's the right time to merge this.

marxin:
Triggers #901.

PR #850, opened 2022-04-27
support $ORIGIN in R{,UN}PATH.

marxin:
The following is valid use-case:
$ readelf -Wd /usr/lib64/qt5/examples/multimedia/spectrum/spectrum | grep RUN
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]


marxin:
Related to #847.

Vogtinator:
`$ORIGIN` is just as invalid as any other path

marxin:
All right, so let me revert that.

PR #859, opened 2022-05-06
ErlangCheck: remove unneeded str() from the test

matwey:
str() triggered beam-was-not-recompiled warning unconditionally.


marxin:
Thanks for the pull request.

PR #865, opened 2022-05-17
SystemdTmpfilesCheck: new check to restrict systemd-tmpfiles configur…

mgerstner:
…ation

This introduces a new check that restricts what kind of systemd-tmpfiles
configuration file entries can be shipped by a package. This supplements
the existing security related checkers like FileDigestCheck. Since this
check requires parsing of the custom systemd-tmpfiles format the generic
existing checkers cannot be reused for this.

This change does not yet raise badness in the strict configuration,
because the check first needs some realife testing and the affected
packages first need to be reviewed and whitelisted before actually
generating badness.

### DIFF
--- a/rpmlint/checks/SystemdTmpfilesCheck.py
+++ b/rpmlint/checks/SystemdTmpfilesCheck.py
@@ -1,0 +1,269 @@
+from pathlib import Path
+import stat
+
+from rpmlint.checks import TmpfilesParser
+from rpmlint.checks.AbstractCheck import AbstractCheck
+
+# don't add tmp directories here, they could still be dangerous, since
+# publicly accessible
+WELL_KNOWN_PATHS = (
+    '/var/lib', '/var/run', '/run', '/var/lock', '/var/cache',
+    '/run/lock', '/var/log', '/var/spool', '/srv', '/etc',
+    # some configs somehow reset /sys permissions here
+    '/sys',
+    # moderately well known, NIS stuff
+    '/var/yp'
+)
+
+# chattr() flags that are not interesting security wise
+BORING_ATTRS = (
+    # disable copy-on-write on supported file-systems
+    'C',
+)
+
+
+class SystemdTmpfilesCheck(AbstractCheck):
+    """This check restricts content of systemd-tmpfiles configuration files.
+
+    tmpfiles configuration can grant a wide range of properties from arbitrary
+    file modes, ACLs to creating device files and more. Some of them can be
+    security relevant, therefore we want to control what can and what cannot
+    enter the distribution.
+    """
+
+    def __init__(self, config, output):
+        super().__init__(config, output)
+        # Build trie for fast lookup of restricted paths
+        self.dropin_dir_trie = {}
+
+        check_config = self.config.configuration['SystemdTmpfiles']
+        # these are the tmpfiles.d drop-in dir locations we are restricting
+        self.dropin_dirs = check_config['DropinDirs']
+        # this is a list of base packages that we won't care about for
+        # simplicity (like systemd packages themselves)
+        self.ignore_packages = check_config['IgnorePackages']
+
+        self._setup_location_trie()
+
+        self.whitelist = self.config.configuration.get('SystemdTmpfilesWhitelist', [])
+        for entry in self.whitelist:
+            self._sanity_check_whitelist_entry(entry)
+            self._normalize_whitelist_entry(entry)
+
+    def _setup_location_trie(self):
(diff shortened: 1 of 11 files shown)


wfrisch (approved):
I can't see anything wrong at first sight.
For completeness' sake I fed all tmpfiles configs in Factory to the parser.

thesp0nge (approved):

rpmlint/checks/SystemdTmpfilesCheck.py
```diff
@@ -0,0 +1,270 @@
...
+
+        if 'path' not in entry:
+            raise KeyError('SystemdTmpfilesCheck: missing "path" key')
+
+    def _is_in_safe_location(self, tf_entry):
+        """Returns whether the given tmpfiles configuration entry affects a
+        safe location that is usually okay to operate in."""
+        # don't add tmp directories here, they could still be dangerous, since
+        # publicly accessible
+        WELL_KNOWN = (
```
marxin, line 109: Please move this to the beginning of the file, similarly as we do for other checks.

rpmlint/checks/SystemdTmpfilesCheck.py
```diff
@@ -0,0 +1,270 @@
...
+                # world readable or writeable, sensitive
+                return True
+
+            # root only access, should be fine
+            return False
+        elif base_type in ('h', 'H'):
+            # sets chattr() attributes, possibly recursively (upper case)
+            attr = self.m_arg
+
+            BORING_ATTRS = (
```
marxin, line 195: Simiarly here.

rpmlint/checks/SystemdTmpfilesCheck.py
```diff
@@ -0,0 +1,270 @@
...
+            return False
+
+        # other / yet unknown type, treat as sensitive
+        return True
+
+    def _check_for_whitelisting(self, pkg, pkgfile, tf_entry):
+        """Checks whether the given tempfile entry has a valid whitelisting
+        configured and returns the boolean result."""
+        for wl_entry in self.whitelist:
+
```
marxin, line 217: No blank like after `for`.

rpmlint/checks/TmpfilesParser.py
```diff
@@ -0,0 +1,166 @@
...
+            # see _supports_perms()
+            self.type = 'f+' + self.type[1:]
+
+    def _supports_perms(self):
+        """Returns whether the current tmpfiles entry supports a file
+        permissions field."""
+        # NOTE: for 'C' the table in the man page does not list permissions
+        # support but actually they *are* supported.
+        # NOTE: F is a deprecated form of 'f+'
+        return self.type[0] in ('f', 'F', 'd', 'D', 'e', 'v', 'q', 'Q', 'p', 'p+', 'c', 'C', 'b', 'z', 'Z')
```
marxin, line 82: I would likely define as a class constant.

rpmlint/checks/TmpfilesParser.py
```diff
@@ -0,0 +1,166 @@
...
+        return '-' if self._supports_perms() else 'N/A'
+
+    def get_type(self):
+        return self.type
+
+    def get_config_file(self):
+        return self.source
+
+    def get_target_path(self):
+        return self.target
```
marxin, line 94: Do we really need this getters?

marxin (changes requested):
Thanks for the submission, in general it looks fine.
Can you please add some tests to https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests ?

rpmlint/checks/TmpfilesParser.py
mgerstner: Not at the moment but I'd like the types's interface to be complete.

mgerstner:
> Can you please add some tests to https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests ?

I've never used that project. How is it supposed to work? Won't a unit test do?

marxin:
> I've never used that project. How is it supposed to work? Won't a unit test do?

Yes, a unit test would be ideal. The OBS location is a place where we save spec files for RPM files used in unit tests.

marxin (approved):

marxin:
Or we can utilize the newly creates `FakePkg` class, use what you prefer.

mgerstner:
All right, I squashed the commits, added test coverage. I think it should be complete now.

marxin:
Thanks for the contribution.

PR #868, opened 2022-05-19
openSUSE: Do not allow files in /etc/NetworkManager/dispatcher.d anymore

schubi2:
Following packages has generated files in /etc/NetworkManager/dispatcher.d: 
chronny, tlp, autofs. 
This has already been fixed to. They are writing to /usr/lib/NetworkManager/dispatcher.d now.
So, users changes can still be done in /etc and will not be overwritten by an update.
NetworkManager is supporting /usr/lib/NetworkManager/dispatcher.d since SLES12.
So there is not reason for writing to /etc/NetworkManager/dispatcher.d  anymore

### DIFF
--- a/configs/openSUSE/opensuse.toml
+++ b/configs/openSUSE/opensuse.toml
@@ -15,6 +15,10 @@ Checks = [
     "SharedLibraryPolicyCheck",
 ]
 
+# List of directory prefixes that are not allowed in packages
+DisallowedDirs = [
+    "/etc/NetworkManager/dispatcher.d",
+]
 
 Filters = [
 # Stuff autobuild takes care about


marxin:
@mgerstner: Can you please chime in? What do you think about the pull request?

schubi2:
> @mgerstner: Can you please chime in? What do you think about the pull request?
Mathias, Thorsten Kukuk has "given the order" to prevent packages from generating files into this directory :-)

mgerstner:
This check is not really securit related so our team is not directly concerned. It is more about package quality management. Notwithstanding, since we're increasingly moving from /etc to /usr I don't see any problems with this.

PR #880, opened 2022-06-07
Put systemd-tmpfiles check into production mode

mgerstner:


FilippoBonazziSUSE (approved):
Looks good to me

configs/openSUSE/systemd-tmpfiles.toml
```diff
@@ -6,3 +6,93 @@ path = "/usr/lib/tmpfiles.d/rpmlint-test.conf"
...
+path = "/usr/lib/tmpfiles.d/tmux.conf"
+entries = [
+	"d /run/tmux 1777 root root -"
+]
+
+[[SystemdTmpfilesWhitelist]]
+package = "libtss2-fapi1"
+bug = "bsc#1150905"
+note = "shared setgid tss directories for the feature API framework"
+path = "/usr/lib/tmpfiles.d/tpm2-tss-fapi-3.1.0.conf"
```
DimStar77, line 23: That's a bad one - we actually just got hit by tpm2-0-tss being updated to version 3.2.0 in parallel to the rpmlint change.

With the path name containing version information, this is now no longer covering the current version

https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:A/tpm2-0-tss/standard/x86_64

mgerstner: That is pretty strange that upstream encodes the version in the systemd-tmpfiles configuration file. Maybe we can change the name in the packaging or via a patch. Maybe the rationale behind it is having multiple versions installed in parallel ... we could also adjust our whitelisting mechanism to allow for wildcards. Coupled with content based whitelistings this would still be safe enough. I will look into it.

mgerstner: Until we find a better solution we can just adjust the whitelisting to the new version.

DimStar77: > Until we find a better solution we can just adjust the whitelisting to the new version.

That would be appreciated, as I currently have all stagings fail on tpm2-0-tss, like this:

https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:K/tpm2-0-tss/standard/x86_64

mgerstner: We should already be able to use wildcards and thereby fix that easily. However the cause for this version number in the filename is in our packaging, not upstream. I just asked Alberto for the reason for this change. When I have all information we should be able to move on one way or another.

aplanas: IIRC the original reason was a subpackage rename, causing a conflict during the upgrade in SLE.

I think that is safe to revert the naming in TW now, and do the same later in SLE if there is some other fix to include in the package. Would this resolve the issue?

mgerstner: In PR #897 and OBS sr#988349 the version number is dropped from the filename again. The original reason for the version number was to avoid a RPM level file conflict in an upgrade situation on SLE. It should be safe to drop that now at least for Factory.

PR #884, opened 2022-06-11
Revert "Use zypper for openSUSE."

Conan-Kudo:
Tumbleweed and Leap 15.4 now use MirrorCache instead of MirrorBrain
with DNF, so this should be considerably more reliable.

This reverts commit 6ae9b2d93652d97fb864507f109d264b8e85cd3e.


marxin:
I welcome that, please merge it.

PR #890, opened 2022-06-17
Fix rpm_buildroot_regex

tobbez:
The regex was completely broken for the case of $RPM_BUILD_ROOT, and the
cause was twofold:

  - In the capture group matching backslashes before a `$`, the asterisk
    (`*`) is escaped when it shouldn't be.

    Before 6de3730a1bda8f2705287186b8af8853dd16e5f9, the capture group
    was incorrectly escaped as `'\\\*'`, but still got interpreted as
    `'\\\\*'` (as `\*` is not a valid Python string escape) so `\\*` was
    passed to the regex module.

    After that commit, with `r'\\\*'`, what the regex module saw was
    instead `\\\*`, meaning: a literal match for the string `\*`.

    We fix this by using the correct number of bashslashes for a raw
    string literal (2).

  - The initial part of the regex, `^[^#]*`, consumed all backslashes
    preceeding a `$`.

    We fix this by making the match non-greedy (`*` => `*?`).

### DIFF
--- a/rpmlint/checks/SpecCheck.py
+++ b/rpmlint/checks/SpecCheck.py
@@ -33,7 +33,7 @@ prereq_regex = re_tag_compile(r'PreReq(\(.*\))')
 
 make_check_regex = re.compile(r'(^|\s|%{?__)make}?\s+(check|test)')
 rm_regex = re.compile(r'(^|\s)((.*/)?rm|%{?__rm}?) ')
-rpm_buildroot_regex = re.compile(r'^[^#]*(?:(\\\*)\${?RPM_BUILD_ROOT}?|(%+){?buildroot}?)')
+rpm_buildroot_regex = re.compile(r'^[^#]*?(?:(\\*)\${?RPM_BUILD_ROOT}?|(%+){?buildroot}?)')
 configure_libdir_spec_regex = re.compile(r'ln |\./configure[^#]*--libdir=(\S+)[^#]*')
 lib_package_regex = re.compile(r'^%package.*\Wlib')
 ifarch_regex = re.compile(r'^\s*%ifn?arch\s')
--- a/test/spec/rpm-buildroot-usage-shell-var.spec
+++ b/test/spec/rpm-buildroot-usage-shell-var.spec
@@ -1,0 +1,33 @@
+Name:           rpm-buildroot-usage-shell-var
+Version:        0
+Release:        0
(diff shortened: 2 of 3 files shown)


marxin:
Thanks for the pull request. Can you please add tests that cover the problem (`test/tests_speccheck.py` tests like `test_check_rpm_buildroot_usage_under_prep` can be inspiration).

tobbez:
Actually, after looking closer at it, I think the `no-cleaning-of-buildroot` check should be outright removed:

It has been broken (producing false positives) ever since 6de3730a1bda8f2705287186b8af8853dd16e5f9, which was first included in rpmlint 1.10 (released September 2017). This indicates that practically no-one uses it (otherwise, at least one person should've noticed and filed a bug report before now).

It's also filtered by both the Fedora and openSUSE configs.

[The comment](https://github.com/rpm-software-management/rpmlint/blob/69c7ba23acfd663fd5e39b01f0a6a483b00a0836/configs/Fedora/fedora.toml#L278-L279) for the entry in Fedora's config states:

> Fedora 12 and newer no longer need a buildroot defined, to have the buildroot cleaned at the beginning of %install, and do not need to define a %clean section unless the default is invalid.

Fedora 12 was released in 2010.

Similarly, the openSUSE spec file guidelines [state that cleaning should not be done manually](https://en.opensuse.org/openSUSE:Specfile_guidelines#Removing_the_buildroot).

In addition, the check is also in direct conflict with `superfluous-%clean-section` ("The spec section %clean should not be used any longer. RPM provides its own clean logic"), producing the warning (`W: no-cleaning-of-buildroot %clean`) when no `%clean` section is present in the spec file.

As a final nail in the coffin, there's [this commit from 2020 adding automatic cleaning in %install upstream in RPM](https://github.com/rpm-software-management/rpm/commit/f49671d6f3442cfff08d6167b618607e96cf0970), stating:
> Variations of this exist in virtually every RPM-based Linux distribution
for more than a decade. At this point, it is expected behavior everywhere
and it is amazing that it has not yet been put in RPM itself until now...

marxin:
> Actually, after looking closer at it, I think the `no-cleaning-of-buildroot` check should be outright removed:

Works for me and I merged the other pull request.

For this one, please add a test coverage.

marxin:
Great, appreciate your contribution.

PR #891, opened 2022-06-21
Remove obsolete check no-cleaning-of-buildroot

tobbez:
As discovered in PR #890.

RPM distributions have had automatic cleaning of the buildroot in %install for a long time. Quoting [the commit](https://github.com/rpm-software-management/rpm/commit/f49671d6f3442cfff08d6167b618607e96cf0970) that finally introduced buildroot cleaning in %install directly in RPM in 2020:

> Variations of this exist in virtually every RPM-based Linux distribution
> for more than a decade. At this point, it is expected behavior everywhere
> and it is amazing that it has not yet been put in RPM itself until now...

The buildroot is also cleaned automatically by injection of a default %clean section (if not present) since RPM 4.8.0 (released 2010), and there is a rpmlint check producing an error if %clean is included in the a spec file (superfluous-%clean-section).

In addition, the check cannot have seen significant use elsewhere, as it has been broken since rpmlint 1.10[1] (released September 2017). The issue manifested as false positives (incorrectly reporting a warning for spec files that did clean the buildroot in %install/%clean). If the check had been in active use, this issue should have been discovered much
earlier.

[1]: commit 6de3730a1bda8f2705287186b8af8853dd16e5f9

### DIFF
--- a/configs/Fedora/fedora.toml
+++ b/configs/Fedora/fedora.toml
@@ -275,9 +275,6 @@ Filters = [
     'dangling-\S*symlink /usr/share/doc/HTML/\S+/common .+/common$',
     'hidden-file-or-dir .*/man5/\.k5login\.5[^/]+$',
     'blender.+ (wrong-script-interpreter|non-executable-script) .+/blender/.+\.py.*BPY.*',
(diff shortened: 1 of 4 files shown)


marxin:
Works for me, thanks.

PR #893, opened 2022-07-04
`keyserver` user and group

tacerus:
Signed-off-by: Georg Pfuetzenreuter <georg.pfuetzenreuter@suse.com>

Hi!

This is for a software we are forking from https://github.com/mailvelope/keyserver and currently still packaging in https://build.opensuse.org/package/show/home:crameleon/keyserver.

Best,
Georg


tacerus:
Thank you very much, @marxin!

PR #897, opened 2022-07-11
systemd-tmpfiles: drop version from libtss2-fapi1 filename

mgerstner:
OBS sr#988349 removes the version number from the filename again,
because it causes trouble during updates. The version number was
originally introduced to fix a file conflict in a package rename
context.


marxin (approved):

marxin:
I've just made a SR with this change:
https://build.opensuse.org/request/show/988371

PR #898, opened 2022-07-12
Whitelistings: Some dbus-service path moves; make sha256 default algorithm

mgerstner:


mgerstner:
Since this is a bigger change I leave merging to Martin, please also wait for a review of at least one security colleague, thanks.

configs/openSUSE/dbus-services.toml
```diff
@@ -240,10 +249,19 @@ nodigests = [
...
+bugs = [ "bsc#1201345" ]
+digests = [
+    {
+        path = "/usr/share/dbus-1/system.d/avahi-dbus.conf",
+        digester = "xml",
+        hash = "047f2e04019b25389c08c3c211dfdc595ca7ee1adfc3da6cf85fabb536ebbb7b"
+    },
+    {
+        path = "/usr/share/dbus-1/system-services/org.freedesktop.Avahi.service",
+        digest = "shell",
```
marxin, line 262: Please fix this (caught by tests now).

marxin (changes requested):

configs/openSUSE/dbus-services.toml
```diff
@@ -33,10 +30,14 @@ digests = [
...
 type = "dbus"
-note = ""
-bug = "bsc#515977"
-nodigests = [
-    "/etc/dbus-1/system.d/cups.conf",
+note = "the well known printing system daemon"
+bugs = [ "bsc#515977", "bsc#1201346" ]
+digests = [
+    {
+        path = "/etc/dbus-1/system.d/cups.conf",
```
FilippoBonazziSUSE, line 37: I thought bsc#1201346 was about moving the config file from `/etc/dbus-1/system.d/cups.conf` to `/usr/share/dbus-1/system.d/cups.conf`. Is this path right?

configs/openSUSE/dbus-services.toml
mgerstner: Good catch. Is fixed now.

configs/openSUSE/pam-modules.toml
```diff
@@ -238,7 +239,8 @@ nodigests = [
 [[FileDigestGroup]]
 package = "sssd"
 type = "pam"
-note = "legacy: not audited"
+note = "authenticate and get user information via sssd suite"
+bug = "bsc#1106600"
```
FilippoBonazziSUSE, line 243: These remain without digests?

FilippoBonazziSUSE (commented):
Other than above questions all looks good to me

configs/openSUSE/dbus-services.toml
mgerstner: Good point. Seems I was too busy with the digests to do the actual task ;). I addressed this now.

Issue #904, opened 2022-07-21
Use tomllib from the standard library on Python 3.11+


hroncok:
Hey,
would you accept a pull request that uses tomllib rom the standard library on Python 3.11+ and fallbacks to tomli, instead of toml?

marxin:
Sure, please do so.

PR #905, opened 2022-07-21
Use tomllib from the standard library on Python 3.11+

hroncok:
Fallback to tomli otherwise, as it has the same API.

Fixes https://github.com/rpm-software-management/rpmlint/issues/904

I wasn't able to run all the tests locally, so opening this PR as a draft.


marxin:
Note we won't need `tomli_w` as we don't write a TOML files (if I remember correctly).

marxin:
You'll need to install the new package (`python3-tomli` and `python3-tomli-w`) in `.github/workflows/main.yml`
and please fix the import order in order to fix `flake8` issue:

```python
from rpmlint.helpers import print_warning
try:
    import tomllib
except ImportError:
    import tomli as tomllib
import tomli_w
from xdg.BaseDirectory import xdg_config_dirs
```

hroncok:
> ~Note we won't need `tomli_w` as we don't write a TOML files (if I remember correctly).~
> 
> EDITTED: Oh we do: `print(tomli_w.dumps(self.configuration))`

Yeah, I assumed that at first as well.

hroncok:
I have trouble understanding the dnf stuff in the GitHub action file. there is no python3-base in Fedora.

.github/workflows/main.yml
```diff
@@ -49,10 +49,13 @@ jobs:
               python3-flake8-quotes
               python3-pyxdg
               python3-zstandard
-              python3-toml
+              python3-tomli-w
               python3-pip
               rpm-build
               git
+        if: ${{ ! contains(matrix.container, 'rawhide') }}
```
marxin, line 56: This line should go below `if: ${{ contains(matrix.container, 'opensuse') }}` as these packages are only for openSUSE.

marxin (changes requested):

.github/workflows/main.yml
hroncok: Right, the if is below the run. OMG. Now I think I get it.

marxin:
Good, the last missing piece are missing package in `.packit/rpmlint.spec`, please add that.

marxin:
Thanks for the pull request.

hroncok:
Thanks for taking it.

PR #906, opened 2022-07-22
Revert "Revert "Use zypper for openSUSE.""

marxin:
Based on the recent openSUSE discussion, we are going to stick with `zypper` and thus I would like to use it for openSUSE.

This reverts commit 846c63d20ffa9ea37ae9fc245636b736a11e2383.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,11 +16,11 @@ jobs:
         container:
           - 'registry.fedoraproject.org/fedora:latest'
           - 'registry.fedoraproject.org/fedora:rawhide'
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
What openSUSE discussion? And I'd prefer to keep using DNF for rpmlint CI on openSUSE.

marxin:
https://jira.suse.com/browse/PED-163 (I don't know if you have access to it or not).

Anyway, openSUSE officially supports `zypper`, so I want to use it.

Conan-Kudo:
I'm not a SUSE employee, so I don't.

cc: @lkocman

marxin:
We're staying with zypper cli as mentioned here: https://code.opensuse.org/leap/features/issue/1

PR #907, opened 2022-07-24
Port TOML configs to tomli library

marxin:
Apparently the tomli library (part of Python 3.11 std library),
does not support when '{' and '}' are placed to a separate line:

```
digests = [
  {
    key = "value"
  }
]
```

it only support when { } are on the same line:

`  { key = "value" }`

Thus, I support more TOML-like (and not JSON-like format) format
to be used.


marxin:
Btw. I raised a question about it: https://github.com/hukkin/tomli/issues/199

marxin:
All right, so what we have now is not a valid TOML by what spec says:

Quoting https://toml.io/en/v1.0.0#inline-table

> No newlines are allowed between the curly braces unless they are valid within a value.

Thus, I'm going to merge this.

configs/openSUSE/cron-whitelist.toml
```diff
@@ -3,207 +3,155 @@ package = "rpmlint-integration-test"
...
-    {
-        path = "/etc/cron.daily/logdigest",
-        hash = "1dcd29ceaa35914c126bf76b0b00b955de332744636573dc2b907c1c542f81c4"
-    },
-    {
-        path = "/etc/cron.d/logdigest-all-ignores-autogeneration",
-        hash = "052d8f93980ade99937c2a4b972d5cf5a77367c9af6c7b02f0e05e8ca5b06797"
-    },
-]
+[[FileDigestGroup.digests]]
```
mgerstner, line 42: How exactly is the coupling between the logdigest package and the digest established? Only by way of the order in which the sections appear?

marxin: Please take a look at the specs here: https://toml.io/en/v1.0.0#table

mgerstner: Okay then to answer my own question by way of the spec:

```
Any reference to an array of tables points to the most recently defined table element of the array. This allows you to define sub-tables, and even sub-arrays of tables, inside the most recent table.
```

So my assumption is correct.

I find this TOML table syntax confusing though, not really intuitive.

marxin: Yeah, fully understand your concern, but it's what TOML is about.

PR #917, opened 2022-08-26
add missing-hash-section check for HASH ELF sections

marxin:
System V requires that .hash sections is present in shared libraries
for dynamic symbol resolution. And we can emit warning when .gnu.hash
section is missing as it rapidly speeds up the symbol resolution.


marxin:
Forgot to mention that it addresses https://www.phoronix.com/news/Glibc-2.36-EAC-Problems.

Issue #918, opened 2022-08-26
new check for man pages


DimStar77:
rpmlint should warn/err on things like this:

> /usr/share/man/man1/libcbor.3.gz

i.e. the check should validate that the directory `/usr/share/man/man<n>` only contains files matching `*.<n>%{?ext_man}` (ext_man is defined to the compression scheme used, e.g. `.gz`

DimStar77:
also: I don't think `/usr/share/man/man<n>` is supposed to contain further subdirectories; another trigger to yell at packagers

DimStar77:
Seems we were a bit too strict with the new checks

e.g it is very common for perl modules to have a file called
topic.3pm.gz in /usr/share/man/man3 - which seems to be valid

this implies the folder name `man<N>` should contain files in the form `*.<N>*.gz`

(I've seen other suffixed than pm)

jengelh:
Due to automake, it has been common on GNU/Linux to place .3pm into /usr/share/man/man3.
This is in conflict with the way things were traditionally done (e.g. Solaris) where the full suffix is used, e.g. .1m ../man1m rather than ../man1.
An inquiry on the automake mailing list is pending.

tmzullinger:
This appears to have caused a regression.  Running rpmlint-2.4.0 against https://kojipkgs.fedoraproject.org//packages/shigofumi/0.9/5.fc37/x86_64/shigofumi-0.9-5.fc37.x86_64.rpm results in:

```
shigofumi.x86_64: W: no-manual-page-for-binary shigofumi
```

The package contains a matching man page.  This works in 2.3.0 or in 2.4.0 if I revert `FilesCheck.py` to 718e1eb9^.

```
$ rpm -qplv shigofumi-0.9-5.fc37.x86_64.rpm | grep -E ' /usr/(bin|share/man)' | grep -w shigofumi
-rwxr-xr-x    1 root     root                   191240 Jul 23 08:38 /usr/bin/shigofumi
-rw-r--r--    1 root     root                     3941 Jul 23 08:38 /usr/share/man/cs/man1/shigofumi.1.gz
-rw-r--r--    1 root     root                     3372 Aug 18  2015 /usr/share/man/man1/shigofumi.1.gz
```

Before I go digging, is the cause obvious to anyone here who worked on this change?  My first thought is that the translated man page is what the code may be tripping over.

marxin:
[...]

PR #919, opened 2022-08-26
extend checks related to manual pages

marxin:
Fixes: #918


marxin:
@DimStar77 Can you please check the suggested names and can you please write check descriptions?
And how about the badness of the newly added errors, any suggestions?

PR #920, opened 2022-08-26
Run flake8 directly as pytest-flake8 is broken.

marxin:
It's broken due to:
https://github.com/tholo/pytest-flake8/issues/87


marxin:
TW will be fixed after https://build.opensuse.org/request/show/999458 gets merged.

PR #842, opened 2022-04-13
Add ergo user and group

tacerus:
Signed-off-by: Georg Pfuetzenreuter <mail@georg-pfuetzenreuter.net>

Hi,

this is for a package I am creating for https://github.com/ergochat/ergo.

Do I need to submit this to the `opensuse` branch separately?

Best,
Georg

PR #926, opened 2022-09-06
improve tools/generate-fedora-users-groups

tmzullinger:
The `uidgid` file in setup was changed to use '#' for comments in [c2d8d46 (uidgid: simplify table format by properly commenting lines, 2022-05-30)](https://pagure.io/setup/c/c2d8d46), released in 2.14.1.

Before then, a few stray comment lines were added in [a2b68d4 (remove uidgid reservation for systemd-journal-gateway (#1174304), 2014-12-18)](https://pagure.io/setup/c/a2b68d4) and [6563bed (Stop creating katello user, 2021-06-16)](https://pagure.io/setup/c/6563bed), released in setup-2.9.3 and setup-2.13.9, respectively.

Skip lines beginning with a comment marker when parsing the uidgid file.

Next, update the script from 5ee8dc83 (extract script generating Fedora users and groups to tools/, 2021-03-24) to produce output suitable for writing directly to configs/Fedora/users-groups.toml.  And sort the user and group lists to provide more stable output.

I have not synced the script in this series yet.  Doing so with setup-2.14.1 from Fedora would remove three users &groups:

    * #systemd-journal-gateway
    * ergo
    * katello

The first is a mistake which is due to not excluding comment lines in the current uidgid file.  It was removed upstream in [a2b68d4 (remove uidgid reservation for systemd-journal-gateway (#1174304), 2014-12-18)](https://pagure.io/setup/c/a2b68d4).  The last was removed upstream in [6563bed (Stop creating katello user, 2021-06-16)](https://pagure.io/setup/c/6563bed).

That leaves only `ergo` which was added directly to rpmlint in #842.  I don't know what the policy is, but I would think that rpmlint should ship a Fedora config which matches what `setup` provides for users & groups.  That would mean getting `ergo` into `setup` upstream before syncing, most likely.


marxin:
Thanks for the contribution.

> That leaves only `ergo` which was added directly to rpmlint in #842. I don't know what the policy is, but I would think that rpmlint should ship a Fedora config which matches what `setup` provides for users & groups. That would mean getting `ergo` into `setup` upstream before syncing, most likely.

Yes, please add `ergo` to `setup` and then re-run `tools/generate-fedora-users-groups.py`.

tmzullinger:
Thanks Martin.

Looking at `ergo` closer, I think it should be removed and not submitted to `setup`.  The user was added for a different ergo than is in Fedora.

The Fedora [ergo](https://src.fedoraproject.org/rpms/ergo) package is a "program for large-scale self-consistent field calculations."  The [ergo](https://github.com/ergochat/ergo) package which was referenced in #842 appears to be an IRC daemon written in Go, which is not in Fedora AFAICT.

There is a build of the latter in the openSUSE Build Service [here](https://build.opensuse.org/package/show/home:crameleon/ergo).  That package doesn't appear to hard-code a uid/gid, so I don't think it belongs in the rpmlint users/groups list.

I'll submit a change to sync the users/groups from the current setup package and tag the author of #842 and hopefully we can make sure it works for everyone.

marxin:
> I'll submit a change to sync the users/groups from the current setup package and tag the author of #842 and hopefully we can make sure it works for everyone.

That would be the best. Appreciate the analysis/archeology work you've done so far ;) Note the OBS `ergo` package hasn't reached `openSUSE:Factory` so that it's a pet package so far.

PR #927, opened 2022-09-07
configs/Fedora: sync user/groups with setup-2.14.1

tmzullinger:
This removes 3 users/groups:

    * #systemd-journal-gateway
    * ergo
    * katello

The `#systemd-journal-gateway` entry was due to not excluding comment lines in the current uidgid file.  It was removed upstream in a2b68d4 (remove uidgid reservation for systemd-journal-gateway (#1174304), 2014-12-18)¹.

The `ergo` entry was added in ab36f34e (Add ergo user and group, 2022-04-13)².  However, the Fedora ergo⁴ package is a "program for large-scale self-consistent field calculations."  The ergo⁵ package which was referenced in #842 is an IRC daemon written in Go, which is not in Fedora.

The `katello` entry was removed upstream in 6563bed (Stop creating katello user, 2021-06-16)³.

This was done with tools/generate-fedora-users-groups.py:

    tools/generate-fedora-users-groups.py >configs/Fedora/users-groups.toml

¹ https://pagure.io/setup/c/a2b68d4
² https://github.com/rpm-software-management/rpmlint/commit/ab36f34e
  https://github.com/rpm-software-management/rpmlint/pull/842
³ https://pagure.io/setup/c/6563bed
⁴ https://src.fedoraproject.org/rpms/ergo
⁵ https://github.com/ergochat/ergo


tmzullinger:
@tacerus am I correct about the ergo package you meant in #842?  If so, do you see any issue with removing it from the Fedora users/groups lists?

This is a follow-up to #926.

tacerus:
Hi @tmzullinger, thanks for checking in! I am still building `ergo` (the IRC server) in [OBS](https://build.opensuse.org/package/show/home:crameleon/ergo) for personal use on Fedora, but given my main focus is openSUSE I am unlikely to submit it to the Fedora repositories soon. So whilst it would be cool to keep it to avoid rpmlint complaining over there, I understand it's not needed. :-)

tmzullinger:
Thanks @tacerus.  If you do need to ignore the `non-standard-uid` warning, I think an `addFilter()` entry in a local rpmlintrc is a good way to handle it.  Something like:

```
addFilter(r'ergo\..*: W: non-standard-uid /path/owned/by/ergo ergo')
```

Packages in Fedora which don't use one of the users/groups defined in the `uidgid` file from `setup` typically do that.  The advantage is that the exclusion only applies to the package which creates the user/group.  In Fedora's dist-git, that would be named ergo.rpmlintrc.  I don't know how such rpmlint config files are handled by OBS, but I presume they have some method.

tacerus:
Thank you very much for the tip and snippet! Yes, we can handle rpmlintrc files the same way - I will add one with a clause for Fedora then.

PR #929, opened 2022-09-13
Whitelist /var/spool/mail in postfix, sendmail and exim (bsc#1179574)

lnussel:
This is just moving a directory from the filesystem package where it already is whitelisted to the actual users of the directory.

### DIFF
--- a/configs/openSUSE/world-writable-whitelist.toml
+++ b/configs/openSUSE/world-writable-whitelist.toml
@@ -39,6 +39,36 @@ owner = "root"
 group = "root"
 
 [[WorldWritableWhitelist]]
+package = "postfix"
+bug = "bsc#1179574"
+note = "Public standard sticky-bit directories"
+[[WorldWritableWhitelist.files]]
+path = '/var/spool/mail'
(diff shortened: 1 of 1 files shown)


configs/openSUSE/world-writable-whitelist.toml
```diff
@@ -46,4 +76,4 @@ note = "nss caching daemon socket. is packaged as %ghost, therefore 'appears' to
 path = '/run/nscd/socket'
 mode = "-rw-rw-rw-"
 owner = "root"
-group = "root"
\ No newline at end of file
+group = "root"
```
marxin, line 79: Please remove this change.

marxin (changes requested):

marxin (approved):

mgerstner:
If the directory is no longer in the file system package can you then please also remove the old entry?

lnussel:
it is still there. I can remove it only when the other packages are in

mgerstner:
Okay, a cleanup PR# once it is possible would be nice. Thanks!

mgerstner (approved):

PR #931, opened 2022-09-13
dbus-services: autofs: move from /etc to /usr

wfrisch:
bsc#1203362

PR #933, opened 2022-09-15
dbus-services: autofs: temporarily accept old path (bsc#1203362)

wfrisch:
dbus-services: autofs: temporarily accept old path (bsc#1203362)

https://bugzilla.suse.com/show_bug.cgi?id=1203362
https://github.com/rpm-software-management/rpmlint/pull/931#issuecomment-1247791018

Due to a conflict in openSUSE Factory staging, I'd like to whitelist the old path temporarily, set a reminder for next week and remove it once autofs is updated.

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -467,9 +467,10 @@ package = "autofs"
 type = "dbus"
 note = "imported from rpmlint1 DBUSServices.WhiteList"
 bugs = ["bsc#782691", "bsc#1203362"]
-nodigests = [
-    "/usr/share/dbus-1/system.d/org.freedesktop.AutoMount.conf",
-]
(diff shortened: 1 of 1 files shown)


mgerstner:
Could you also add a digest based whitelisting for the new path when it's done? We should continuously switch to them as we touch entries.

wfrisch:
> Could you also add a digest based whitelisting for the new path when it's done? We should continuously switch to them as we touch entries.

OK, will do.

marxin:
Is this still necessary as we've got https://build.opensuse.org/request/show/1003880 ?

wfrisch:
> Is this still necessary as we've got https://build.opensuse.org/request/show/1003880 ?

Technically not, as this request is likely to be accepted.
We should merge it nevertheless, because I added the file digest, as suggested by @mgerstner

PR #935, opened 2022-09-28
Whitelist /var/spool/mail in postfix-bdb (bsc#1179574)

varkoly:
Visit https://build.opensuse.org/request/show/1002629

### DIFF
--- a/configs/openSUSE/world-writable-whitelist.toml
+++ b/configs/openSUSE/world-writable-whitelist.toml
@@ -39,6 +39,16 @@ owner = "root"
 group = "root"
 
 [[WorldWritableWhitelist]]
+package = "postfix-bdb"
+bug = "bsc#1179574"
+note = "Public standard sticky-bit directories"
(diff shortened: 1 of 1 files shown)


configs/openSUSE/world-writable-whitelist.toml
```diff
@@ -76,4 +86,4 @@ note = "nss caching daemon socket. is packaged as %ghost, therefore 'appears' to
 path = '/run/nscd/socket'
 mode = "-rw-rw-rw-"
 owner = "root"
-group = "root"
\ No newline at end of file
+group = "root"
```
FilippoBonazziSUSE, line 89: Please drop this hunk

FilippoBonazziSUSE (changes requested):

marxin:
Patch applied with the removed hunk as 5ca9841b8fbbed962c209c67d400e47a2a5640f9.

PR #942, opened 2022-10-05
Move autosetup regex into config options file

sergey-239:
There is a `forgemeta` macro in Fedora which provides `forgeautosetup` macro with the same functionality as `autosetup`, so we would like `autosetup` regex pattern to be customizable.

### DIFF
--- a/rpmlint/checks/SpecCheck.py
+++ b/rpmlint/checks/SpecCheck.py
@@ -241,7 +241,7 @@ class SpecCheck(AbstractCheck):
                 if current_section != 'prep':
                     self.output.add_info('W', pkg, '%autopatch-not-in-prep')
             else:
-                res = autosetup_regex.search(line)
+                res = self.autosetup_regex.search(line)
                 if res:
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
The forge macros are unmaintained in Fedora, so unless someone is stepping up to maintain them and mainline them, I don't think we should do anything.

sergey-239:
> 
Neal, but forgemeta is quite in use: there are 231 spec out of ~22k in FC36 have it. As well it is mentioned in packaging guidelines. How that comes out?

Conan-Kudo:
> Neal, but forgemeta is quite in use: there are 231 spec out of ~22k in FC36 have it. As well it is mentioned in packaging guidelines. How that comes out?

It was added to the guidelines when @nim-nim wrote the macros and made the Go macros dependent on them. They haven't seen significant wider adoption beyond that. When he left Fedora, the macros basically became unmaintained because nobody else knows how to work with that code.

marxin:
Thus, I'm closing this PR for now.

PR #944, opened 2022-10-12
iio-sensor-proxy: move dbus system.d file to /usr (bsc#1204055)

FilippoBonazziSUSE:


mgerstner:
same issue here as in the other two whitelisting adjustments

FilippoBonazziSUSE:
I don't get it - I got the digest value from OBS in the rpmlint failure log. Is that wrong?

FilippoBonazziSUSE:
Oh, I see. I guess that does not apply to the `shell` digest, and that's how I was able to get a match to an existing file using `sha256` in #945 .

marxin:
`shell` digest also ignores whitespaces.

You can see if there's a diff by running something like `./tools/get_whitelisting_digest.py /etc/dbus-1/system.d/net.hadess.SensorProxy.conf --filter=xml --show-result > after && diff -u /etc/dbus-1/system.d/net.hadess.SensorProxy.conf after`

FilippoBonazziSUSE:
I see. I just got lucky then... Will fix

mgerstner (approved):

PR #945, opened 2022-10-12
geoclue2: move dbus system.d file to /usr (bsc#1204054)

FilippoBonazziSUSE:


mgerstner:
These checksums also seems to come from `sha256sum`, not from `get_whitelisting_digest.py --filter {xml,shell}`

FilippoBonazziSUSE:
I obtained all the digests I could from the rpmlint failure log. For the value that wasn't there (`"/usr/share/dbus-1/system-services/org.freedesktop.GeoClue2.service"`) I did get it via `sha256sum`, but I checked with some other file that already had a digest first... So I generated the digest of a known file and compared the output of `sha256` to the one already in `configs/openSUSE/dbus-services.toml`. They matched, so I didn't think any further.

TBF I didn't even know about `get_whitelisting_digest.py`. Let me check everything

mgerstner:
> I obtained all the digests I could from the rpmlint failure log.

Yeah that is a bit unfortunate, I thought about doing some heuristics in the rpmlint check to output the digest for the correct filter. For a human the pattern it is pretty clear, for the machine it's a bit more intricate than it looks at first.

Maybe we can just enhance the check to simply output all possible checksums to pick from.

mgerstner (approved):

Issue #940, opened 2022-10-05
"rpmdiff -i S old new" is incompatible with rpmlint-2.x


tmzullinger:
Downstream in Fedora, we received [this bug report](https://bugzilla.redhat.com/1969534) (some time back, sorry for not forwarding it sooner).  It notes that syntax which used to work with rpmlint-1.x is broken with rpmlint-2.x:

```
$ rpmdiff -i S old.rpm new.rpm
usage: rpmdiff [-h] [-V] [-i {S,M,5,D,N,L,V,U,G,F,T} [{S,M,5,D,N,L,V,U,G,F,T} ...]] [-e GLOB [GLOB ...]] RPM_ORIG RPM_NEW
rpmdiff: error: argument -i/--ignore: invalid choice: 'old.rpm' (choose from 'S', 'M', '5', 'D', 'N', 'L', 'V', 'U', 'G', 'F', 'T')
```

Workarounds included separating the option values and positional arguments with `--` or listing the positional parameters before the options.  The handling of multiple values is also different than rpmdiff-1.x, as that required `-i S -i T` while with rpmdiff-2.x, the last `-i` option wins and all others are _silently_ ignored.  Users must use `-i S T` instead.  That's arguably a better syntax, but it is a mildly surprising difference at first.

Unfortunately, I'm not sure that `argparse` is capable of handling this well without adding a custom parser or something similar.

We can at least document the usage better though, which I'll do in a PR shortly.  But I'd be happy to hear of better options to make this less of a footgun for users.

marxin:
Thanks for the report.
Yes, we define `-i` with `nargs='+'` ([^1]) and one can't really distinguish between the argument values and `RPM_ORIG RPM_NEW` other than with `--`. So thanks for the documentation improvement.

[^1]: https://docs.python.org/3/library/argparse.html#nargs

marxin:
Can we please close this as we can't do much more?

tmzullinger:
Yeah, no reason to hold it open.  I did notice just the other day that my documentation change needlessly mentions multiple values.  I'll submit the following to fix that:

```diff
diff --git i/rpmlint/cli.py w/rpmlint/cli.py
index 62bc2127..a363602e 100644
--- i/rpmlint/cli.py
+++ w/rpmlint/cli.py
@@ -37,7 +37,7 @@ def process_diff_args(argv):
 
     parser = argparse.ArgumentParser(prog='rpmdiff',
[...]

PR #949, opened 2022-10-14
rpmdiff: clarify help text on -i/-e options

tmzullinger:
The epilog added in f551541a (rpmdiff: improve help for multiple values with -i/-e options, 2022-10-04) mentions multiple values for the -i/-e options.  However, the issue is present regardless of the number of values.  Remove the reference to multiple values to simplify the help text.

Related: #940


marxin:
Thanks!

PR #950, opened 2022-10-17
SpecCheck: Split big line by line for-loop in methods

danigm:
This patch makes the code a bit more readable, breaking the big for-loop that check the spec file line by line in a series of methods in the SpecCheck class.

With this change it's easier to extend the line-by-line check with small functions and the if-else-if-... complexity is reduced a bit.

See #225


marxin (changes requested):
I really appreciate the pull request, please fix the one comment I have and I'm going to merge it.

marxin:
Thank you!

PR #951, opened 2022-10-18
FilesCheck: Split big check method

danigm:
This patch makes the code a bit more readable, breaking the big check method in a serie of methods in the FilesCheck class.

See #225


marxin:
Please address the flake8 issues, apart from that, I'm fine with the PR.

PR #954, opened 2022-10-25
Add python-module-def check to SpecCheck

danigm:
This patch adds a new warning when detecting the use of python_module macro definition in the spec file.

It's included by default in python-rpm-macros so every modern distro should have this module defined.

This macro was added by py2pack by default in new packages, but now it's removed:
https://github.com/openSUSE/py2pack/pull/166


marxin (changes requested):
Please document the new warning in `rpmlint/descriptions/SpecCheck.toml`.

PR #955, opened 2022-10-25
Add python check

danigm:
We can recover this PR #430 , to fix the issue #423


marxin:
Adding stakeholders from the previous PR: @jgrassler, @mcepl and @hroncok.

mcepl (approved):
LGTM

marxin:
About the newly added RPM file: How did you create them? We might want to collect them in https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests, what do you think?

marxin:
> I don't know how these packages were generated, but maybe I can recreate them, so we can keep track of the source spec files.

They were created manually. So please create new dummy packages and commit there the corresponding spec files. Thanks!

rpmlint/checks/PythonCheck.py
```diff
@@ -0,0 +1,64 @@
...
+
+        for path in WARN_PATHS:
+            path_re = re.compile(path)
+            key = WARN_PATHS[path]
+
+            if path_re.match(filename):
+                if key == 'tests':
+                    # Ignore "-test" and "-tests" packages since these are
+                    # supposed to contain tests.
+                    if pkg.name.endswith('test') or pkg.name.endswith('tests'):
```
marxin, line 46: Please use `pkg.name.endswith(('test', 'tests'))`

rpmlint/checks/PythonCheck.py
```diff
@@ -0,0 +1,64 @@
...
+            if path_re.match(filename):
+                self.output.add_info('E', pkg, ERRS[key], filename)
+
+    def check_egginfo(self, pkg, filename):
+        """
+        Check type of egg-info metadata and check Requires against egg-info
+        metadata if applicable.
+        """
+        # Check for (deprecated) distutils style metadata.
+        if os.path.isfile(os.path.join(pkg.dir_name(), filename.lstrip('/'))):
```
marxin, line 63: please use `Path` class from `pathlib` (if possible)

marxin (changes requested):

jgrassler:
LGTM (caveat: I only looked; did not test).

> They were created manually. So please create new dummy packages and commit there the corresponding spec files. Thanks!

Indeed they were. I'm not going to step up for this myself, since I no longer work for SUSE and no longer have a proper development environment for building and testing RPM packages. That being said, feel free to use the spec I built them from back then: https://build.opensuse.org/package/view_file/devel:openSUSE:Factory:rpmlint:tests/pythoncheck/pythoncheck.spec?expand=1

rpmlint/checks/PythonCheck.py
```diff
@@ -0,0 +1,64 @@
+import os
+import re
+
+from rpmlint.checks.AbstractCheck import AbstractFilesCheck
+
+# Warning messages
+WARNS = {'tests': 'python-tests-in-package',
```
mcepl, line 7: Actually, I have to correct myself. After some discussion with my colleague, I agree that these should be ERR. Yes, I can see @hroncok ’s [comment](https://github.com/rpm-software-management/rpmlint/pull/430#discussion_r428141147), but I think these packages are truly exceptional and that’s what `rpmlintrc` is for. On the other hand the proliferation of warnings make people just ignore the output of `rpmlint` more and more, it is not healthy. Warning should be just for warnings not for our lack of guts to decide.

mcepl (changes requested):

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
+python-tests-in-package="""
+test/ or tests/ directory in package directory. Tests in %{python_sitelib}/<pkgname> should be packaged in a separate test or tests subpackage."""
+python-doc-in-package="""
+doc/ or docs/ directory in package directory. Documentation should go into %{docdir}, not %{python_sitelib}/<pkgname>"""
+python-egg-info-distutils-style="""
+The package's egg-info is a distutils style file, rather than a setuptools
+style directory. Please convert to setuptools style.
```
hroncok, line 7: The correct message we want to spread is to update to dist-info standardized core metadata rather than recommending the replacement of deprecated stuff with a bit less deprecated stuff.

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
...
+"""
+python-tests-in-site-packages="""
+test/ or tests/ directory in %{python_sitelib}. This should never happen since
```
hroncok, line 10: Should this also work for python_sitearch? Same for the next ones.

rpmlint/checks/PythonCheck.py
hroncok: I don't really care whether this is a warning or an error.

hroncok (commented):
Not looking at the behavior or code yet, just commenting on the new errors/warnings.

rpmlint/checks/PythonCheck.py
```diff
@@ -59,6 +63,9 @@ def check_egginfo(self, pkg, filename):
         Check type of egg-info metadata and check Requires against egg-info
         metadata if applicable.
         """
+
+        filepath = Path(pkg.dir_name())
+        filepath = filepath.joinpath(filename.lstrip('/'))
```
marxin, line 68: `filepath = Path(pkg.dir_name(), filename.lstrip('/'))` please

marxin: Even better: https://github.com/rpm-software-management/rpmlint/blob/a2acb99e8a2dfe16fb18d0c3d880620cddd2f59d/rpmlint/pkg.py#L523

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
+python-tests-in-package="""
+test/ or tests/ directory in package directory. Tests in %{python_sitelib}/<pkgname> should be packaged in a separate test or tests subpackage."""
+python-doc-in-package="""
+doc/ or docs/ directory in package directory. Documentation should go into %{docdir}, not %{python_sitelib}/<pkgname>"""
```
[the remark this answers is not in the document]
s-t-e-v-e-n-k: Personally, I think site-packages is the wrong place -- it isn't easily discoverable, and furthermore, it leads to uninstallability bugs when two packages can not be co-installed because they happen to make a doc or docs module under site-packages.

marxin (approved):

marxin:
I'm fine with the implementation side of the pull request. Once the discussion with the Python stakeholders settles down, I'm ready to merge it.

rpmlint/descriptions/PythonCheck.toml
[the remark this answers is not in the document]
hroncok: Yes, ownign side-packages/doc(s) is bad, but not because it's not in %{docdir} but because it clashes with others.

rpmlint/checks/PythonCheck.py
hroncok: I've just pointed out in the comment you link that if it has `__init__.py` it should not produce the warning/error.

mcepl: Except, @danigm , I cannot find the code checking for the presence of `__init__.py`, did I miss it?

hroncok: I don't think it's there either.

danigm: I've just added a new commit to do the check for `__init__.py` file, so python modules called `doc` won't raise the warning: https://github.com/rpm-software-management/rpmlint/pull/955/commits

marxin:
@danigm: Please mark as resolved all the conversions that are resolved.

Generally speaking, what is remaining to be discussed before we can merge this? Is there something?

hroncok:
> Generally speaking, what is remaining to be discussed before we can merge this? Is there something?

I'd like to give it a try and see if all the checks make sense to me, but I'm on PyCon PL until Sunday, so won't be able to do that this week.

marxin:
> I'd like to give it a try and see if all the checks make sense to me, but I'm on PyCon PL until Sunday, so won't be able to do that this week.

Sure, don't hurry, we can finish it this week.

hroncok:
On it.

rpmlint/checks/PythonCheck.py
```diff
@@ -0,0 +1,76 @@
...
+
+SITELIB_RE = '/usr/lib[^/]*/python[^/]*/site-packages'
+
+# Paths that shouldn't be in any packages, ever, because they clobber global
+# name space.
+ERR_PATHS = [
+    (re.compile(f'{SITELIB_RE}/tests?$'), 'tests'),
+    (re.compile(f'{SITELIB_RE}/docs?$'), 'doc'),
+    (re.compile(f'{SITELIB_RE}/src$'), 'src'),
+    (re.compile(f'{SITELIB_RE}/[^/]+/tests?$'), 'tests-package'),
```
hroncok, line 28: So site-packages/foo/tests is an error, but site-packages/foo/bar/tests is fine?

```pycon
>>> import re
>>> SITELIB_RE = '/usr/lib[^/]*/python[^/]*/site-packages'
>>> r = re.compile(f'{SITELIB_RE}/[^/]+/tests?$')
>>> r.match('/usr/lib/python3.11/site-packages/foo/tests')
<re.Match object; span=(0, 43), match='/usr/lib/python3.11/site-packages/foo/tests'>
>>> r.match('/usr/lib/python3.11/site-packages/foo/bar/tests')
```

That does not sound like reasonable behavior to me.

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
+python-tests-in-package="""
+test/ or tests/ directory in package directory. Tests in %{python_sitelib}/<pkgname> should be packaged in a separate test or tests subpackage or not packaged."""
```
hroncok, line 2:
```suggestion
test/ or tests/ directory in Python package directory. Tests in %{python_sitelib}/<pkgname> should be packaged in a separate test or tests subpackage or not packaged."""
```

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
...
+python-doc-in-package="""
+doc/ or docs/ directory in package directory. Documentation should go into %{docdir}, not %{python_sitelib}/<pkgname>"""
```
hroncok, line 4:
```suggestion
doc/ or docs/ directory in Python package directory. Documentation should go into %{docdir}, not %{python_sitelib}/<pkgname>"""
```

rpmlint/descriptions/PythonCheck.toml
```diff
@@ -0,0 +1,20 @@
...
+python-egg-info-distutils-style="""
+The package's egg-info is a distutils style file.
```
hroncok, line 6:
```suggestion
The Python package's egg-info is a distutils style file.
```

rpmlint/checks/PythonCheck.py
danigm: Good point. This is to detect a common pattern, that usually have tests in the package module, but it's true that tests can be found in any subpackage so I think that it shouldn't be difficult to find python packages that have this kind of folders/subfolders, so maybe in this case it's better to look for all the tests folders:

```python
    (re.compile(f'{SITELIB_RE}/.+/tests?$'), 'tests-package'),
```

hroncok: BTW this is my only remaining concern.

hroncok: This does not seem resolved.

mcepl:
@danigm Congratulations!

hroncok:
https://github.com/rpm-software-management/rpmlint/pull/955#discussion_r1022681043 was not resolved

bnavigator:
Sorry, I am late to the party. IMHO `python-tests-in-package` needs to be revised. Packaging tests as submodule is a common pattern and brought out as [valid and useful upstream](https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#tests-as-part-of-application-code).

This rpmlint error has bitten me twice today and it is unnecessary burden on the packager.

The check should only test for toplevel test modules.

mcepl:
> Sorry, I am late to the party. IMHO `python-tests-in-package` needs to be revised. Packaging tests as submodule is a common pattern and brought out as [valid and useful upstream](https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#tests-as-part-of-application-code).

1. Please, file a new ticket, it probably doesn’t make sense to continue in this PR.
2. This is about *installation* of `package/tests` in `%{python_sitelib}`. Is this really so widely done, that it cannot be just overridden with `*.rpmlintrc`? We generally don’t want to install test suite into our binary RPM packages.

Issue #833, opened 2022-03-24
Do not consider files-duplicate for SMALL files


socketpair:
In Python RPMS:

```
$ xxd usr/lib/python3.10/site-packages/pyrad-2.4-py3.10.egg-info/zip-safe
00000000: 0a                                       .
$ xxd usr/lib/python3.10/site-packages/pyrad-2.4-py3.10.egg-info/dependency_links.txt
00000000: 0a
```
These two files both contain just one empty newline. It's not a bug to have two such files.

marxin:
That seems reasonable to me. We can add a config value where one can define a minimum file size that should be reported as `files-duplicate`.

PR #959, opened 2022-11-09
Do not consider files-duplicate for SMALL files

danigm:
Adds an option to do not check for duplicates files that are smaller than a certain size.

It could be possible to set to 4 by default, but that will [break one test](https://github.com/rpm-software-management/rpmlint/blob/main/test/test_duplicates.py#L16), and I don't know where are the sources of that binary package to generate it with bigger duplicate files.

Fix https://github.com/rpm-software-management/rpmlint/issues/833


marxin:
Apparently, we don't have sources for the package:
https://build.opensuse.org/project/show/devel:openSUSE:Factory:rpmlint:tests

I do support the patch, so please adjust the problematic test case and I'm ready to merge it.

PR #961, opened 2022-11-15
PythonCheck: Do not allow tests folder in subpackages

danigm:
See https://github.com/rpm-software-management/rpmlint/pull/955#discussion_r1022681043


hroncok:
Thanks

Issue #969, opened 2022-11-30
shared-library-without-dependency-information false positives


jengelh:
Given this shared library:

```
$ cat x.c
int compute_bla(int x) { return x * x * x; }
$ gcc -c x.c -Wall -O2
$ gcc -shared -o libx.so x.o -Wl,--as-needed
$ ldd libx.so
        statically linked
```

rpmlint complains like so

```
gap-datastructures.aarch64: E: shared-library-without-dependency-information /usr/lib64/gap/pkg/datastructures-0.3.0/bin/aarch64-suse-linux-gnu-default64-kv8/datastructures.so                                                       
```

openSUSE uses -Wl,--as-needed, and I seem to remember some other distro doing it too. So it's not like the compilation step has willingly used the -nostdlib that rpmlint was after.

marxin:
Confirmed, however, I'm curious how can we distinguish in between a library that was really statically linked against `glibc` and one that doesn't use any symbol from it?

jengelh:
>distinguish in between a library that was really statically linked against glibc

You could look for libc symbols in the shared object. But that would depend on being able to create such a monster… and in general, it does not appear like you can:

* amd64 is picky about PIC. Trying to link libc.a into a shared object won't work because of the insufficient reloc type:

  ```
  $ gcc -fPIC -shared rand_wrap.c -Wl,--as-needed,-Bstatic,-lc,-Bdynamic 
  /usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: /usr/lib64/gcc/x86_64-suse-linux/12/../../../../lib64/libc.a(malloc.o): relocation R_X86_64_TPOFF32 against `tcache' can not be used when making a shared object; recompile with -fPIC
  /usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: failed to set dynamic section sizes: bad value
  collect2: error: ld returned 1 exit status
  ```

  openSUSE has no `libc_pic.a`, so that's that.

* On i586, where PIC requirements are nil, this is still difficult:

  ```
[...]

PR #970, opened 2022-12-01
Remove shared-library-without-dependency and library-not-linked-against-libc

jengelh:
Libraries that do not use any libc symbols are legit (e.g. libs with just mathematical computations or lookup tables only). libc.so.6 (and sometimes libgcc_s.so.1) used to be linked into all shared libraries because gcc implicitly adds -lc -lgcc_s to *everything*. But ever since distributions started using --as-needed, -lc/-lgcc_s too are subject to being thrown at out at the linker stage.

References: #969


rpmlint/checks/BinariesCheck.py
```diff
@@ -444,11 +444,9 @@ def _check_library_dependency(self, pkg, pkgfile):
         dyn_section = self.readelf_parser.dynamic_section_info
         if not len(dyn_section.needed) and not (dyn_section.soname and
                                                 self.ldso_soname_regex.search(dyn_section.soname)):
-            if self.is_shobj:
-                msg = 'shared-library-without-dependency-information'
-            else:
+            if not self.is_shobj:
                 msg = 'statically-linked-binary'
-            self.output.add_info('E', pkg, msg, pkgfile.name)
+                self.output.add_info('E', pkg, msg, pkgfile.name)
```
marxin, line 449: Please use the `msg` string literal directly in the `add_info`.

rpmlint/checks/BinariesCheck.py
```diff
@@ -459,10 +457,8 @@ def _check_library_dependency(self, pkg, pkgfile):
                     if 'libc.' in lib:
                         return
                 if self.is_shobj:
```
marxin, line 459: Here you miss `not` in `if self.is_shobj`, am I right?

marxin (changes requested):

PR #976, opened 2023-01-02
Franklin Street address is ok

bkmgit:
Franklin Street Address is used in:
 https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt 
and is the current address listed at:
 https://www.fsf.org/about/contact/


bkmgit:
One might provide a warning that the latest GPL license is not being used if the Franklin St. address is found. Old licenses at https://www.gnu.org/licenses/old-licenses/ have Franklin St. Address.

marxin:
Thank you for your contribution.

marxin:
> One might provide a warning that the latest GPL license is not being used if the Franklin St. address is found.

Feel free to extend the existing check if you want.

Issue #73, opened 2016-04-27
Suppress only-non-binary-in-usr-lib if there are symlinks to binaries in lib dir


scop:
For example unversioned `*.so` links to versioned binary `*.so.x.y` which are common in devel packages. Either just do this for all such symlinks, or actually examine the target of the symlink to see if it's binary. The latter is cumbersome and would only work for installed packages, while the former might be a bit broad. See https://bugzilla.redhat.com/show_bug.cgi?id=1330804

StefanBruens:
This issue is still present with current rpmlint, though the message has slightly changed.

I think the severity of this issue should be raised, as:  
1. This is an error now
2. It happens for every package which has a devel package

Example rpmplint output:
```
gpsnmea-devel.x86_64: E: no-binary
quat-devel.x86_64: E: no-binary
The package should be of the noarch architecture because it doesn't contain
any binaries.
```

StefanBruens:
The relevant code is here:

https://github.com/rpm-software-management/rpmlint/blob/3c828527f0e1435dd3779d5e47d7be5698929511/rpmlint/checks/BinariesCheck.py#L634-L635

I think the `ISREG` is wrong for two reasons:  
1. It triggers the warning here
2. It fails to detect any directory or symlink in `/usr/lib64` and allows it in a noarch package.

Installing such a noarch package would install stuff in `/usr/lib64` even on a 32 bit platform.

marxin:
I agree with your analysis and I've created a pull request. Feel free to comment there.

PR #983, opened 2023-01-06
Relax no-binary and noarch-with-lib64 checks

marxin:
Set `has_file_in_lib64` to True if a package contains a symlink to a file in that location. It's very a common pattern for shared libraries where foo-devel package provides only the symlink.

Fixes: #73

### DIFF
--- a/rpmlint/checks/BinariesCheck.py
+++ b/rpmlint/checks/BinariesCheck.py
@@ -631,9 +631,7 @@ class BinariesCheck(AbstractCheck):
 
             # find out if we have a file in /usr/lib64/ directory (needed later
             # for the package checks)
-            if stat.S_ISREG(pkgfile.mode) and \
-                    (fname.startswith('/usr/lib64') or
-                     fname.startswith('/lib64')):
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Is this actually relaxing it? It sounds like it does the opposite (makes the check stricter by covering more things...)

marxin:
> Is this actually relaxing it?

Yes, as it sets `has_file_in_lib64` in more cases and thus:
https://github.com/rpm-software-management/rpmlint/blob/1b2b958892da3a570c0737e0b86cfecdc4106a5e/rpmlint/checks/BinariesCheck.py#L207-L209

is relaxed (note right now it emits for `1526` packages in `openSUSE:Factory`). While the other consumer place:
https://github.com/rpm-software-management/rpmlint/blob/1b2b958892da3a570c0737e0b86cfecdc4106a5e/rpmlint/checks/BinariesCheck.py#L211-L217

is stricter, but the check triggers only once in `openSUSE:Factory`.

Does it make sense?

Conan-Kudo:
Yeah, I get it now.

Conan-Kudo (approved):

PR #994, opened 2023-01-19
openSUSE: new whitelisting restriction for sysctl drop-in files (bsc#1174722)

mgerstner:
Based on the file digest whitelisting check this introduces a new whitelisting restriction for installation of sysctl.d drop-in configuration files. The files whitelisted here have been reviewed.

Content based whitelistings seem feasible since there is only a small number of packages that ship files for this and the ones that do only rarely change in content.


wfrisch (approved):
Looks good to me.

Out of curiosity I checked out the sysctl source code and verified its behavior. procps (v3.x) and procps4 seem to behave identically in the relevant sections, e.g. config search paths.

wfrisch:
https://build.opensuse.org/request/show/1059842

mgerstner:
Thanks @wfrisch for reviewing!

DimStar77:
Breaks systemd: https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:E/systemd/standard/x86_64

marxin:
```
[  760s] systemd.x86_64: E: sysctl-file-symlink (Badness: 10000) /usr/lib/sysctl.d/99-sysctl.conf
[  760s] Symlinks are not allowed for sysctl.d drop-in configuration files
```

fbuihuu:
`/etc/sysctl.conf ` is not read by systemd-sysctl. Upstream recommended to work around this by creating a symlink pointing to sysctl.conf in /etc/sysctl.d or /usr/lib/sysctl.d.

SUSE currently uses /usr/lib/sysctl.d but I think /etc/sysctl.d makes more sense. I'll change it.

Could you whitelist /etc/sysctl.d/99-sysctl.conf ?

DimStar77:
> `/etc/sysctl.conf ` is not read by systemd-sysctl. Upstream recommended to work around this by creating a symlink pointing to sysctl.conf in /etc/sysctl.d or /usr/lib/sysctl.d.
> 
> SUSE currently uses /usr/lib/sysctl.d but I think /etc/sysctl.d makes more sense. I'll change it.
> 
> Could you whitelist /etc/sysctl.d/99-sysctl.conf ?

disagree - a package installing to /etc never makes more sense than installing to /usr - /etc is ADMIN space to configure the system

fbuihuu:
Well systemd package already ships a couple of default config files in /etc. Not really sure why it would be different for 99-sysctl.conf.

OTOH I don't see any drawback to ship the symlink in /usr so fine to let it in /usr. I was influenced by the recommendation made by upstream which is to put the symlink in /etc (not in /usr unlike what I stated in my previous comment).

mgerstner:
> `/etc/sysctl.conf ` is not read by systemd-sysctl. Upstream recommended to work around this by creating a symlink pointing to sysctl.conf in /etc/sysctl.d or /usr/lib/sysctl.d.

Hmm that is a strange corner. `sysctl` from procps *does* read `/etc/sysctl.conf`, this effectively means that regular `sysctl` will read and apply these settings twice.

The symlink is packaged in systemd but the file it's pointing to is packaged in `aaa_base`. Thus I can only whitelist the symlink as is, not verifying the content (the file is empty on a clean installation).

I will see how I can adjust this new check to fix that.

PR #995, opened 2023-01-24
temporarily drop badness for sysctl-file-* errors

marxin:
It breaks the current Staging projects as mentioned here: https://github.com/rpm-software-management/rpmlint/pull/994#issuecomment-1401471987

Related to: #994


marxin:
@DimStar77 needs a quick fix, that's why I'm going to merge this now.
@mgerstner

PR #997, opened 2023-01-25
opensuse: also exclude systemd-mini symlink from new sysctl.d restriction

mgerstner:
Along with this make it possible to share exception entries between multiple packages to avoid duplication.


FilippoBonazziSUSE (approved):
LGTM

PR #999, opened 2023-02-06
d-bus services: adjust NetworkManager-fortisslvpn from /etc to /usr

wfrisch:
Quick review: [bsc#1207585](https://bugzilla.suse.com/show_bug.cgi?id=1207585)


FilippoBonazziSUSE (approved):
LGTM

FilippoBonazziSUSE:
I just noticed that the commit title does not contain the bsc# reference... @wfrisch for next time this would be ideal, so that the `changesgenerate` OBS service can work best. In this case I fixed the changelog manually.

PR #1014, opened 2023-03-03
TagsCheck: handle license exception in grouping

tmzullinger:
A valid license exception which is in a grouping is reported as invalid. This is due to the trailing ')' being included in the regex.  The following license:

    (GPL-2.0-only OR GPL-3.0-only WITH Qt-GPL-exception-1.0) AND MIT

returns:

    W: invalid-license-exception Qt-GPL-exception-1.0)

Adjust the license_exception_regex to exclude trailing parenthesis.

Fixes: https://bugzilla.redhat.com/2175241


tmzullinger:
I'm not terribly familiar with writing tests for rpmlint, so please let me know if the added test should be adjusted.

danigm (approved):
Looks good to me

marxin:
Apart from that, I'm fine with the change.

tmzullinger:
Sure, I extracted it from the src.rpm which is part of the change.  I used an source instead of a binary rpm to ensure it could be easily reproduced and/or extracted.  :)

[valid-exception-in-grouping.spec.txt](https://github.com/rpm-software-management/rpmlint/files/10900131/valid-exception-in-grouping.spec.txt) (it has a .txt extension because GitHub is dumb.)

Thanks @danigm and @marxin!

marxin:
Thank you @tmzullinger!

Issue #1012, opened 2023-03-02
Move metadata from setup.py / setup.cfg to pyproject.toml


danigm:
In the [setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setuppy-discouraged) we can read:

> We also recommend users to expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the setup.py minimal with only the dynamic parts (or even omit it completely if applicable).

Right now we don't have anything special in the `setup.py`, just metadata, so it's possible to migrate all to the existing `pyproject.toml` file.

fabiobarkoski:
I will try to work on this issue. Do I need to assign myself?

danigm:
> Do I need to assign myself to work on this issue?

Just the comment that you're working on it is okay.

fabiobarkoski:
Ok, so I'm working on this issue.

fabiobarkoski:
It's better to reference the license like `license = {text = "GPL-2.0-or-later"}` or like `license = {file = "COPYING"}`?
And it's not a good idea to write on `README.md` to instead of installing the test dependencies separately, install like:
```bash
pip install "rpmlint[test]"
```
or
```bash
pip install -e ".[test]"
```
Since the `pyproject.toml` allow us to do this.

Conan-Kudo:
We haven't moved to `pyproject.toml` because it doesn't work properly on SUSE distributions. PEP-517 compliance for Python packaging doesn't work there.

fabiobarkoski:
According to the [setuptools docs](https://setuptools.pypa.io/en/latest/references/keywords.html), the `tests_require` is deprecated. Can the metadata be migrated to the `setup.cfg`? Another possibility is to use [tox](https://tox.wiki/en/latest/), even [nox](https://nox.thea.codes/en/stable/) or just use [`extras_require`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) instead.

danigm:
> We haven't moved to `pyproject.toml` because it doesn't work properly on SUSE distributions. PEP-517 compliance for Python packaging doesn't work there.

[...]

PR #1017, opened 2023-03-07
Setuptools deprecations

tmzullinger:
Minor fixes to silence `SetuptoolsDeprecationWarning`s.  These were noticed when building the Fedora packages against newer python releases.

* rename license_file to license_files

  In setuptools-56.0.0¹ license_file was changed to license_files.  The
  former causes a deprecation warning.

  ¹ https://github.com/pypa/setuptools/issues/2620
    https://setuptools.pypa.io/en/latest/history.html?highlight=license_files#v56-0-0

* include rpmlint.descriptions in packages

  Since setuptools-62.3.0², a warning is issued when setuptools finds an
  importable package is listed in package_data rather than in packages.

  ² https://github.com/pypa/setuptools/pull/3308
     https://setuptools.pypa.io/en/latest/history.html#v62-3-0

It feels a little strange to list `rpmlint.descriptions` in the `packages` section as well as `package_data`.  I don't know if there's a better way to do that or not.  (I did try moving the line as-is from `package_data` to `packages` but that wasn't valid.)

Perhaps at some point, reworking things to just use a toml file as the more modern setuptools describes might be good?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [metadata]
-license_file = COPYING
+license_files = COPYING
 
 [aliases]
 test=pytest
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ setup(
     packages=[
         'rpmlint',
         'rpmlint.checks',
+        'rpmlint.descriptions',
     ],
     package_data={
         'rpmlint': ['configdefaults.toml'],


danigm (approved):
Looks good. This is related to https://github.com/rpm-software-management/rpmlint/issues/1012

PR #1019, opened 2023-03-10
Migrate metadata from setup.py to pyproject.toml

fabiobarkoski:
Related to #1012. According to the recommendation of the [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setuppy-discouraged) the metadata was migrated from setup.py to pyproject.toml in order to maintain the setup.py with the minimal configuration.


danigm (commented):
Looks good to me. I've been testing and it generates almost the same wheel package. The only thing that differ is the License and the Readme inclusion in the metadata file.

About the license, we discussed this in in the [issue](https://github.com/rpm-software-management/rpmlint/issues/1012#issuecomment-1454732467), but looking at the generated package, maybe it's better to keep the license info as text and just add the `license-files` entry in the `tools.setuptools` group.

Conan-Kudo:
This raises the minimum version of setuptools though, doesn't it? I don't think pyproject data for setuptools has been around that long.

danigm:
> This raises the minimum version of setuptools though, doesn't it? I don't think pyproject data for setuptools has been around that long.

Yes, I didn't see that, maybe [v61.0.0](https://setuptools.pypa.io/en/latest/history.html#v61-0-0) is the minimum version required to have the pyproject.toml configuration file working.

Conan-Kudo:
Then a setup dependency needs to be declared.

danigm:
There's a conflict in `setup.py` with the latest changes in main, please rebase so we can continue pushing forward this change

PR #1021, opened 2023-03-15
No need to mention 'wheel'

afrid18:
As per PEP 517, there is no need of explicitly mentioning `wheel` in `install_requires` in pyproject.toml. [Reference](https://github.com/pypa/setuptools/commit/f7d30a95) [pypa/setuptools](https://github.com/pypa/setuptools) describes that "it will be installed by
PEP 517 front-ends automatically, when building wheels."

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
-    "setuptools>=39.2",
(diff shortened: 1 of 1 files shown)


danigm (approved):
Looks good to me.

PR #1028, opened 2023-03-20
TagsCheck: handle license exception in first item of a grouping

tmzullinger:
The change in 7d707f7f (TagsCheck: handle license exception in grouping, 2023-03-03) is insufficient when the exception is on the first item in a grouping, e.g.:

    License: (Apache-2.0 WITH LLVM-exception OR NCSA) AND BSD-3-Clause

Adjust `license_exception_regex` to exclude a leading '(' as well as a trailing ')'.

This fixes a variation of the same issue which #1014 intended to fix.  I don't know why I thought a leading parenthesis would be handle by the `license_regex` and not needed in `license_exception_regex`.  (Just wishful thinking, perhaps.)

Here is the spec file used to create the additional test srpm:
[valid-exception-begin-grouping.spec.txt](https://github.com/rpm-software-management/rpmlint/files/11013417/valid-exception-begin-grouping.spec.txt)

If this doesn't warrant a separate srpm from the previous test, I could replace the srpm in the existing test case.  This new srpm has an exception at the start and end of the grouping (by virtue of being a grouping with only one entry).

As always, I'm happy to adjust things as needed.  Thanks!


danigm (approved):
Looks good, but maybe it's worth to keep the `\s` exception in the regular expression? After the usage of this regex there's a `str.strip` so it's working as expected, but maybe it's a better regex this one:

```
r'([^(\s]+)\s(?:WITH|with)\s([^)\s]+)'
```

tmzullinger:
> Looks good, but maybe it's worth to keep the `\s` exception in the regular expression? After the usage of this regex there's a `str.strip` so it's working as expected, but maybe it's a better regex this one:
> 
> ```
> r'([^(\s]+)\s(?:WITH|with)\s([^)\s]+)'
> ```

Yes, I think you're right about that.  Relying on a later `strip()` to clean up the whitespace has the potential to cause problems if the code is re-arranged.  That's worth avoiding.  I added a preparatory commit which restores the `\s` exclusion to the existing regex at the trailing end and then amended the commit which aims to fix the beginning to include it.

Thanks!

PR #1029, opened 2023-03-20
Releasing rpmlint as pre-commit hook

jpodivin:
Minimum setup needed for integration with the pre-commit system.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+- id: rpmlint
+  name: Check RPM package errors
+  description: rpmlint is a tool for checking common errors in RPM packages. 
+  entry: rpmlint
+  language: python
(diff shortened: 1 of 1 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: rpmlint
+  name: Check RPM package errors
+  description: rpmlint is a tool for checking common errors in RPM packages. 
+  entry: rpmlint
+  language: python
+  files: \.rpm
```
inknos, line 6: Hey, it would make sense to add the file types [here](https://github.com/pre-commit/identify/blob/main/identify/extensions.py#L193). rpmlint would work with `spec`, `src.rpm` and `rpm`. I suggest to add the missing extensions to the identify library used by pre-commit.

PR #1034, opened 2023-03-24
PythonCheck: Remove spaces from requirements before compare

danigm:
I've detected errors with some packages like cloud-init.

### DIFF
--- a/rpmlint/checks/PythonCheck.py
+++ b/rpmlint/checks/PythonCheck.py
@@ -176,6 +176,11 @@ class PythonCheck(AbstractFilesCheck):
                 continue
 
             module_name = match.group('name').strip().lower()
+            # ignore python-base packages, that are provided by python
+            # base interpreter library
+            if module_name == 'base':
+                continue
+
             names = set(self._module_names(module_name))
 
             if not (names & requirements):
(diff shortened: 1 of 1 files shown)


rpmlint/checks/PythonCheck.py
```diff
@@ -168,14 +168,20 @@ def _check_leftover_requirements(self, pkg, requirements):
...
             match = pythonpac.match(req)
             if not match:
                 continue
 
-            module_name = match.group('name').lower()
+            module_name = match.group('name').strip().lower()
+            # ignore python-base, python-devel packages, that are
+            # not python modules
+            base = ['base', 'devel']
+            if module_name in base:
```
marxin, line 182: Please use it directly `if module_name in ('base', 'devel')` ...

marxin (changes requested):

marxin:
May I merge it now, are you find with the current shape of the PR?

danigm:
> May I merge it now, are you find with the current shape of the PR?

I think we can merge now. I've started today to review the list of warnings in packages and this is what I've detected that can be improved.

It's possible that more PR to improve this will come in the future if I discover more false positive, but right now that's what I've found.

Issue #1036, opened 2023-03-26
Use tmp_path instead of tmpdir


afrid18:
Pytest documentation suggests that we should use tmp_path instead of tmpdir. As tmpdir returns a `py.path.local` object when compared to tmp_path uses stdlib `pathlib` version.

[Refernece](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures) to official docs from pytest.
[Stackoverflow](https://stackoverflow.com/a/36071167/12725151) answer recommendation.

PR #1037, opened 2023-03-26
Pytest documentation recommends refactoring "tmpdir" to "tmp_path"

afrid18:
According to pytest documentation, it is recommended to use tmp_path instead of tmpdir. closes rpm-software-management/rpmlint#1036


marxin:
Great, I welcome the change!

PR #1040, opened 2023-03-28
Exclude systemd directory in UsrLibBinaryException

### LINKED ISSUE #1035: only-non-binary-in-usr-lib reported for systemd files
I have an RPM that packages a systemd service unit file (to be installed under /usr/lib/systemd/system/) and a systemd preset file (to be installed under /usr/lib/systemd/system-preset/). When I run rpmlint on the RPM I get only-non-binary-in-usr-lib. I assume that the warning is for one or both of these files as they are the only ones being installed to /usr/lib/ (though it would be nice if the warning would mention the problematic file). I see in rpmlint/configdefaults.toml that there is a setting for UsrLibBinaryException, but it does not include the systemd directory. Should it?

This seems similar to https://bugzilla.redhat.com/show_bug.cgi?id=794777

```shell
$ rpmlint --version
2.4.0
$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="37 (Container Image)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora Linux 37 (Container Image)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f37/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
SUPPORT_END=2023-11-14
VARIANT="Container Image"
VARIANT_ID=container
```

cyungmann:
Update UsrLibBinaryException in configdefaults.toml to exclude the systemd directory since it is common to install non-binary preset and systemd unit files there.

Closes #1035

### DIFF
--- a/rpmlint/configdefaults.toml
+++ b/rpmlint/configdefaults.toml
@@ -259,7 +259,7 @@ SystemLibPaths = [
 PieExecutables = []
 # Architecture dependent paths in which packages are allowed to install files
 # even if they are all non-binary
-UsrLibBinaryException = '^/usr/lib(64)?/(perl|python|ruby|menu|pkgconfig|ocaml|lib[^/]+\.(so|l?a)$|bonobo/servers/|\.build-id|firmware)'
+UsrLibBinaryException = '^/usr/lib(64)?/(perl|python|ruby|menu|pkgconfig|ocaml|lib[^/]+\.(so|l?a)$|bonobo/servers/|\.build-id|firmware|systemd)'
 # List of compilation flags that are mandatory
 MandatoryOptflags = []
 # List of forbidden compilation flags


marxin:
Thank you for your contribution. Please provide a test-case for the extended regular expression.

cyungmann:
@marxin do you have any suggestions on how to go about doing that? Do I need to create one or more new binary RPMs to add to the test/binary/ directory and then update test/test_binaries.py? Or is there a simpler test you are thinking of?

Thanks,
Chris

marxin:
> @marxin do you have any suggestions on how to go about doing that? Do I need to create one or more new binary RPMs to add to the test/binary/ directory and then update test/test_binaries.py? Or is there a simpler test you are thinking of?

You can use a mockup class `FakePkg` and you can use it to mock dummy files that should not be reported by the check with your patch. Let me know if it works.

marxin:
Let me add the test, thanks for the contribution.

cyungmann:
Thank you so much, @marxin. I was about to take a look at this today.

PR #1041, opened 2023-03-29
Start using python-magic package in CI.

marxin:


marxin:
That would align with what we have in https://github.com/rpm-software-management/rpmlint/blob/36ea497cce532b3392a7741541ea02b7c6a65ea7/setup.py#L50

PR #1047, opened 2023-04-04
Introducing new file 'pytest.ini' a conf file for pytest

afrid18:
Having a dedicated file for storing pytest settings will be more professional and convenient.

[ref](https://docs.pytest.org/en/7.1.x/reference/customize.html)

image

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,1 +1,0 @@
-# Having this file, pytest will add project root to sys.path.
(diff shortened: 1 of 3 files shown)


danigm:
Looks good to me. In combination with [pyproject.toml](https://github.com/rpm-software-management/rpmlint/pull/1019) could be a step forward to remove the setup.cfg. When we have the pyproject configuration it's also possible to move from the pytest.ini to that new file using `[tool.pytest.ini_options]` section.

marxin (approved):

Issue #461, opened 2020-09-09
Add support for %autopatch/%apply_patch


voxik:
Using `%autopatch`/`%apply_patch` [[1]], the patches are detected as not applied. It would be nice to provide support for this RPM feature.

[1]: https://rpm.org/user_doc/autosetup.html

mcepl:
Or even better, forcing `%autosetup -p1` instead of `%setup -q`.

marxin:
Can you please provide more context for this? I would like to implement it, but don't know what's expected?

mcepl:
This is mostly task for `py2pack/pyp2rpm` or stuff like that, but whenever I create new SPEC file I immediately change `%setup -q` to `%autosetup -p1`, because a) it is more convenient and it guarantees consistency of patches, b) what OP wrote here: it checks that there are no forgotten patches and all patches are ready to be applied. Thus I would argue `%autopatch -p1`/`%autosetup -p1` makes better SPEC files and it is in the scope of rpmlint.

marxin:
Thanks for the explanation. I tend to close this issue then.

marxin:
I wrongly interpreted what @mcepl wrote, sorry.

danigm:
I think we can close this one, the missing patch check is now there and there are a lot of tests, for example: https://github.com/rpm-software-management/rpmlint/blob/main/test/test_speccheck.py#L42-L44

bernhardkaindl:
There is an issue where 

%patch -P1 -p1

Is not detected as applying Patch1, it is only detected as:

%patch -P 1 -p1

Example: 
%patch -P1 -p1
in 
https://download.copr.fedorainfracloud.org/results/bkaindl/uftrace/fedora-rawhide-x86_64/05749047-uftrace/uftrace.spec

Results in
uftrace.spec: W: patch-not-applied Patch1: gethostbyname.patch
in 
https://download.copr.fedorainfracloud.org/results/@fedora-review/fedora-review-2180989-uftrace/fedora-rawhide-x86_64/05749104-uftrace/fedora-review/review.txt

PR #1049, opened 2023-04-10
SpecCheck: Improve patch applied regex

danigm:
This patch improves the SpecCheck regular expression to detect patterns like `%patch -P1 -p1`.

Fix https://github.com/rpm-software-management/rpmlint/issues/461


marxin:
Thanks Daniel!

PR #1056, opened 2023-04-20
pam-modules: whitelist pam_wtmpdb (bsc#1209963)

mgerstner:


FilippoBonazziSUSE:
There is an abundance of bsc# in the commit message and PR title

FilippoBonazziSUSE (approved):
Whitelisting looks good.

Just fix the commit message if you don't mind, and we're good to go.

PR #1066, opened 2023-05-25
rpm 4.19.0 api fixes

tmzullinger:
The rpm-4.19.0 alpha release landed in Fedora Rawhide recently.  The changes broke rpmlint and rpmdiff.  The following patches fix the test suite and works in some basic testing.

This should definitely get some extra eyes and cautious review.   I'm far from familiar with the rpm API.

I tried to make the changes as minimal as possible.

Thanks!

### DIFF
--- a/rpmlint/checks/DocCheck.py
+++ b/rpmlint/checks/DocCheck.py
@@ -60,7 +60,7 @@ class DocCheck(AbstractCheck):
         core_reqs = {}  # dependencies of non-doc files
         doc_reqs = {}   # dependencies of doc files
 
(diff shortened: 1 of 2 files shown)


danigm (approved):
Looks good to me, and I've been checking and this API is compatible with rpm 4.18, so it's something safe to apply.

pmatilai (approved):
Looks good, thanks for porting this!

FWIW, this change is compatible all the way to rpm 4.12, not just 4.18. So it works in RHEL 8-9 but not 7 or older.

PR #1079, opened 2023-06-26
Implement a proper testing framework (GSoC23)

afrid18:
# Overview

This PR is intended work of Google Summer of Code 2023 [Project](https://github.com/openSUSE/mentoring/issues/189). This project aims to get rid of binary rpm packages that are being used for testing [rpmlint](https://github.com/rpm-software-management/rpmlint) codebase because RPM packages consume relatively more space. Hence mocking them in a way which no longer need Binary RPM packages just for testing purposes.

## Description

In this PR, I have mainly worked on PythonCheck tests, which I eventually able to remove 6 binary files after mocking those tests.  This approach can be carry forwarded to other checks as well.


.github/workflows/main.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [main, opensuse]
+    branches: [main, opensuse, GSoC23]
```
danigm, line 5: Let's remove this from the workflow, because right now there's a PR so this branch should be tested by the `pull_request` action.

rpmlint/pkg.py
```diff
@@ -759,26 +759,145 @@ def __init__(self, name, is_source=False):
...
+            rpm.RPMTAG_SUGGESTVERSION: '',
+            rpm.RPMTAG_ENHANCENAME: '',
+            rpm.RPMTAG_ENHANCEFLAGS: '',
+            rpm.RPMTAG_ENHANCEVERSION: '',
+            rpm.RPMTAG_SUPPLEMENTNAME: '',
+            rpm.RPMTAG_SUPPLEMENTFLAGS: '',
+            rpm.RPMTAG_SUPPLEMENTVERSION: '',
+        }
+
+    def _gather_aux(self, header, xs, nametag, flagstag, versiontag,
```
danigm, line 790: This is the same function that we've in the `Pkg` class, so we can move to the `AbstractPkg` class and remove the code from the `Pkg` class, so any child class will inherit the same code.

rpmlint/pkg.py
```diff
@@ -759,26 +759,145 @@ def __init__(self, name, is_source=False):
...
+            for loop in range(len(versions)):
+                name = byte_to_string(names[loop])
+                evr = stringToVersion(byte_to_string(versions[loop]))
+                if prereq is not None and flags[loop] & PREREQ_FLAG:
+                    prereq.append((name, flags[loop] & (~PREREQ_FLAG), evr))
+                else:
+                    xs.append(DepInfo(name, flags[loop], evr))
+        return xs, prereq
+
+    def _gather_dep_info(self):
```
danigm, line 806: The same as above, just move this method to the `AbstractPkg` class, so it's inherited and can be removed from the `Pkg` class too.

test/Testing.py
```diff
@@ -60,3 +60,16 @@ def get_tested_spec_package(name):
...
+
+
+def get_tested_mock_package(files=None, real_files=None, header=None):
+    mockPkg = FakePkg('mockPkg')
+    if files is not None:
+        for path, file in files.items():
+            if file.get('create_dirs'):
+                for i in PurePath(path).parents[:file.get('include_dirs', -1)]:
+                    mockPkg.add_dir(str(i))
+            mockPkg.add_file_with_content(path, file.get('content'), real_files=real_files)
```
danigm, line 72: Nitpick: This is getting complicated, maybe it's worth to move this code to a method in the `FakePkg` class, or to another helper function to reduce the if-for nesting. It's just 5 lines and it's not too complex right now, so we could keep the code as it is, but if we find a way to make it more readable, it could be an improvement.

danigm (commented):
Good work, we've starting to have some tests replaced with `FakePkg`.

PR #1084, opened 2023-07-11
pkg_resource is deprecated, importlib and packaging are standard now

afrid18:
In test/test_erlang.py file, 
`import pkg_resource` is deprecated. To replace we use `importlib` and `packaging` modules


danigm (approved):
Looks good to me.

PR #1085, opened 2023-07-13
Update search regexes for %service_* macros

dcermak:
These macros used to call systemctl directly, but as of May 2023 they rely on systemd-update-helper in openSUSE. This results in false negatives for every package using these macros.

### DIFF
--- a/.packit/rpmlint.spec
+++ b/.packit/rpmlint.spec
@@ -52,6 +52,8 @@ BuildRequires:  dash
 BuildRequires:  /usr/bin/appstream-util
 BuildRequires:  /usr/bin/checkbashisms
 BuildRequires:  /usr/bin/desktop-file-validate
+# required for the systemd test
+BuildRequires:  rpm_macro(_unitdir)
 
 %if 0%{?suse_version}
 BuildRequires:  myspell-en_US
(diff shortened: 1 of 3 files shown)


Vogtinator:
With filetriggers this is obsolete anyway?

CC @dimstar77 @fbuihuu

dcermak:
> With filetriggers this is obsolete anyway?

Then how about we warn about these macros being present?

Vogtinator:
The hard part is that not all distros use them yet - so it might be wrong to warn if it's missing, but it's not wrong to use them (yet).

fbuihuu:
> With filetriggers this is obsolete anyway?

Nope, the `service_*` macros are still needed. The reason (I think) is that we would need to know the name of each service introduced by a package install/update from within the %posttransfile script if we want to use filetriggers in this case.

Also the other rpm macros (such as %tmpfiles_create, %sysusers_create) are unfortunately still needed until bsc#1212798 and bsc#1041742 are addressed but you know the story already. BTW still no feedback about their progress :(

Vogtinator:
> > With filetriggers this is obsolete anyway?
> 
> Nope, the `service_*` macros are still needed. The reason (I think) is that we would need to know the name of each service introduced by a package install/update from within the %posttransfile script if we want to use filetriggers in this case.

The list of files are passed to scriptlets such as `%filetriggerin`, so you can do with that info whatever is necessary.

fbuihuu:
> The list of files are passed to scriptlets such as `%filetriggerin`, so you can do with that info whatever is necessary.

Ok so maybe some of the `%service_` macros could be converted to filetrigger but that's not done currently.

dirkmueller (approved):

PR #1087, opened 2023-07-13
bsc#1213292: adjust timesync1 dbus whitelisting

FilippoBonazziSUSE:
This commit moves `timesync1.service` and `timesync1.conf` from the `systemd` to the `udev` package, and likewise from the `systemd-mini` to the `udev-mini` ones.

These files had been reviewed and whitelisted in 2018, before we tracked whitelisting digests. I have briefly reviewed the changes that have been applied to these files since then (`timesync1.service` unchanged, `timesync1.conf` minor changes), and have now added per-file digests.

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -138,17 +138,6 @@ nodigests = [
 ]
 
 [[FileDigestGroup]]
-package = "systemd-mini"
-type = "dbus"
-note = "imported from rpmlint1 DBUSServices.WhiteList"
-bug = "bsc#1111254"
-nodigests = [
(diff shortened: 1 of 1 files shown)


FilippoBonazziSUSE:
I see that the failing CI check for `fedora:rawhide` is failing for the whole repo, and not just for this PR. @danigm can we ignore this failing CI job for the purpose of this PR?

danigm:
> I see that the failing CI check for `fedora:rawhide` is failing for the whole repo, and not just for this PR. @danigm can we ignore this failing CI job for the purpose of this PR?

Yes, the failure is because of the python version, looks like they are moving to python 3.12, but some libraries are not updated correctly. So it's safe to ignore that CI fail.

Issue #1105, opened 2023-08-23
Replace existing binary rpm in tests with FakePkg


danigm:
With the new testing framework it's possible to create `FakePkg` with metadata and files, so it's possible to replace binary rpm that are used just for testing simple functionality. This way it's possible to remove the number of binary files and it'll be easier to understand what the tests are doing.

Of course we cannot remove all the files inside `test/binary/` because it's good test with real rpms, but a lot of tests are using small rpm created just to test a check and the sources of these rpm are not in the repository so it's hard to modify it, so any test that we can update to use `FakePkg` instead of binary rpm will be an improvement.

* [ ] test_zypp_syntax.py
* [ ] test_zip.py
* [ ] test_xinetd.py
* [ ] test_tmp_files.py
* [ ] test_tags.py
* [ ] test_sysvinitonsystemd.py
* [ ] test_signature.py
* [ ] test_readelf_parser.py
* [ ] test_shlib_policy.py
* [ ] test_python.py
* [ ] test_pam_modules.py
* [ ] test_pkgconfig.py
* [ ] test_menuxdg.py
* [ ] test_mixed_ownership.py
* [ ] test_LSB.py
* [ ] test_logrotate.py
* [ ] test_lint.py
* [ ] test_lib_dependency.py
* [ ] test_icon_sizes.py
* [ ] test_filter.py
* [ ] test_files.py
* [ ] test_FHS.py
* [ ] test_erlang.py
* [ ] test_duplicates.py
* [ ] test_doc.py
* [ ] test_diff.py
* [ ] test_dbus_policy.py
* [ ] test_config.py
* [ ] test_config_files.py
* [ ] test_build_date.py
* [ ] test_binaries.py
* [ ] test_bashisms.py
* [ ] test_build_root.py
* [ ] test_appdata.py
* [ ] test_alternatives.py

PR #1115, opened 2023-10-03
test: Replace bin in test_zypp_syntax

danigm:
See https://github.com/rpm-software-management/rpmlint/issues/1105


danigm:
> Hi, I am interested in the GSoC project so I have been trying to understand what is being done in this PR.
> 
> I have a few questions:
> 
>     1. How did you determine what should be the requires for the fake package? (Are they standard rpmlibs that I would just add to any mock package or were they specifically chosen for these ones?)
> 

In this case I'm replacing an existing rpm file in a test, so the idea is to replicate what we are testing here.

>     2. Is 'bee' being rejected in the first test because it is not wrapped in an `otherproviders()`? (i.e if it was instead `otherproviders(bee)` then the assert statement would be `assert 'suse-zypp-otherproviders otherproviders(bee)' in out`)
> 
> 

Nothing is being rejected, 'bee' is not in the output because this package doesn't produce any error or warning with this particular check related to 'bee', so the 'bee' conflict defined is correct.

> So for any future PRs, I assume it is expected that just from looking at what the assert statements are I should be able to draft a whole mock package to replace the existing binary rpm without knowing the specfile used to generate the binary. (I couldn't find a pre-existing specfile for packageand from a commit before it was deleted)

Yes, that's the main problem with the current binaries, we don't have the spec file for a lot of them so we should inspect the internal rpm definition and try to replicate with the FakePkg. That's one of the reasons for this project, to reduce the number of binaries that are hard to modify or check.

PR #1116, opened 2023-10-05
polkit-rules-whitelist: add fudo rules (bsc#1215948)

mgerstner:
Package build log for comparison can be found here:

https://build.opensuse.org/package/live_build_log/security/fudo/openSUSE_Factory/x86_64


FilippoBonazziSUSE:
Broken, seems to be missing a `[[FileDigestGroup]]` header for the very first `fudo` entry.

FilippoBonazziSUSE (approved):
You're not specifying the `digester` field for any of the hashes, I assume there is a sane default?

Other than that LGTM.

mgerstner:
> You're not specifying the `digester` field for any of the hashes, I assume there is a sane default?

Yes the default is to hash the unmodified file contents. Since these Polkit rules are Java script we don't have a matching filter yet and the default suffices.

FilippoBonazziSUSE:
I think the mageia-cauldron CI failures are unrelated. Merging

PR #1117, opened 2023-10-05
Whitelist dbus service and configuration for policycoreutils-dbus (bsc#1213435)

FilippoBonazziSUSE:
Whitelist dbus service and configuration for policycoreutils-dbus (bsc#1213435)

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -1300,3 +1300,17 @@ bug = "bsc#1206163"
 path = "/usr/share/dbus-1/system.d/org.opensuse.tukit.Updated.conf"
 digester = "xml"
 hash = "79448ff58b92e27ed0fecfb0d496582e4df8467b015147bb30aef3189e27df57"
+
+[[FileDigestGroup]]
+package = "policycoreutils-dbus"
+type = "dbus"
+note = "management DBUS API use to manage an SELinux environment"
+bug = "bsc#1213435"
+[[FileDigestGroup.digests]]
+path = "/usr/share/dbus-1/system-services/org.selinux.service"
+digester = "shell"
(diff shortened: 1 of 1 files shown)


mgerstner:
Looks good content wise. I wonder if the file `/etc/dbus-1/system.d/org.selinux.conf` should be moved to `/usr/share/dbus-1/system.d`, since there is an effort to move all such files to /usr. There is still a number of packages that ship their stuff in /etc, but for new packages it could make sense to do it right from the start.

FilippoBonazziSUSE:
I saw that yesterday as well and was wondering the same. Is it really as simple as packaging the same file under `/usr/share` and everything will still work the same? Because right now I'm not keen to drop everything and investigate if it doesn't...

mgerstner:
> I saw that yesterday as well and was wondering the same. Is it really as simple as packaging the same file under `/usr/share` and everything will still work the same? Because right now I'm not keen to drop everything and investigate if it doesn't...

Yes there shouldn't be any side effects to that. From `man dbus-daemon`:

```
The standard system bus normally reads additional XML files from /usr/share/dbus-1/system.d. Third-party packages should install the default policies necessary for correct operation into that directory, which has been
supported since dbus 1.10 (released in 2015).

The standard system bus normally also reads XML files from /etc/dbus-1/system.d, which should be used by system administrators if they wish to override default policies.

Third-party packages would historically install XML files into /etc/dbus-1/system.d, but this practice is now considered to be deprecated: that directory should be treated as reserved for the system administrator.
```

FilippoBonazziSUSE:
Alright, let me fix the packaging, do some testing and update this PR

FilippoBonazziSUSE:
Packaging fix submitted in https://build.opensuse.org/request/show/1116005

I've updated the whitelisting here as well, can you recheck?

mgerstner (approved):

PR #1132, opened 2023-10-25
zypper-plugins: whitelist zypp-boot-plugin (bsc#1215988)

mgerstner:
Note: the plugin is a binary ELF object so we cannot use digests in the whitelisting


FilippoBonazziSUSE (approved):
lgtm

PR #1134, opened 2023-10-27
SystemdInstallCheck: Support systemd-update-helper

danigm:
systemd macros expands to a script using systemd-update-helper, this patch add some regular expressions to detect the usage of this macro in pre, post, preun and postun, to avoid false possitive when using this macro.

See https://bugzilla.suse.com/show_bug.cgi?id=1091680


danigm:
[Another PR](https://github.com/rpm-software-management/rpmlint/pull/1085/files) has been merged with a solution for this, so we can close.

PR #1138, opened 2023-11-02
dbus-services: whitelist passim service (bsc#1216434)

mgerstner:
OBS build log can be found here:

https://build.opensuse.org/package/live_build_log/home:dimstar:Factory/passim/openSUSE_Factory/x86_64


wfrisch (approved):
OK.
* The patches were reviewed by mgerstner: https://bugzilla.suse.com/show_bug.cgi?id=1216434#c11
* I had a brief second look at the recent commits in 0.1.4 and it seems fine to me. At least the security-related issues (NULL ptr deref and symlink following) were fixed reasonably well.
* Whitelisting paths and hashes are OK.

PR #1141, opened 2023-11-13
polkit-rules-whitelist: self auth as admin (bsc#1215652)

wfrisch:
https://bugzilla.suse.com/show_bug.cgi?id=1215652
see also jsc#PED-260

### DIFF
--- a/configs/openSUSE/polkit-rules-whitelist.toml
+++ b/configs/openSUSE/polkit-rules-whitelist.toml
@@ -30,10 +30,23 @@ hash = ""
 package = "polkit"
 type = "polkit"
 note = "default rule shipped by polkit, allows uid 0 to do everything"
-bug = "bsc#1125314"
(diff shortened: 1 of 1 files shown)


configs/openSUSE/polkit-rules-whitelist.toml
```diff
@@ -30,10 +30,23 @@ hash = ""
 package = "polkit"
 type = "polkit"
 note = "default rule shipped by polkit, allows uid 0 to do everything"
-bug = "bsc#1125314"
+bugs = ["bsc#1125314". "bsc#1215652"]
```
FilippoBonazziSUSE, line 33: CI complains about a "unclosed array". This line seems to be the culprit

wfrisch: Ah yeah, thanks, missed the period

FilippoBonazziSUSE (approved):
LGTM

PR #1142, opened 2023-11-15
polkit-rules-whitelist: self auth as admin (bsc#1215652)

wfrisch:
fix typo

### DIFF
--- a/configs/openSUSE/polkit-rules-whitelist.toml
+++ b/configs/openSUSE/polkit-rules-whitelist.toml
@@ -36,16 +36,22 @@ path = "/usr/share/polkit-1/rules.d/50-default.rules"
 hash = "3b5781af8a450c5184c7a2d5408f4af7d3c65f23548ee0962ad0eabb70072c32"
 
 [[FileDigestGroup]]
-package = "sudo"
+package = "sudo-policy-sudo-auth-self"
 type = "polkit"
-note = "see also https://build.opensuse.org/request/show/1114990"
+note = "self auth as admin"
 bug = "bsc#1215652"
 [[FileDigestGroup.digests]]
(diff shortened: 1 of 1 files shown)


FilippoBonazziSUSE:
Wow, so the hashes were both wrong and off by 1 character at the end.

Also, is the package name correct or shouldn't it be fixed as well?

wfrisch:
It should be fine now. Please double-check though.

FilippoBonazziSUSE (approved):
Package names are correct and hashes are identical to the ones reported by OBS

wfrisch:
Thanks, and sorry for the trouble.
A basic linter for the toml files would be helpful.

FilippoBonazziSUSE:
@danigm we've been seeing these mageia cauldron CI failures for a few days now. I believe they are unrelated to our changes, so I will merge this PR. Do you know what's up with them though?

danigm:
> @danigm we've been seeing these mageia cauldron CI failures for a few days now. I believe they are unrelated to our changes, so I will merge this PR. Do you know what's up with them though?

Looks like a problem related with the rpm package and a dnf plugin that's not updated, python3-dnf-plugins-core version 4.3.1 and it's fixed since version 4.4.1

danigm:
> Looks like a problem related with the rpm package and a dnf plugin that's not updated, python3-dnf-plugins-core version 4.3.1 and it's fixed since version 4.4.1

I've just [reported this bug](https://bugs.mageia.org/show_bug.cgi?id=32527) and disabled the [mageia in CI](https://github.com/rpm-software-management/rpmlint/commit/57e75f34f8ea369379234841e0f6e232b80bd6e3) until it's solved.

PR #1145, opened 2023-11-30
zypper-plugins: match the new shebang (bsc#1212476)

wfrisch:
see also:
- https://bugzilla.suse.com/show_bug.cgi?id=1212476
- https://build.opensuse.org/request/show/1129913

This new patch dynamically patches the shebang. In this case `/usr/bin/python3` gets changed to `/usr/bin/python3.11`.


wfrisch:
We decided to adapt the `shell` filter in rpmlint instead, so that it filters out the Python version in the shebang before hashing. The hashes will then stay the same and no further action will be necessary.

PR #1146, opened 2023-11-30
FileDigestCheck: Filter minor versions in shebangs.

wfrisch:
openSUSE is now dynamically patching shebangs of Python script in order to specify the exact minor version, instead of just the major version, e.g. !/usr/bin/python3 will be substituted with !/usr/bin/python3.11. This not only alters existing whitelistings but will also require us to update whitelistings for every minor version bump of the Python interpreter.

This commit makes the shell filter ignore minor interpreter versions in shebangs, e.g.  python3.11 will be read as python3. Arguments are supported: !/usr/bin/python3.11 -v
will be read as
!/usr/bin/python3 -v

See also:
https://bugzilla.suse.com/show_bug.cgi?id=1212476

### DIFF
--- a/rpmlint/checks/FileDigestCheck.py
+++ b/rpmlint/checks/FileDigestCheck.py
@@ -4,6 +4,7 @@ import hashlib
 from pathlib import Path
 import stat
 import xml.etree.ElementTree as ET
+import re
 
 from rpmlint.checks.AbstractCheck import AbstractCheck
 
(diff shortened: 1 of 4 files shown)


wfrisch:
Please don't merge yet.
This is just a suggestion, up for discussion

wfrisch:
P.S. regarding the use of the "shell" filter for Python, it was intended as such, according to the source:
```
# - shell like files (configuration files, shell scripts, Python
# scripts, Perl scripts etc) that can contain empty lines or comments
# introduced by '#'.
```

wfrisch:
Closing. We should discuss and possibly rework this first.

Issue #1147, opened 2023-12-03
Unable to decode utf-8 when rpmlint some debuginfo packages


carlosrodfern:
## How to reproduce:

```
$ curl -LO https://kojipkgs.fedoraproject.org//packages/libcap/2.69/1.fc40/x86_64/libcap-debuginfo-2.69-1.fc40.x86_64.rpm

$ rpmlint libcap-debuginfo-2.69-1.fc40.x86_64.rpm

(none): E: fatal error while reading libcap-debuginfo-2.69-1.fc40.x86_64.rpm: 'utf-8' codec can't decode byte 0xcc in position 448: invalid continuation byte

```

## The Cause

```sh

$ rpm2cpio libcap-debuginfo-2.69-1.fc40.x86_64.rpm | cpio -imvd

$ readelf -W -l ./usr/lib/debug/usr/lib64/libpsx.so.2.69-2.69-1.fc40.x86_64.debug

Elf file type is DYN (Shared object file)
Entry point 0x3a50
There are 13 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  PHDR           0x000040 0x0000000000000040 0x0000000000000040 0x0002d8 0x0002d8 R   0x8
  INTERP         0x004200 0x0000000000004200 0x0000000000004200 0x00001c 0x00001c R   0x10
      [Requesting program interpreter: �BY"��[�]
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x001318 0x001318 R   0x1000
  LOAD           0x002000 0x0000000000002000 0x0000000000002000 0x001c2d 0x001c2d R E 0x1000
  LOAD           0x004000 0x0000000000004000 0x0000000000004000 0x000713 0x000713 R   0x1000
  LOAD           0x004c40 0x0000000000005c40 0x0000000000005c40 0x0003c8 0x000608 RW  0x1000
  DYNAMIC        0x004c58 0x0000000000005c58 0x0000000000005c58 0x000200 0x000200 RW  0x8
  NOTE           0x000318 0x0000000000000318 0x0000000000000318 0x000040 0x000040 R   0x8
  NOTE           0x000358 0x0000000000000358 0x0000000000000358 0x0000ac 0x0000ac R   0x4
  GNU_PROPERTY   0x000318 0x0000000000000318 0x0000000000000318 0x000040 0x000040 R   0x8
[...]

PR #1148, opened 2023-12-03
fix readelf output encoding handling of errors

carlosrodfern:
Fixes: #1147


danigm (approved):
Looks good. Maybe we should add the `errors='replace'` to all the `readelf` calls, we've 4 more calls like this in the `readelfparser.py` file.

carlosrodfern:
> Looks good. Maybe we should add the `errors='replace'` to all the `readelf` calls, we've 4 more calls like this in the `readelfparser.py` file.

Thank you @danigm for the review. I added `errors='replace'` to the rest as well and rebased the PR.

PR #1149, opened 2023-12-04
FileDigestCheck: Filter minor versions in shebangs.

wfrisch:
openSUSE is now dynamically patching shebangs of Python script in order to specify the exact minor version, instead of just the major version, e.g. `#!/usr/bin/python3` will be substituted with `#!/usr/bin/python3.11`. This not only alters existing whitelistings but will also require us to update whitelistings for every minor version bump of the Python interpreter.

This commit makes the shell filter ignore minor interpreter versions in shebangs, e.g.  python3.11 will be read as python3. Arguments are supported: `#!/usr/bin/python3.11 -v` will be read as `#!/usr/bin/python3 -v`

See also:
https://bugzilla.suse.com/show_bug.cgi?id=1200664
https://bugzilla.suse.com/show_bug.cgi?id=1212476
https://build.opensuse.org/request/show/1129913/changes

### DIFF
--- a/rpmlint/checks/FileDigestCheck.py
+++ b/rpmlint/checks/FileDigestCheck.py
@@ -2,6 +2,7 @@ import contextlib
 from fnmatch import fnmatch
 import hashlib
 from pathlib import Path
+import re
 import stat
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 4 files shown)


FilippoBonazziSUSE (approved):

FilippoBonazziSUSE:
LGTM. Leaving open for @mgerstner to take a look from his side, otherwise I can merge and submit to OBS tomorrow.

wfrisch:
Thanks for the review, Filippo. I'll take care of the submission as soon as Matthias reviewed it.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -74,8 +75,11 @@ def parse_content(self):
                     # skip empty lines
                     continue
                 elif line_nr == 0 and stripped.startswith('#!'):
-                    # keep shebang lines instact
-                    pass
+                    # keep shebang lines mostly intact,
+                    # but ignore minor interpreter versions
+                    shebang = re.sub(r'\bpython3\.\d+\b', 'python3', line)
+                    yield shebang.encode()
```
mgerstner, line 81: could it make sense not to `yield` here another time, but to say `line = re.sub(...)` and continue using the `yield` after the if clauses?

That way any logic that is generally applied to lines stays in one place.

PR #1151, opened 2023-12-11
polkit-rules, sudoers: add plasma-branding-Kalpa for special wheel gr…

mgerstner:
…oup handling

This new flavour of openSUSE uses a similar approach to Aeon and the exact same files to achieve privileged handling of the wheel group.


FilippoBonazziSUSE (changes requested):
Commit message does not mention bsc# numbers.

LGTM after this is fixed.

FilippoBonazziSUSE (approved):

PR #1155, opened 2023-12-14
Fix Aeon/Kalpa whitelisting distinct filenames (bsc#1217706, bsc#1217707)

FilippoBonazziSUSE:
In #1151, we erroneously whitelisted Kalpa versions of Aeon files with the Aeon name. We mistakenly assumed that since the file contents and whitelisting hashes were the same across the two distributions, that the filenames would be the same as well. This is unfortunately not the case, and I missed it in my review.


mgerstner:
It seems the happiness about the alleged possibility to merge the two packages was too much for spotting the differing file names.

The change looks got to me. Will merge.

mgerstner (approved):

PR #1160, opened 2023-12-28
opensuse branch: pam-modules: separate standard PAM module that have been already revi…

mgerstner:
…ewed

This is only a documentation update:

- pam_namespace has been recently reviewed by me
- faillock, usertype and setquota also have had dedicated reviews some time ago

Create separate entries for these modules so we can better keep track of what is what.


mgerstner:
This is not urgent and all possible reviewers are currently unavailable due to the christmas season. Keeping this in the queue until reviewers become available again.

FilippoBonazziSUSE (approved):

PR #1174, opened 2024-01-26
FileDigestCheck: harmonize and extend support for globbing in path and package names

mgerstner:


rpmlint/checks/FileDigestCheck.py
```diff
@@ -444,7 +465,7 @@ def _check_symlink_exceptions(self, pkg, name):
         """
 
         for symlink_exception in self.symlink_exceptions:
-            if pkg.name not in self._gather_packages_from_dict(symlink_exception):
+            if self._matches_pkg(symlink_exception, pkg):
```
FilippoBonazziSUSE, line 468: Did you drop a `not` here? Looks to me like you're inverting the behaviour

configs/openSUSE/sysctl-whitelist.toml
```diff
@@ -137,31 +137,19 @@ digester = "shell"
...
 [[FileDigestGroup]]
-packages = [
-    "kubernetes1.24-kubeadm",
-    "kubernetes1.25-kubeadm",
-    "kubernetes1.26-kubeadm",
-    "kubernetes1.27-kubeadm",
-    "kubernetes1.28-kubeadm",
-    "kubernetes1.29-kubeadm"
-]
+package = "glob:kubernetes1.[2-9][0-9]-kubeadm"
```
FilippoBonazziSUSE, line 150: This glob would match all `kubernetes1.20` to `1.23` already whitelisted in the previous `FileDigestGroup`. I assume this works because entries are matched in order?

How would we handle a third subsequent `FileDigestGroup` with a different hash starting from e.g. a `kubernetes1.32`? Maybe you can explain this to me privately

FilippoBonazziSUSE (changes requested):

rpmlint/checks/FileDigestCheck.py
```diff
@@ -273,8 +273,16 @@ def _gather_packages_from_dict(self, d):
...
             return d['packages']
 
-    def _matches_pkg(self, digest_group, pkg):
-        return pkg.name in self._gather_packages_from_dict(digest_group)
+    def _matches_pkg(self, config_dict, pkg):
+        for candidate in self._gather_packages_from_dict(config_dict):
+            if pkg.name == candidate:
+                return True
+            elif candidate.startswith('glob:'):
+                pattern = candidate.split(':', 1)[1]
```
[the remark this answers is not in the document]
mgerstner: I dislike repeating the fixed string and using a variable for it would increase the line count and hurt readability. I'd keep it that way.

rpmlint/checks/FileDigestCheck.py
mgerstner: Good catch. I'll fix it.

configs/openSUSE/sysctl-whitelist.toml
mgerstner: Yeah this overmatching is a bit unfortunate but shouldn't hurt the algorithm.

The checker looks at all entries if it can find a combination where package name, path names and digests match.

In this instance this means that `kubernetes1.20 to 1.23` could now also ship the newer version (digest) of the sysctl files. An acceptable risk I guess.

rpmlint/checks/FileDigestCheck.py
mgerstner: I force pushed a rebased commit that fixes this.

FilippoBonazziSUSE: Should this have been caught by a test, btw?

configs/openSUSE/sysctl-whitelist.toml
FilippoBonazziSUSE: A bit unclean, but should be alright. Could it ever move backwards? That would be problematic

rpmlint/checks/FileDigestCheck.py
mgerstner: Obviously not, since it isn't covered by the tests :-P

It **could have** been caught. There's never 100 % coverage I guess.

configs/openSUSE/sysctl-whitelist.toml
FilippoBonazziSUSE: Sounds good. Should there ever be a version of a whitelisted file which could be dangerous to go back to, we can take special care to avoid that.

FilippoBonazziSUSE (approved):

Issue #1171, opened 2024-01-25
Spurious warning 'W: python-missing-require' when inspecting Fedora Python package


hroncok:
Downstream report https://bugzilla.redhat.com/show_bug.cgi?id=2260169

------

After updating to version 2.5.0 of rpmlint, I'm seeing false warning when inspecting Python RPM packages:

rpmlint: 2.5.0
configuration:
    /usr/lib/python3.12/site-packages/rpmlint/configdefaults.toml
    /etc/xdg/rpmlint/fedora-legacy-licenses.toml
    /etc/xdg/rpmlint/fedora-spdx-licenses.toml
    /etc/xdg/rpmlint/fedora.toml
    /etc/xdg/rpmlint/scoring.toml
    /etc/xdg/rpmlint/users-groups.toml
    /etc/xdg/rpmlint/warn-on-functions.toml
checks: 32, packages: 2

fawltydeps.noarch: W: python-missing-require importlib_metadata
fawltydeps.noarch: W: python-missing-require isort
fawltydeps.noarch: W: python-missing-require pip-requirements-parser
fawltydeps.noarch: W: python-missing-require pydantic
fawltydeps.noarch: W: python-missing-require setuptools

Downgrading rpmlint to 2.4.0 makes this go away:

rpmlint: 2.4.0
configuration:
    /usr/lib/python3.12/site-packages/rpmlint/configdefaults.toml
    /etc/xdg/rpmlint/fedora-legacy-licenses.toml
    /etc/xdg/rpmlint/fedora-spdx-licenses.toml
    /etc/xdg/rpmlint/fedora.toml
    /etc/xdg/rpmlint/scoring.toml
    /etc/xdg/rpmlint/users-groups.toml
    /etc/xdg/rpmlint/warn-on-functions.toml
checks: 31, packages: 1

Reproducible: Always

Steps to Reproduce:
1. Upgrade rpmlint to 2.5.0
2. Grab an RPM from koji (e.g. https://kojipkgs.fedoraproject.org//packages/fawltydeps/0.13.3/1.fc40/noarch/fawltydeps-0.13.3-1.fc40.noarch.rpm)
3. Run rpmlint on the RPM
Actual Results:  
Spurious warning: 'W: python-missing-require'

Expected Results:  
No warning when there's no issue.

`rpm -q --requires` shows all dependencies as expected:

$ rpm -q --requires -p fawltydeps-0.13.3-1.fc40.noarch.rpm 
[...]

PR #1175, opened 2024-01-29
config: Filter python require checks for Fedora

danigm:
In Fedora dependencies are generated from upstream metadata, so python checks are not needed here.

https://github.com/rpm-software-management/rpmlint/issues/1171

### DIFF
--- a/configs/Fedora/fedora.toml
+++ b/configs/Fedora/fedora.toml
@@ -328,6 +328,11 @@ Filters = [
     ' info-files-without-install-info-postin',
     ' info-files-without-install-info-postun ',
     ' postin-without-install-info ',
+
+# In Fedora dependencies are generated from upstream metadata, so python checks
+# are not needed here. https://github.com/rpm-software-management/rpmlint/issues/1171
+    'python-missing-require',
+    'python-leftover-require',
 ]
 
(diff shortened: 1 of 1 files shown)


configs/Fedora/fedora.toml
```diff
@@ -328,6 +328,11 @@ Filters = [
     ' info-files-without-install-info-postin',
     ' info-files-without-install-info-postun ',
     ' postin-without-install-info ',
+
+# In Fedora dependencies are generated from upstream metadata, so python checks
+# are not needed here. https://github.com/rpm-software-management/rpmlint/issues/1171
+    'python-missing-require',
+    'python-leftover-require',
```
hroncok, line 335: This is still useful. Especially when packages have leftover manual requires.

danigm: ok, so maybe it's good to do not filter at all or just filter the missing-require one?

hroncok: That is what I mean, yes. Sorry if it wasn't clear.

danigm: https://github.com/rpm-software-management/rpmlint/pull/1180

PR #1176, opened 2024-01-29
PythonCheck: Check for python deps like python3.12dist(foo)

danigm:
Fix https://github.com/rpm-software-management/rpmlint/issues/1171


hroncok:
Thanks!

PR #1178, opened 2024-01-30
sendmail: permissions moved from /etc to /usr (bsc#1219339)

wfrisch:
The contents has not changed, just the location. Also I decided to keep the old /etc entries temporarily, in order to avoid problems, just in case the permissions package is not updated simultaneously with sendmail. We can remove these entries once the dust has settled.

### DIFF
--- a/configs/openSUSE/permissions-whitelist.toml
+++ b/configs/openSUSE/permissions-whitelist.toml
@@ -36,15 +36,18 @@ hash = "d5e51380e7ec868a42d336c868fc012ab95cac771d95361504cc6040b8d86221"
 
 [[FileDigestGroup]]
 package = "sendmail"
+bugs = ["bsc#1219339"]
 type = "permissions"
 note = """Be careful about calculating digest sums here, the sendmail package
       replaces file contents during RPM build time with architecture dependent
-      values"""
+      values.
+      The globbing is due to the UsrMerge. We can remove /etc once the dust has
+      settled."""
 [[FileDigestGroup.digests]]
-path = "/etc/permissions.d/sendmail"
+path = "glob:{/etc,/usr/share/permissions}/permissions.d/sendmail"
 hash = "e09ca5efebd0b3c123afc2364f9745f4d85c4327fa83f709bccbaa64da764486"
 [[FileDigestGroup.digests]]
-path = "/etc/permissions.d/sendmail.paranoid"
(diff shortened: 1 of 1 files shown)


configs/openSUSE/permissions-whitelist.toml
```diff
@@ -39,13 +39,21 @@ package = "sendmail"
 type = "permissions"
 note = """Be careful about calculating digest sums here, the sendmail package
       replaces file contents during RPM build time with architecture dependent
-      values"""
+      values.
+      The duplicate entries are due to the UsrMerge. We can remove the /etc
+      entries once the dust has settled. See bsc#1219339"""
```
mgerstner, line 44: Now that we have a bug for this could you also add a `bug = "bsc#..."` field

configs/openSUSE/permissions-whitelist.toml
```diff
@@ -39,13 +39,21 @@ package = "sendmail"
...
 hash = "e09ca5efebd0b3c123afc2364f9745f4d85c4327fa83f709bccbaa64da764486"
 [[FileDigestGroup.digests]]
 path = "/etc/permissions.d/sendmail.paranoid"
 hash = "2d5c56cdfb00ec169c182de791cf2934331159842f1849c5f2d7059f0086bd2c"
+[[FileDigestGroup.digests]]
+path = "/usr/share/permissions/permissions.d/sendmail"
+hash = "e09ca5efebd0b3c123afc2364f9745f4d85c4327fa83f709bccbaa64da764486"
+[[FileDigestGroup.digests]]
+path = "/usr/share/permissions/permissions.d/sendmail.paranoid"
+hash = "2d5c56cdfb00ec169c182de791cf2934331159842f1849c5f2d7059f0086bd2c"
```
mgerstner, line 56: An alternative might be to used file globbing:

`path = "glob:{/etc,/usr/share/permissions}/permissions.d/sendmail`. Don't know if it's more intuitive, it's less redundant in any case.

configs/openSUSE/permissions-whitelist.toml
wfrisch: Added

configs/openSUSE/permissions-whitelist.toml
wfrisch: Reasonable; updated.

Issue #981, opened 2023-01-05
Known failing CI checks


marxin:
Can you please take a look @Conan-Kudo? It fails e.g. here:
https://github.com/rpm-software-management/rpmlint/pull/980/checks?check_run_id=10458797936

marxin:
And I also noticed the `aarch64` builder is broken for some reason:
https://download.copr.fedorainfracloud.org/results/packit/rpm-software-management-rpmlint-980/opensuse-tumbleweed-aarch64/05201667-rpmlint/builder-live.log.gz

```
Updating /etc/sysconfig/security ...
/var/tmp/rpm-tmp.QAB7Ka: line 27:  5557 Segmentation fault      (core dumped) /usr/bin/chkstat --system
```

Can you please investigate that as well?

marxin:
@Conan-Kudo: May I please remind this?

marxin:
`srpm_build_deps` seems to be resolved, keeping that opened for `opensuse-tumbleweed-aarch64`.

marxin:
Right now, we face an issue with `rpm-build:opensuse-tumbleweed-x86_64` that is related to the recent update of TW to `openssl-3`:

```
Public key for terminfo-base-6.4.20230121-3.1.x86_64.rpm is not installed. Failing package is: terminfo-base-6.4.20230121-3.1.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/opensuse/RPM-GPG-KEY-openSUSE
Error: GPG check FAILED
```

Hopefully it gets resolved by Packit Service soon.

marxin:
> Hopefully it gets resolved by Packit Service soon.

Hasn't been resolved yet:
https://github.com/packit/packit/issues/1852

Conan-Kudo:
This is related to openSUSE rotating the GPG key for Tumbleweed. It needs to be updated in https://github.com/xsuchy/distribution-gpg-keys.

marxin:
Note `rpm-build:mageia-cauldron-*` fails due to:

```
E   ImportError: zstd C API versions mismatch; Python bindings were not compiled/linked against expected zstd version (10504 returned by the lib, 10502 hardcoded in zstd headers, 10502 hardcoded in the cext)
```

marxin:
> Note `rpm-build:mageia-cauldron-*` fails due to:
> 
> ```
[...]

PR #1187, opened 2024-02-07
sysctl-whitelist: update YAMA config (bsc#1219656)

wfrisch:
https://bugzilla.suse.com/show_bug.cgi?id=1219656


wfrisch:
The 2 CI failures can be ignored, as per https://github.com/rpm-software-management/rpmlint/issues/981

FilippoBonazziSUSE (approved):

PR #1190, opened 2024-02-13
[opensuse] SUIDPermissionsCheck: support new /usr/share/permissions/packages.d dir

mgerstner:
- use common constant to refer to /usr/share/permissions
- adjust comments to current situation
- consider the new packages.d directory as well (new chkstat from permissions package supports this).

Due to the generator used in `_paths_to()` this is getting a bit ugly to combine now. We should only consider `packages.d` in the /usr dir, not in /etc, where the name is too generic.

### DIFF
--- a/rpmlint/checks/SUIDPermissionsCheck.py
+++ b/rpmlint/checks/SUIDPermissionsCheck.py
@@ -5,13 +5,15 @@ import rpm
 from rpmlint.checks.AbstractCheck import AbstractCheck
 from rpmlint.permissions import PermissionsParser, VariablesHandler
 
+SHARE_DIR = '/usr/share/permissions'
+
 
 class SUIDPermissionsCheck(AbstractCheck):
     def __init__(self, config, output):
         super().__init__(config, output)
         self.perms = {}
 
-        self.var_handler = VariablesHandler('/usr/share/permissions/variables.conf')
+        self.var_handler = VariablesHandler(f'{SHARE_DIR}/variables.conf')
 
         for fname in self._paths_to('permissions', 'permissions.secure'):
             if not os.path.exists(fname):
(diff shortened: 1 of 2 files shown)


rpmlint/checks/SUIDPermissionsCheck.py
```diff
@@ -100,38 +103,42 @@ def _paths_to(*file_names):
...
     def check(self, pkg):
         if pkg.is_source:
             return
 
         permfiles = set()
-        # first pass, find and parse permissions.d files
+        # first pass, find and parse per-package drop-in files
         for f in pkg.files.keys():
-            for prefix in self._paths_to('permissions.d/'):
+            for prefix in itertools.chain(
```
FilippoBonazziSUSE, line 116: Why did you have to use this chain construction rather than add another yield in _paths_to?

FilippoBonazziSUSE: I mean, we are sourcing files from 3 well known directories. Do we need to generate their name at runtime?

FilippoBonazziSUSE (approved):
Other than the comments above,  the rest looks ok

rpmlint/checks/SUIDPermissionsCheck.py
mgerstner: Maybe I'm blind but I seen no way to do that using the current interface of `_paths_to()`. It is being used for other kinds of path lookups. Passing `packages.d` as a parameter will yield an excess result.

FilippoBonazziSUSE: Yes, I realized after the first comment that you can't just reuse `_paths_to`. But in this case I think we can just do without the whole permissions dir name generation and use a static list

mgerstner: I believe I have found a compromise. Instead of turning generators inside out, simply transform the output to a list and use regular concatenation. This should be clearer now.

mgerstner:
Now the tumbleweed rpm-build fails. But I don't see this related to my change.

danigm:
> Now the tumbleweed rpm-build fails. But I don't see this related to my change.

The failing CI is not related to this change: https://github.com/rpm-software-management/rpmlint/issues/981#issuecomment-1929280102

FilippoBonazziSUSE (approved):
Looks cleaner now

PR #1191, opened 2024-02-14
[opensuse] whitelist a series of KDE6 changes

mgerstner:
The packages for referencing the hash digests for this can be found in `KDE:Frameworks` on OBS.

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -605,6 +605,16 @@ nodigests = [
 ]
 
 [[FileDigestGroup]]
+package  = "kf6-kauth"
+type     = "dbus"
+note     = "KDE6 kauth helper service - framework built around Polkit authentication for KDE applications"
+bug      = "bsc#1217178"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system.d/org.kde.kf6auth.conf"
+digester = "xml"
+hash     = "0309e22585e819d00213a570949be1b219b087a150ac90ae3ad1d2ed67545ac6"
+
+[[FileDigestGroup]]
 package = "firewalld"
 type = "dbus"
 note = "imported from rpmlint1 DBUSServices.WhiteList"
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -1084,6 +1084,20 @@ nodigests = [
 ]
 
 [[FileDigestGroup]]
+package  = "plasma6-disks"
+type     = "dbus"
+note     = "allows to invoke smartctl on block devices"
+bug      = "bsc#1217185"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system.d/org.kde.kded.smart.conf"
+digester = "xml"
+hash     = "83b4e002b9bdab964dd49e9aacca5531e5b3f28c524f43b90de74020b05e90ce"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system-services/org.kde.kded.smart.service"
+digester = "shell"
+hash     = "cce8e0b79baedb4dbc75ae1f4dbdfa4db626ce4161f6cc702ba85ba8bfbcc590"
+
+[[FileDigestGroup]]
 package = "kdenetwork-filesharing"
 type = "dbus"
 note = "Samba configuration helper"
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -196,6 +196,20 @@ nodigests = [
 ]
 
 [[FileDigestGroup]]
+package  = "plasma6-desktop"
+type     = "dbus"
+note     = "helper for setting time, date, ntp, timezone"
+bug      = "bsc#1217184"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system.d/org.kde.kcontrol.kcmclock.conf"
+digester = "xml"
+hash     = "ed225dc29086a779b4ff52cafbf9c65b7b31edb7436768726f7b5ba1f0539168"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system-services/org.kde.kcontrol.kcmclock.service"
+digester = "shell"
+hash     = "727328d0cf3839edcee6a95bebd7593df539fb36799839d3c574b2cde96aaebd"
+
+[[FileDigestGroup]]
 package = "plasma5-workspace"
 type = "dbus"
 note = "legacy: not audited"
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -393,6 +393,36 @@ nodigests = [
 ]
 
 [[FileDigestGroup]]
+package  = "powerdevil6"
+type     = "dbus"
+note     = "handles battery, backlight and GPU hardware settings"
+bug      = "bsc#1217187"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system.d/org.kde.powerdevil.backlighthelper.conf"
+digester = "xml"
+hash     = "f02b05e676bbb34326f20660226474344da140ff79f987998fb841946e23619b"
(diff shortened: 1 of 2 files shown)


configs/openSUSE/dbus-services.toml
```diff
@@ -604,6 +604,16 @@ nodigests = [
     "/usr/share/dbus-1/system.d/org.kde.kf5auth.conf",
 ]
 
+[[FileDigestGroup]]
+package  = "kf6-kauth"
+type     = "dbus"
+note     = "KDE6 kauth helper service - framework built around Polkit authentication for KDE applications"
+bug      = "bsc#1217178"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system.d/org.kde.kf6auth.conf"
```
FilippoBonazziSUSE, line 685: Ok

configs/openSUSE/dbus-services.toml
```diff
@@ -1299,6 +1299,20 @@ path = "/usr/share/dbus-1/system.d/org.kde.kinfocenter.dmidecode.conf"
 digester = "xml"
 hash = "20bdf679561f1bc1df39cda8365be3d876a04c4b6d38d9223c6a70ae1f607fc1"
 
+[[FileDigestGroup]]
+package  = "kinfocenter6"
```
FilippoBonazziSUSE, line 1389: Ok

configs/openSUSE/dbus-services.toml
```diff
@@ -214,6 +214,20 @@ nodigests = [
...
+note     = "send signals to and change scheduling properties of arbitrary processes"
+bug      = "bsc#1217182"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system.d/org.kde.ksysguard.processlisthelper.conf"
+digester = "xml"
+hash     = "e08be1fffd5b7ebddb31f396f603e7a9fac08784c5473f37195262b55c663d45"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system-services/org.kde.ksysguard.processlisthelper.service"
+digester = "shell"
+hash     = "dc9ea7f2e2c649aa9bd30ddfd3e497d23de6cf801424460723a0df0a56514aa"
```
FilippoBonazziSUSE, line 257: It seems that you dropped a leading "a" from the digest. OBS has the digest as:

```
adc9ea7f2e2c649aa9bd30ddfd3e497d23de6cf801424460723a0df0a56514aa
```

By the way: is it right that hashes are 65 characters long?

configs/openSUSE/dbus-services.toml
```diff
@@ -214,6 +214,20 @@ nodigests = [
     "/usr/share/dbus-1/system.d/org.kde.ksysguard.processlisthelper.conf",
 ]
 
+[[FileDigestGroup]]
+package  = "libksysguard6-plugins"
+type     = "dbus"
+note     = "send signals to and change scheduling properties of arbitrary processes"
+bug      = "bsc#1217182"
+[[FileDigestGroup.digests]]
+path     = "/usr/share/dbus-1/system.d/org.kde.ksysguard.processlisthelper.conf"
```
FilippoBonazziSUSE, line 251: Ok

configs/openSUSE/pam-modules.toml
```diff
@@ -421,10 +421,10 @@ nodigests = [
...
 
 [[FileDigestGroup]]
-package = "pam_kwallet"
+packages = ["pam_kwallet", "pam_kwallet6"]
 type = "pam"
 note = "imported from rpmlint1 PAMModules.WhiteList"
-bug = "bsc#993806"
+bugs = ["bsc#993806", "bsc#1217183"]
 nodigests = [
     "glob:*/security/pam_kwallet5.so",
```
FilippoBonazziSUSE, line 429: Ok

Very strange that they still use the *5 names for kde6 components

configs/openSUSE/dbus-services.toml
```diff
@@ -195,6 +195,20 @@ nodigests = [
     "/usr/share/dbus-1/system.d/org.kde.kcontrol.kcmclock.conf",
 ]
 
+[[FileDigestGroup]]
+package  = "plasma6-desktop"
```
FilippoBonazziSUSE, line 199: Ok. Package is still failing build on OBS though, might still change

configs/openSUSE/dbus-services.toml
```diff
@@ -1083,6 +1083,20 @@ nodigests = [
     "/usr/share/dbus-1/system.d/org.kde.kded.smart.conf",
 ]
 
+[[FileDigestGroup]]
+package  = "plasma6-disks"
```
FilippoBonazziSUSE, line 1131: Ok

configs/openSUSE/dbus-services.toml
```diff
@@ -392,6 +392,36 @@ nodigests = [
     "/usr/share/dbus-1/system.d/org.kde.powerdevil.backlighthelper.conf",
 ]
 
+[[FileDigestGroup]]
```
FilippoBonazziSUSE, line 409: All ok

configs/openSUSE/dbus-services.toml
```diff
@@ -219,6 +219,20 @@ nodigests = [
     "/usr/share/dbus-1/system.d/org.kde.fontinst.conf",
 ]
 
+[[FileDigestGroup]]
```
FilippoBonazziSUSE, line 222: Commit message for this change does not contain bugzilla reference (bsc#1217186)

configs/openSUSE/dbus-services.toml
```diff
@@ -219,6 +219,20 @@ nodigests = [
...
+package  = "plasma6-workspace"
+note     = "supports managing system fonts - this is a high risk legacy D-Bus interface"
+bug      = "bsc#1217186"
+type     = "dbus"
+[[FileDigestGroup.digests]]
```
FilippoBonazziSUSE, line 227: Whitelisting contents ok

FilippoBonazziSUSE (changes requested):

FilippoBonazziSUSE:
I accidentally performed my review on the OBS project `KDE:Unstable:Frameworks`, and not on `KDE:Frameworks`. It should be ok anyway, since I didn't spot any significant differences in rpmlint outputs compared to the ones in this PR.

configs/openSUSE/dbus-services.toml
FilippoBonazziSUSE: Failing build was on `KDE:Unstable:Frameworks`, `KDE:Frameworks` is green.

configs/openSUSE/pam-modules.toml
mgerstner: I mentioned that explicitly for the PAM module but nobody bothered.

configs/openSUSE/dbus-services.toml
mgerstner: The hexdigests should be 64 characters, where do you see it differently?

I'll fix this missing `a`. We could add a unit test that verifies that hashes always have the proper length. This seems to be a common error.

mgerstner: I rebased the corresponding commit.

configs/openSUSE/dbus-services.toml
mgerstner: fixed

configs/openSUSE/dbus-services.toml
FilippoBonazziSUSE: > The hexdigests should be 64 characters, where do you see it differently?

I'm dumb, I checked it with `echo "..." | wc -c` instead of `echo -n "..." | wc -c`, so I counted the newline as well. I knew it didn't make sense that sha256 produces a 65 character output.

FilippoBonazziSUSE: > I'll fix this missing a. We could add a unit test that verifies that hashes always have the proper length. This seems to be a common error.

Yes that was one of the main issues that I wanted to address in CI. However now with the `buildlog2whitelist` helper script, hopefully such instances should just disappear. I assume you produced this entry manually or before the script was in use?

FilippoBonazziSUSE (approved):
LGTM

PR #1197, opened 2024-02-19
dbus-services: power-profiles-daemon (bsc#1219956)

wfrisch:
The service was renamed from `net.hadess.PowerProfiles`, however the old service is being kept to ensure backwards compatibility for a while.
https://bugzilla.suse.com/show_bug.cgi?id=1219956#c2

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -1187,7 +1187,15 @@ nodigests = [
 package = "power-profiles-daemon"
 type = "dbus"
 note = "a daemon dealing with system power settings"
-bugs = [ "bsc#1189900", "bsc#1201125" ]
+bugs = ["bsc#1189900", "bsc#1201125", "bsc#1219957"]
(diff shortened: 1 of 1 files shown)


wfrisch:
Known unrelated CI failures:
> rpm-build:opensuse-tumbleweed-aarch64 — RPMs failed to be built.
> rpm-build:opensuse-tumbleweed-x86_64 — RPMs failed to be built.

https://github.com/rpm-software-management/rpmlint/issues/981#issuecomment-1929280102

Conan-Kudo:
Or we could do c) ask the maintainer to add polkit configuration for both

wfrisch:
True, but if we change the package I would prefer to remove the old service instead in order to keep things minimal.

mgerstner:
Hmm, this is strange, is the old service supposed to work with the new Polkit configuration?

wfrisch:
Updated PR without the old D-Bus services, that have since been dropped in the RPM:
https://build.opensuse.org/request/show/1148703

PR #1200, opened 2024-02-21
SpecCheck: Add patch-macro-old-format check

danigm:
This new check is to warning about the usage of %patchN format to apply patches that is not supported by rpm >= 4.20

### DIFF
--- a/rpmlint/checks/SpecCheck.py
+++ b/rpmlint/checks/SpecCheck.py
@@ -18,6 +18,8 @@ def re_tag_compile(tag):
 
 
 patch_regex = re_tag_compile(r'Patch(\d*)')
+# rpm 4.20 doesn't support %patchN anymore, we should warn about this
+applied_patch_rpm420_regex = re.compile(r'^%patch(\d+)')
 applied_patch_regex = re.compile(r'^%patch\s*(\d*)')
 applied_patch_p_regex = re.compile(r'\s-P\s*(\d+)\b')
 applied_patch_pipe_regex = re.compile(r'\s%\{PATCH(\d+)\}\s*\|\s*(%\{?__)?patch\b')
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
This should be an error, not a warning.

kstreitova:
@Conan-Kudo I think that it's too soon for an "error". It's for RPM 4.20 which will be released in September-ish. Now, we would like to use the warning to detect affected packages and fix them before 4.20 hits the ground.

Conan-Kudo:
> @Conan-Kudo I think that it's too soon for an "error". It's for RPM 4.20 which will be released in September-ish. Now, we would like to use the warning to detect affected packages and fix them before 4.20 hits the ground.

It needs to be an error or it won't get fixed. Warnings do not matter as a general rule.

DimStar77:
> @Conan-Kudo I think that it's too soon for an "error". It's for RPM 4.20 which will be released in September-ish. Now, we would like to use the warning to detect affected packages and fix them before 4.20 hits the ground.

E: counts as one error point, unless we give it a higher badness. Regular builds are passing with up to 999 error points; so, IMHO, ERR is ok - badness of 1. Gives more urgency to the reader, and should still not fail the build)

kstreitova:
Sounds fair, let's go with error then :+1:

danigm:
I've updated it to be an error instead of a warning

rpmlint/checks/SpecCheck.py
```diff
@@ -18,6 +18,8 @@ def re_tag_compile(tag):
 
 
 patch_regex = re_tag_compile(r'Patch(\d*)')
+# rpm 4.20 doesn't support %patchN anymore, we should warning about this
```
DimStar77, line 21: "we should warn"

rpmlint/descriptions/SpecCheck.toml
```diff
@@ -16,6 +16,14 @@ use a directory for building, use $RPM_BUILD_ROOT instead.
 patch-not-applied="""
 A patch is included in your package but was not applied.
 """
+patch-macro-old-format="""
+The usage of %patchN is not supported by RPM >= 4.20. The preferred way to
+apply patches are, in order:
+ * %autosetup -p1
+ * %autosetup -N / %autopatch -p1
+ * %setup / %patch -P <N> -p 1 (upper case P denotes patch number,
+   lower case is the usual patch -p>num> strip level)
```
DimStar77, line 25: -p<num> (I'm sure you copied that from me.. sorry)

danigm: yep, copy/paste from your email :P

Conan-Kudo (approved):

PR #1201, opened 2024-02-22
CI: Disable tumbleweed builds for packit

danigm:
This is currently broken because of the Tumbleweed docker image.

https://github.com/rpm-software-management/rpmlint/issues/981#issuecomment-1929280102

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -14,8 +14,11 @@ jobs:
     - fedora-rawhide-aarch64
 #    - mageia-cauldron-x86_64
 #    - mageia-cauldron-aarch64
-    - opensuse-tumbleweed-x86_64
-    - opensuse-tumbleweed-aarch64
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@danigm looks like you have fixes in flight?

* https://build.opensuse.org/request/show/1145856
* https://build.opensuse.org/request/show/1149214

danigm:
> @danigm looks like you have fixes in flight?
> 
>     * https://build.opensuse.org/request/show/1145856
>     * https://build.opensuse.org/request/show/1149214

Yes, the fix should land at some point. I don't know if it's worth to disable right now or just wait a couple of days more and we don't need to touch anything.

Conan-Kudo:
Let's wait a bit and see. On the other hand, there's a separate commit in here to turn Mageia back on. Can we have a separate PR for that?

danigm:
All CI checks are working now, this is not needed.

PR #1209, opened 2024-03-05
[opensuse] updated whitelistings for gdm, systemd_loadkey and whitelisting maintenance

mgerstner:


configs/openSUSE/dbus-services.toml
FilippoBonazziSUSE, line 1: Ok

configs/openSUSE/pam-modules.toml
FilippoBonazziSUSE, line 1: Ok

FilippoBonazziSUSE (approved):
All ok

FilippoBonazziSUSE:
@mgerstner it looks like we jumped the gun with the GDM whitelisting, perhaps we should have kept both old and new for a while (see https://build.opensuse.org/request/show/1155534#comment-1905827). Can you look into this?

mgerstner:
> @mgerstner it looks like we jumped the gun with the GDM whitelisting, perhaps we should have kept both old and new for a while (see https://build.opensuse.org/request/show/1155534#comment-1905827). Can you look into this?

Usually the new packages go directly into Factory and that's it. If there is a problem in this case then it is no problem to re-create the old entry. Even you can do it :-)

PR #1211, opened 2024-03-08
Provide legacy whitelisting during bsc#1218922 transition period

FilippoBonazziSUSE:
Follow-up to #1209 for the gdm whitelisting (bsc#1218922) .

As requested in https://build.opensuse.org/request/show/1155534#comment-1905827, provide both the old and the new whitelisting values for the interim period until GNOME 46 is released.


FilippoBonazziSUSE:
I have added a Task in my calendar to review whether this temporary legacy whitelisting can be dropped in a couple of weeks, so that this does not get forgotten.

jsegitz (approved):

PR #1213, opened 2024-03-13
SpecCheck: Add no-%check-section warning

danigm:


Conan-Kudo (changes requested):

Conan-Kudo (approved):
This is a bit of a weird check, but it's only a warning...

danigm:
> This is a bit of a weird check, but it's only a warning...

Yep, It's a good practice to run tests when possible, but it'll be a lot of packages without check section. But in any case this could be filtered in configuration if it's too noisy in some distributions.

PR #1216, opened 2024-03-20
Mocked package in `test_xinetd.py`

chaitanyabisht:
Updated the `test_xinetd.py` to use FakePkg. I have also defined the mock package in the `test/mockdata` directory.

All tests are passing.


danigm (approved):
LGTM

PR #1218, opened 2024-03-28
Add forgejo group and user

Pi-Cla:
One day I snooped at the rpmlint for the Forgejo package on openSUSE https://build.opensuse.org/package/show/devel:tools:scm/forgejo and it asked me to add forgejo to the list so I have.

See also this lovely page on what Forgejo is: https://forgejo.org/


danigm (approved):
LGTM

PR #1221, opened 2024-04-03
dbus-services: power-profiles-daemon: reinstate legacy D-Bus whitelis…

mgerstner:
…ting (bsc#1219957)

The daemon actively still registers the old interface names and fails if this is not possible. Thus bring back the original whitelisting. Keep one entry for both legacy and current since these are actively used by the current package.


wfrisch (approved):
Looks good to me. Thanks!

wfrisch:
I'm unable to respond on BugZilla myself currently. Thanks for taking care of it, Matthias.

I will submit to OBS.

wfrisch:
https://build.opensuse.org/request/show/1164328

PR #1222, opened 2024-04-05
Clarify what the real intent is of this check

darix:
with the original message we got package submissions with empty %check sessions. Apparently this was not really the intention. so clarify the message.


danigm (approved):
LGTM

danigm:
The CI failing is not related to the change.

PR #1223, opened 2024-04-05
pam-modules: whitelist pam_lastlog2 now moved to util-linux (bsc#1222…

mgerstner:
…329)

Keep the old whitelisting in place for a bit longer until the old package is deleted from Factory.


FilippoBonazziSUSE (approved):
Ok. Let's remember to revisit this to remove the old whitelisting at some point.

Do we also need an action item to discuss creation of an ALP rpmlint branch now?

PR #1227, opened 2024-04-23
Whitelisting cleanup

mgerstner:


configs/openSUSE/pam-modules.toml
```diff
@@ -422,9 +422,9 @@ nodigests = [
...
 
 [[FileDigestGroup]]
-packages = ["pam_kwallet", "pam_kwallet6"]
+packages = ["pam_kwallet6"]
 type = "pam"
-note = "imported from rpmlint1 PAMModules.WhiteList"
+note = "unlocks the kwallet during the session phase of PAM login"
 bugs = ["bsc#993806", "bsc#1217183"]
 nodigests = [
     "glob:*/security/pam_kwallet5.so",
```
FilippoBonazziSUSE, line 430: Is the kwallet 6 PAM module still called pam_kwallet5.so even after the full KDE 6 release?

configs/openSUSE/dbus-services.toml
```diff
@@ -1148,35 +1148,36 @@ nodigests = [
...
-hash = "0b8742779a2d7bece8c8b552f505b4ff24628ee134e45b9af50ec7acc849ce22"
-[[FileDigestGroup.digests]]
-path = "/usr/share/dbus-1/system.d/org.deepin.dde.LocaleHelper1.conf"
-digester = "xml"
-hash = "/usr/share/dbus-1/system.d/org.deepin.dde.LocaleHelper1.conf"
-[[FileDigestGroup.digests]]
-path = "/usr/share/dbus-1/system-services/org.deepin.dde.LocaleHelper1.service"
-digester = "shell"
-hash = "b23c6c4f60b3b23859f5a0fba5c762c5dc487be4d94ddcc1c292801ee5e4062b"
+# put on hold until bsc#1211374 is resolved
```
FilippoBonazziSUSE, line 1151: Should we not rather drop this, and reference the dropping commit in the bsc# so that it can easily be reverted when (if) needed at a later date? I wouldn't like to intentionally introduce commented-out lines.

configs/openSUSE/pam-modules.toml
mgerstner: yes it's still named like this, although I mentioned it in the AUDIT bug

configs/openSUSE/dbus-services.toml
mgerstner: All right I dropped it. I guess already tomorrow the packager will come back to life, demanding this very entry from us :-P

FilippoBonazziSUSE: Indubitably

FilippoBonazziSUSE (approved):

FilippoBonazziSUSE:
CI seems to be stuck, I don't think we care. Merging

PR #1228, opened 2024-05-02
permissions: whitelist certification-sles-eal4 (bsc#1223682)

wfrisch:
https://bugzilla.suse.com/show_bug.cgi?id=1223682

### DIFF
--- a/configs/openSUSE/permissions-whitelist.toml
+++ b/configs/openSUSE/permissions-whitelist.toml
@@ -58,3 +58,10 @@ hash = "c4d3d806535c6737a07828ac84e297b87f6a406f6df67c9928c4fac71f47a17d"
 path = "/etc/permissions.d/texlive.texlive"
 hash = "c4d3d806535c6737a07828ac84e297b87f6a406f6df67c9928c4fac71f47a17d"
 
+[[FileDigestGroup]]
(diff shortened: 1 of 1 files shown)


FilippoBonazziSUSE (approved):
The whitelisting entry seems fine.

I am not sure where this should be submitted to, though. I guess it's fine to start with having it in Factory, but this is needed for SLE15SP6 only I imagine.

mgerstner:
> The whitelisting entry seems fine.
> 
> I am not sure where this should be submitted to, though. I guess it's fine to start with having it in Factory, but this is needed for SLE15SP6 only I imagine.

The package seems not to exist in Factory. Therefore having it in there is unnecessary and will likely also show up as an unused whitelisting in our checker. The latter can be suppressed, but as long as no rpmlint2 based product uses this I don't thing it makes sense to place it into rpmlint2.

For SLE15-SP6 we need to whitelist this on IBS in the old rpmlint. It needs a maintenance update which will take its time.

wfrisch:
> The package seems not to exist in Factory. Therefore having it in there is unnecessary and will likely also show up as an unused whitelisting in our checker. 
> For SLE15-SP6 we need to whitelist this on IBS in the old rpmlint. It needs a maintenance update which will take its time.

Issue #878, opened 2022-06-01
rpmlint check for ghost files without %attr()


darix:
having all permissions bits set to 0 should be rarely a valid case. rpmlint should warn packagers if they forgot to set the mode.

and in combination with systemd-tmpfiles it should ideally check that the mode is the same as in the tmpfiles.d config.

marxin:
@mgerstner What do you think about it? Can you please somehow include it to a check?

darix:
https://bugzilla.opensuse.org/show_bug.cgi?id=1223578

danigm:
I've just created a PR with a new check to warn about files without permissions, what do you think about this?
https://github.com/rpm-software-management/rpmlint/pull/1230/files

PR #1230, opened 2024-05-03
FilesCheck: warn about files with zero perms

danigm:
See https://github.com/rpm-software-management/rpmlint/issues/878

### DIFF
--- a/rpmlint/checks/FilesCheck.py
+++ b/rpmlint/checks/FilesCheck.py
@@ -326,6 +326,36 @@ def script_interpreter(chunk):
         if res and res.start() == 0 else (None, '')
 
 
+def find_perm_in_tmpfiles(pkg, fname):
+    """
+    Look for file path in all tmpdfiles.d file declared in this pkg and return
+    the permission column.
+    """
+
+    tmpd = []
+    perms = '0644'
+    user = 'root'
+    group = 'root'
+
+    for k, v in pkg.files.items():
+        if 'tmpfiles.d' not in k:
+            continue
+        if not os.path.exists(v.path):
+            continue
+        with open(v.path) as f:
+            tmpd += f.readlines()
+
+    for line in tmpd:
+        if f' {fname} ' not in line:
+            continue
+        try:
+            _t, _p, perms, user, group, *_rest = line.split()
+        except IndexError:
+            continue
+
+    return perms, user, group
+
+
 class FilesCheck(AbstractCheck):
     man_regex = re.compile(r'/man(?:\d[px]?|n)/')
     info_regex = re.compile(r'(/usr/share|/usr)/info/')
@@ -637,7 +667,13 @@ class FilesCheck(AbstractCheck):
         mode = pkgfile.mode
         perm = mode & 0o7777
         if not perm:
-            self.output.add_info('W', pkg, 'zero-perms', fname, '%o' % perm)
+            if pkgfile.is_ghost:
+                perms, user, group = find_perm_in_tmpfiles(pkg, pkgfile.name)
+                suggestion = f'Suggestion: "%ghost %attr({perms},{user},{group}) {pkgfile.name}"'
+                # Suggest the attrs if there's a tmpfiles.d
+                self.output.add_info('W', pkg, 'zero-perms-ghost', suggestion)
+            else:
+                self.output.add_info('W', pkg, 'zero-perms', fname, '%o' % perm)
 
     def _check_file_unexpandaed_macro(self, pkg, fname):
         for match in self.macro_regex.findall(fname):
--- a/rpmlint/checks/FilesCheck.py
+++ b/rpmlint/checks/FilesCheck.py
@@ -336,6 +336,7 @@ def find_perm_in_tmpfiles(pkg, fname):
     perms = '0644'
     user = 'root'
     group = 'root'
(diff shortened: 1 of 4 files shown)


rpmlint/descriptions/FilesCheck.toml
```diff
@@ -409,3 +409,11 @@ A file in /usr/bin is a link to a script in a different place with a shebang.
 rpm won't be able to inject the needed interpreter as dependency, so it should
 be done manually.
 """
+
```
darix, line 412: we should work on the guidance here. for all non %ghost entries it is often better to fix things in %install e.g.

only for %ghost we would need %attr() in %files. and then we should point the user to systemd tmpfiles.d (we could maybe even check if the file is covered by tmpfiles.d stuff and then spit out the correct %ghost %attr() ... line based on what we see in tmpfiles.d


if that comfort step is not possible we should mention that attr should match what is in the tmpfiles.d

danigm: What do you think should be the best way to handle this? To improve the error message or to improve the check and show a different warning message depending on if the file is ghost.

darix: well if we have access to the tmpfiles.d files I would definitely prefer to have the code spit out the correct stuff to give our packagers more guidance. do we have access to unpacked files at the point?

darix: otherwise the description should at least give a hint like:

`please ensure that the file mode (directory/file/symlink), ownership and permissions match what will be created on the filesystem later`

danigm: > well if we have access to the tmpfiles.d files I would definitely prefer to have the code spit out the correct stuff to give our packagers more guidance. do we have access to unpacked files at the point?

During this check we've access to all files defined in the rpm package. The package is expanded in the temporal folder, and we read those files to check for shebang, for example: https://github.com/rpm-software-management/rpmlint/blob/main/rpmlint/checks/FilesCheck.py#L879-L880

danigm: Help me to understand this better, do you know about some example pkg that we can use in a test case?

darix: redis e.g.

darix: sysconfig with its sysconfig-netconfig subpackage is also a good candidate probably.

rpmlint/checks/FilesCheck.py
```diff
@@ -326,6 +326,36 @@ def script_interpreter(chunk):
...
+
+    for k, v in pkg.files.items():
+        if 'tmpfiles.d' not in k:
+            continue
+        if not os.path.exists(v.path):
+            continue
+        with open(v.path) as f:
+            tmpd += f.readlines()
+
+    for line in tmpd:
```
darix, line 348: we should canonicalize the path in both entries do not have trailing slashes

rpmlint/checks/FilesCheck.py
```diff
@@ -336,6 +336,8 @@ def find_perm_in_tmpfiles(pkg, fname):
     perms = '0644'
     user = 'root'
     group = 'root'
+    if fname.endswith("/"):
```
darix, line 339:
```
>>> os.path.realpath("/some/thing/with///")
'/some/thing/with'
>>> 
```

no need to reinvent the wheel.

DimStar77:
I see a few build failures that seem to be related to this new check

```
[  287s] (none): E: fatal error while reading //home/abuild/rpmbuild/RPMS/x86_64/systemd-256.4-3.2.x86_64.rpm: [Errno 21] Is a directory: '/tmp/rpmlint.systemd-256.4-3.2.x86_64.rpm.87jsfd_o/etc/tmpfiles.d'
[  287s] Traceback (most recent call last):
[  287s]   File "/opt/testing/bin/rpmlint.real", line 8, in <module>
[  287s]     sys.exit(lint())
[  287s]              ^^^^^^
[  287s]   File "./rpmlint/cli.py", line 184, in lint
[  287s]   File "./rpmlint/lint.py", line 123, in run
[  287s]   File "./rpmlint/lint.py", line 91, in _run
[  287s]   File "./rpmlint/lint.py", line 264, in validate_files
[  287s]   File "./rpmlint/lint.py", line 290, in validate_file
[  287s]   File "./rpmlint/lint.py", line 283, in validate_file
[  287s]   File "./rpmlint/lint.py", line 299, in run_checks
[  287s]   File "./rpmlint/checks/FilesCheck.py", line 507, in check
[  287s]   File "./rpmlint/checks/FilesCheck.py", line 563, in _check_files
[  287s]   File "./rpmlint/checks/FilesCheck.py", line 606, in _check_file
[  287s]   File "./rpmlint/checks/FilesCheck.py", line 672, in _check_file_zero_perms
[  287s]   File "./rpmlint/checks/FilesCheck.py", line 346, in find_perm_in_tmpfiles
[  287s] IsADirectoryError: [Errno 21] Is a directory: '/tmp/rpmlint.systemd-256.4-3.2.x86_64.rpm.87jsfd_o/etc/tmpfiles.d'
```

(in this case, systemd build, same is observed in PackageKit and a few others)

PR #1238, opened 2024-06-14
users-groups.toml: add caddy

Idesmi:
Add user and group "caddy" to users-groups.toml

User and group are defined [here](https://build.opensuse.org/projects/openSUSE:Factory/packages/caddy/files/caddy.sysusers).


danigm (approved):
LGTM

PR #1242, opened 2024-06-19
dbus-services: systemd: keep the old whitelisting temporarily (bsc#1225317)

wfrisch:
It will take a while until the new systemd version will get submitted. Whitelist both versions of `org.freedesktop.home1.conf` until then.

<https://bugzilla.suse.com/show_bug.cgi?id=1225317>


mgerstner:
I'm afraid that won't work this way. The TOML syntax we've arrived at doesn't make it very clear any more, but a `[[FileDigestGroup]]` is supposed to be a logical entity of paths belonging together. With your change the same path appears twice, which will cause unexpected results in the checker, because a `set()` if paths is created for each group (something we might actually check for in the tests, or the code).

So to create a backward or forward compatibility entry a separate `[[FileDigestGroup]]` containing only the old/new files and digests is required.

mgerstner (approved):

wfrisch:
One CI failure, unrelated to our changes. Overriding.

PR #1246, opened 2024-06-28
Update main.yml

i12momal:
The file needs a change to fix the following error: ruff <path> has been removed. Use ruff check <path> instead.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,7 +114,7 @@ jobs:
       - uses: actions/checkout@v3
       - run: pytest
       - run: flake8
-      - run: /github/home/.local/bin/ruff check.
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -114,7 +114,7 @@ jobs:
       - uses: actions/checkout@v3
       - run: pytest
       - run: flake8
-      - run: /github/home/.local/bin/ruff .
+      - run: /github/home/.local/bin/ruff check.
```
danigm, line 117: There should be a space between the `check` parameter and the path:

```suggestion
      - run: /github/home/.local/bin/ruff check .
```

danigm (changes requested):

PR #1248, opened 2024-07-10
Replace bin in easy tests

i12momal:
Also, creating mock files to improve code cleanup of files.


test/binary/usr/share/appdata/broken-xml.metainfo.xml
danigm, line 1: These files should go to `test/files` folder.

test/mockdata/mock_appdata.py
```diff
@@ -0,0 +1,14 @@
+from Testing import get_tested_mock_package
+
+APPDATA = get_tested_mock_package(
+files={
+'/usr/share/appdata/broken.appdata.xml': {'content_path': '/usr/share/appdata/broken.appdata.xml'},
```
danigm, line 5: The correct path to use after move files to `test/files` will be:
```
'/usr/share/appdata/broken.appdata.xml': {'content-path': 'files/broken.appdata.xml'},
```

And the correct key is `content-path` not `content_path`, that's why even with the full system path this is working.

danigm (changes requested):
The test files should go to `test/files` folder. we don't need the full directory structure for tests.

The content path usage in the test is not right. Looks like these tests are working, even with these packages that are not exactly the same as what were trying to mock, but that could be because the tests are not checking a lot of things.

PR #1251, opened 2024-07-29
configs/openSUSE/users-groups.toml: add user and group openbao

johanneskastl:
User and group are defined here: https://build.opensuse.org/projects/openSUSE:Factory/packages/system-user-openbao/files/system-user-openbao.conf


danigm (approved):
LGTM

PR #1252, opened 2024-07-29
configs/openSUSE/users-groups.toml: add user and group vault-sync 

johanneskastl:
User is defined here: https://build.opensuse.org/projects/openSUSE:Factory/packages/system-user-vault-sync/files/system-user-vault-sync.conf


danigm (approved):
LGTM

PR #1253, opened 2024-07-29
configs/openSUSE/users-groups.toml: add user and group woodpecker 

johanneskastl:


danigm (commented):
Still not in factory, but it's on its way: https://build.opensuse.org/request/show/1190327

PR #1254, opened 2024-07-29
configs/openSUSE/users-groups.toml: add user and group alloy

johanneskastl:
User is defined here: https://build.opensuse.org/projects/openSUSE:Factory/packages/system-user-alloy/files/system-user-alloy.conf


danigm (approved):
LGTM

PR #1261, opened 2024-08-06
Update users-groups.toml

sbradnick:
Adding 'xpra' to 'StandardGroups' as suggested by rpmlint "A file in this package is owned by an unregistered group id."


danigm:
Group added by package xpra https://build.opensuse.org/package/show/openSUSE:Factory/xpra

danigm (approved):

PR #1267, opened 2024-08-18
Update users-groups.toml 3proxy

IlyaIndigo:
Added 3proxy group and user for 3proxy tiny proxy servers set.

### DIFF
--- a/configs/openSUSE/users-groups.toml
+++ b/configs/openSUSE/users-groups.toml
@@ -1,4 +1,5 @@
 StandardGroups = [
+    '3proxy',
     'aegis',
     'alloy',
     'antivir',
@@ -230,6 +231,7 @@ StandardGroups = [
 ]
 
 StandardUsers = [
+    '3proxy',
     'aegis',
     'alloy',
     'amanda',
--- a/configs/openSUSE/users-groups.toml
+++ b/configs/openSUSE/users-groups.toml
@@ -1,5 +1,4 @@
 StandardGroups = [
-    '3proxy',
     'aegis',
     'alloy',
     'antivir',
@@ -152,6 +151,7 @@ StandardGroups = [
     'privoxy',
     'prometheus',
     'prosody',
+    'proxy',
     'ptokax',
     'public',
     'pulse',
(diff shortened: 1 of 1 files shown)


danigm (commented):
As far as I can tell, this package is not part of Tumbleweed of Leap yet, so I think we shouldn't include.

IlyaIndigo:
> As far as I can tell, this package is not part of Tumbleweed of Leap yet, so I think we shouldn't include.

I actually want to include it in rpmlint first, and then add it to Factory.
I also want to find out if I can use usernames and group names starting with a number?

danigm (approved):
Okay, looks good with the new `proxy` name.

FilippoBonazziSUSE:
I think this accidentally went to the main branch instead of the opensuse branch... :grimacing: 

@danigm If this is intentional, then can you merge main into opensuse so that we get this change as well?

danigm:
> I think this accidentally went to the main branch instead of the opensuse branch... 😬
> 
> @danigm If this is intentional, then can you merge main into opensuse so that we get this change as well?

Yes, I'll merge main into opensuse today so all the changes to the configs file lands there. We've the `configs/openSUSE` files so I think it's okay to have these changes also in main branch.

IlyaIndigo:
So I should be submitting this PR to the opensuse branch, not main?

Can I send the PR again to the opensuse branch if I need to?

IlyaIndigo:
It might be worth updating the template of this post to make it clearer where to send PRs.
```3proxy.x86_64: W: non-standard-gid /var/lib/3proxy proxy
3proxy.x86_64: W: non-standard-gid /var/lib/3proxy/3proxy.cfg proxy
3proxy.x86_64: W: non-standard-gid /var/log/3proxy proxy
A file in this package is owned by an unregistered group id. To register the
group, please branch the devel:openSUSE:Factory:rpmlint rpmlint package, add
the group to the "config" file and send a submitrequest.
```

wfrisch:
> It might be worth updating the template of this post to make it clearer where to send PRs.

Good idea.
⇒ https://github.com/rpm-software-management/rpmlint/pull/1283

PR #1269, opened 2024-08-30
Add user and group sssd

scabrero:
Starting with sssd 2.10, it is possible to run sssd as an unprivileged user [1].

Fedora will use 'sssd' user and group name [2].

[1] https://sssd.io/release-notes/sssd-2.10.0-beta1.html
[2] https://src.fedoraproject.org/rpms/sssd/blob/rawhide/f/sssd.spec#_6


danigm (approved):
LGTM

PR #1270, opened 2024-09-03
configs/openSUSE/users-groups.toml: add nqptp (boo#1213060)

wfrisch:
nqptp [1] implemented my hardening suggestions [2], including a dedicated user/group for its daemon.

[1] https://github.com/mikebrady/nqptp
[2] https://bugzilla.suse.com/show_bug.cgi?id=1212951


FilippoBonazziSUSE (approved):
Changes are ok, I'm not sure the mentioned boo# is the right one though.

mgerstner:
Hmm, this user/group management is not strictly in the area of our security whitelistings, so formally it would be up the regular rpmlint maintainers to decide about this.

I'm not sure about the guidelines for adding users and groups there. Is every random user/group allowed to be added here? Seems like a cumbersome task with little value?

wfrisch:
> Changes are ok, I'm not sure the mentioned boo# is the right one though.

Yes, it's a bit unclear, but correct. The bug in the subject will be fixed by this update:
https://bugzilla.suse.com/show_bug.cgi?id=1213060

wfrisch:
> Hmm, this user/group management is not strictly in the area of our security whitelistings, so formally it would be up the regular rpmlint maintainers to decide about this.

I've committed new users/groups in the past and nobody complained:
https://github.com/rpm-software-management/rpmlint/commit/409e8a4ccb112e6c38c18fdc42abdbdd9e0cea29

> I'm not sure about the guidelines for adding users and groups there. Is every random user/group allowed to be added here? Seems like a cumbersome task with little value?

Sure, we could discuss this.

PR #1275, opened 2024-09-12
sssd dbus and polkit whitelisting entries (bsc#1230051)

mgerstner:


wfrisch (approved):
LGTM

Skimmed the bug report, double-checked metadata and hashes.

PR #1279, opened 2024-09-16
dbus-services: sssd: temporarily whitelist old version (bsc#1230051)

wfrisch:
This package will probably receive a minor update with the old D-Bus service in place, so we need to keep the old hashes intact. They can be removed after the next major update.

<https://bugzilla.suse.com/show_bug.cgi?id=1230051#c10>


mgerstner:
You need to create a dedicated `[[FileDigestGroup]]` for the old entry. I'm not sure what will happen when adding the same paths twice into the same group, but likely nothing good :-P

wfrisch:
Thanks. It should be OK now.

mgerstner (approved):

PR #1283, opened 2024-09-18
opensuse: explain pull requests more precisely

wfrisch:
i.e. mention our opensuse branch

### DIFF
--- a/configs/openSUSE/opensuse.toml
+++ b/configs/openSUSE/opensuse.toml
@@ -318,13 +318,15 @@ paths = [
 [SystemdTmpfilesWhitelist]
 
 [Descriptions]
-non-standard-uid = '''A file in this package is owned by an unregistered user id.
-To register the user, please make a pull request to the rpmlint config file
-configs/openSUSE/users-groups.toml in the rpmlint repository.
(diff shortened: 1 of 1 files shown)


wfrisch:
See also the discussion of this PR: https://github.com/rpm-software-management/rpmlint/pull/1267

danigm:
Tests are failing because the merge from main breaks some of them. I've just fixed it so rebasing this PR should fix the tests

Conan-Kudo:
I'm starting to wonder if we should pull the openSUSE branch back out again. I wasn't the biggest fan of it when it was instituted, but it's actively confusing to have openSUSE specific PRs here when we went through a lot of effort to make it easier for distro policies to be maintained externally.

FilippoBonazziSUSE:
I'll defer the discussion about the `opensuse` branch to another occasion when at least @jsegitz and @mgerstner can meaningfully participate (they're both unavailable right now).

For me personally the opensuse branch is not confusing and has been working fine for a long time.

FilippoBonazziSUSE (approved):
In the meantime, these changes are a net improvement which is both helpful and uncontroversial. Approving.

Issue #1292, opened 2024-10-31
chkstat has been renamed to permctl


DimStar77:
https://github.com/rpm-software-management/rpmlint/blob/1f09e5065c4f538e2e7fabd2d2deeaff2254dc6a/rpmlint/checks/SUIDPermissionsCheck.py#L62

The set_permissions and verify_permissions macro have been moved to the opensuse/permissions repo in

https://github.com/openSUSE/permissions/commit/fc19402c0c2c93e1b46e8d36a6c3e1f63f1ad868

The follow-up commit simplified the script to only accept permctl - as the macros are installed together with permctl, the name is defined 

Rpmlint now complains though that the set_permissions and verify_permissions would be missing - as it checks for chkstat in the script

DimStar77:
The actual rename happened in https://github.com/openSUSE/permissions/pull/188

danigm:
This should be solved now, so let's close.

PR #1324, opened 2025-03-10
SUIDPermissionsCheck: Check for permctl instead of chkstat

scabrero:
W: permissions-missing-verifyscript is incorrectly reported.

Related to https://github.com/rpm-software-management/rpmlint/issues/1292

### DIFF
--- a/rpmlint/checks/SUIDPermissionsCheck.py
+++ b/rpmlint/checks/SUIDPermissionsCheck.py
@@ -81,7 +81,8 @@ class SUIDPermissionsCheck(AbstractCheck):
             found = False
             if script:
                 for line in script.split('\n'):
-                    if '/chkstat' in line and path in line:
+                    escaped = re.escape(path)
+                    if re.search(fr'(chkstat|permctl) -n .* {escaped}', line):
                         found = True
                         break
 
--- a/test/test_suid_permissions.py
+++ b/test/test_suid_permissions.py
@@ -156,11 +156,21 @@ PERMCTL_PKG = get_tested_mock_package(
     /usr/bin/permctl -n --set --system /var/lib/perms/test || : \
   fi \
 """,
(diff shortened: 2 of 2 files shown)


scabrero:
Example from https://build.opensuse.org/build/openSUSE:Factory/standard/x86_64/sssd/_log:
```
[  122s] sssd.x86_64: W: permissions-missing-verifyscript missing %verify_permissions -e /usr/libexec/sssd/selinux_child
[  122s] sssd.x86_64: W: permissions-missing-verifyscript missing %verify_permissions -e /usr/libexec/sssd/sssd_pam
[  122s] sssd-krb5-common.x86_64: W: permissions-missing-verifyscript missing %verify_permissions -e /usr/libexec/sssd/krb5_child
[  122s] sssd-krb5-common.x86_64: W: permissions-missing-verifyscript missing %verify_permissions -e /usr/libexec/sssd/ldap_child
[  122s] Please add a %verifyscript section
```

rpmlint/checks/SUIDPermissionsCheck.py
```diff
@@ -81,7 +81,7 @@ def _check_post_scriptlets(self, pkg, path, need_verifyscript):
             found = False
             if script:
                 for line in script.split('\n'):
-                    if '/chkstat' in line and path in line:
+                    if re.search(fr'(chkstat|permctl) -n .* {path}', line):
```
danigm, line 84: There's a similar line before, maybe we should just do the same and `re.escape` the path here too.
```suggestion
                    escaped = re.escape(path)
                    if re.search(fr'(chkstat|permctl) -n .* {escaped}', line):
```

danigm (changes requested):
Looks good to me. Just a simple comment to escape the path before using it in the regex.

danigm (approved):
LGTM

PR #1329, opened 2025-04-04
Limit regexp in buildroot check

dirkmueller:


danigm (approved):
LGTM. Looks like mageia CI tests are failing but it's not related to this change.

PR #1331, opened 2025-04-10
FilesCheck: Don't fail with mtime in pyc > py

danigm:
This patch changes the python-bytecode-inconsistent-mtime to do not show if the pyc mtime is later than the source. We are seeing a lot of false positive errors like this building locally with OBS and this will fix the issue.

See https://bugzilla.suse.com/show_bug.cgi?id=1240894


danigm:
I'm not sure if this is the correct fix. Looks like cpython is checking that the dates are exactly the same, so if we modify this check we can pass packages that will be "wrong" and the interpreter will recreate .pyc files on first import / run.

https://github.com/python/cpython/blob/main/Lib/compileall.py#L236

mcepl:
> I'm not sure if this is the correct fix. Looks like cpython is checking that the dates are exactly the same, so if we modify this check we can pass packages that will be "wrong" and the interpreter will recreate .pyc files on first import / run.
> 
> https://github.com/python/cpython/blob/main/Lib/compileall.py#L236

Exactly, also for any messing with these fields in `*.pyc` files make @bmwiedemann unhappy in terms of binary reproducibility, and you don’t want to make Don Wiedemann unhappy, don’t you? Surely, comrade, you don't want Jones back?

bmwiedemann:
Normally, `SOURCE_DATE_EPOCH` is set during OBS builds, which triggers the hash-based verification mode of .pyc files (since python-3.6) ... but maybe there are places left that produce time-based .pyc files (e.g. `%check` section seems to not have S_D_E set) and those mtimes get adjusted to cause issues with mtime-based verification of .pyc freshness. 

In any case, this patch is not the right solution.

Issue #1330, opened 2025-04-09
"Provides: bundled()" and "Provides: deprecated()" are not ignored on Fedora


scfc:
`rpmlint` warns about `unversioned-explicit-provides` when linting for example Fedora's `binutils.spec`:

````
$ rpmlint binutils.spec 
========================================================================================================== rpmlint session starts =========================================================================================================
rpmlint: 2.6.1
configuration:
    /usr/lib/python3.13/site-packages/rpmlint/configdefaults.toml
    /etc/xdg/rpmlint/fedora-legacy-licenses.toml
    /etc/xdg/rpmlint/fedora-spdx-licenses.toml
    /etc/xdg/rpmlint/fedora.toml
    /etc/xdg/rpmlint/scoring.toml
    /etc/xdg/rpmlint/users-groups.toml
    /etc/xdg/rpmlint/warn-on-functions.toml
checks: 32, packages: 1

binutils.spec:328: W: unversioned-explicit-provides bundled(libiberty)
binutils.spec:467: W: unversioned-explicit-provides deprecated()
binutils.spec: W: no-%check-section
binutils.spec:1102: W: mixed-use-of-spaces-and-tabs (spaces: line 19, tab: line 1102)
binutils.spec: W: invalid-url Source0: binutils-with-gold-2.44.50-7109ea04ac7.tar.xz
binutils.spec:857: W: configure-without-libdir-spec
===================================================================== 0 packages and 1 specfiles checked; 0 errors, 6 warnings, 1 filtered, 0 badness; has taken 0.2 s ====================================================================
$ sed -ne '328p; 467p;' binutils.spec 
Provides: bundled(libiberty)
Provides: deprecated()
$
````

The Fedora Packaging Guidelines require both `Provides:`s to be present: `bundled()` by ["Bundling and Duplication of System Libraries"](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling), and `deprecated()` by ["Marking a package deprecated"](https://docs.fedoraproject.org/en-US/packaging-guidelines/deprecating-packages/#_marking_a_package_deprecated).

It would be nice if `rpmlint` could ignore those `Provides:`s and optionally enforce Fedora's guidelines on them (`bundled()` must be used as `bundled(<libname>)`, `deprecated()` may only be used as `deprecated()` or as `deprecated() = YYYYMMDD`).

scfc:
[...]

PR #1332, opened 2025-04-10
Fedora: filter unversioned-explicit warnings

danigm:
Fix https://github.com/rpm-software-management/rpmlint/issues/1330


Conan-Kudo:
We don't want this filtered out universally, just for those provides. It's still wrong normally.

Issue #1339, opened 2025-05-01
RPM declarative buildsystem throws: patch-not-applied


hroncok:
Consider this perfectly valid Fedora spec:

```
Name:           python-aiohappyeyeballs
Version:        2.6.1
Release:        %autorelease
Summary:        Happy Eyeballs for asyncio

License:        PSF-2.0
URL:            https://github.com/aio-libs/aiohappyeyeballs
Source:         %{url}/archive/v%{version}/aiohappyeyeballs-%{version}.tar.gz

Patch:          0001-Downstream-only-remove-pytest-options-for-coverage-a.patch

BuildSystem:            pyproject
BuildOption(install):   -L aiohappyeyeballs

BuildArch:      noarch

%description
...


%package -n     python3-aiohappyeyeballs
Summary:        %{summary}

%description -n python3-aiohappyeyeballs
...


%files -n python3-aiohappyeyeballs -f %{pyproject_files}
%license LICENSE
%doc CHANGELOG.md
%doc README.md


%changelog
%autochangelog
```

Note that it uses a declarative BuildSystem and has an implicit `%prep`.

Run rpmlint and you'll get:

```
python-aiohappyeyeballs.spec: W: patch-not-applied Patch0: 0001-Downstream-only-remove-pytest-options-for-coverage-a.patch
```

But that's bogus. There is an implicit %prep section with %autosetup.

I believe that rpmlint should not issue this warning when:

 - specfile uses a declarative BuildSystem
 - %prep is not explicitly overriden

hroncok:
https://github.com/rpm-software-management/rpmlint/pull/1340

PR #1340, opened 2025-05-05
When Declarative BuildSystem is used, assume patches are auto-applied

hroncok:
Fixes https://github.com/rpm-software-management/rpmlint/issues/1339


danigm (approved):
LGTM

PR #1343, opened 2025-05-08
Add a new warning for `setup.py install` usage

danigm:
[Setuptools](https://setuptools.pypa.io/en/latest/history.html#id6) added a deadline of Oct 31 to the setup.py install deprecation.

The usage of `setup.py install` should be replaced with `pip`. This PR adds a new warning when this command is found in the spec file or if the usage of the `%python_install` macro is found.

### DIFF
--- a/rpmlint/checks/SpecCheck.py
+++ b/rpmlint/checks/SpecCheck.py
@@ -83,6 +83,7 @@ pkgname_regex = re.compile(r'\s+(?:-n\s+)?(\S+)')
 tarball_regex = re.compile(r'\.(?:t(?:ar|[glx]z|bz2?)|zip)\b', re.IGNORECASE)
 
 python_setup_test_regex = re.compile(r'^[^#]*(setup.py test)')
+python_setup_install_regex = re.compile(r'^[^#]*(setup.py install|%\{?python\d*_install)')
 python_module_def_regex = re.compile(r'^[^#]*%{\?!python_module:%define python_module()')
 python_sitelib_glob_regex = re.compile(r'^[^#]*%{python_site(lib|arch)}/\*\s*$')
 
@@ -393,6 +394,7 @@ class SpecCheck(AbstractCheck):
         self._checkline_valid_groups(line)
         self._checkline_macros_in_comments(line)
         self._checkline_python_setup_test(line)
+        self._checkline_python_setup_install(line)
         self._checkline_python_module_def(line)
         self._checkline_python_sitelib_glob(line)
         self._checkline_shared_dir_glob(line)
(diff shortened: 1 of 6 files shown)


danigm:
@hroncok what do you think about this warning? Would the description work for fedora?

hroncok:
Our macro is `%py3_install`. Please also warn with that.

I also intend to deprecate the macro itself: https://src.fedoraproject.org/rpms/python-rpm-macros/pull-request/180

hroncok:
Thanks :heart:

test/spec/python-setup-python_install.spec
```diff
@@ -0,0 +1,33 @@
...
+%prep
+%setup -q
+
+%build
+%python_build
+
+%install
+%python_install
+%python3_install
+%python312_install
```
mcepl, line 23: see above for `%python3.12_install`.

rpmlint/checks/SpecCheck.py
```diff
@@ -83,6 +83,7 @@ def re_tag_compile(tag):
 tarball_regex = re.compile(r'\.(?:t(?:ar|[glx]z|bz2?)|zip)\b', re.IGNORECASE)
 
 python_setup_test_regex = re.compile(r'^[^#]*(setup.py test)')
+python_setup_install_regex = re.compile(r'^[^#]*(setup.py install|%\{?py(thon)?\d*_install)')
```
mcepl, line 86: `%python3.12_install` is not possible? That’s probably a request for comments from other distros.

test/test_spellchecking.py
```diff
@@ -51,7 +51,7 @@ def test_spellchecking():
     result = spell.spell_check(text, 'Description({}):')
     assert len(result) == 2
     assert result['tihs'].startswith('Description(en_US): tihs -> ')
-    assert get_suggestions(result['tihs']) == ['hits', 'this', 'ties']
```
mcepl, line 54: Nothing against anything, but shouldn’t this be a separate PR?

mcepl (changes requested):
Please, consider my comments.

test/test_spellchecking.py
hroncok: It's a different commit, good enough for me.

rpmlint/checks/SpecCheck.py
hroncok: Not possible. Macro names don't have dots.

```
$ LANG=C rpm --define 'python3.12_install ha'
warning: Macro %python3 needs whitespace before body
warning: Macro %python3 needs whitespace before body
```

mcepl: Perfect, thank you.

test/test_spellchecking.py
mcepl: ok

mcepl (approved):
All was explained

PR #1349, opened 2025-05-29
Add openQA users to users-groups.toml

b10n1k:
openQA defines and uses the `geekotest` and `_openqa-worker` users https://build.opensuse.org/projects/devel:openQA/packages/openQA/files/_service:set_version:obs_scm:openQA.spec?expand=1


configs/openSUSE/users-groups.toml
```diff
@@ -297,6 +298,7 @@ StandardUsers = [
     'ftp',
     'games',
     'gdm',
+    'geekotest'
```
danigm, line 301: There's a missing comma here, it should be

```suggestion
    'geekotest',
```

danigm (approved):
Looks good to me.

PR #1352, opened 2025-06-02
add user and group for opencloud-server

johanneskastl:
Based on current state in main. I can remove the commits that are not from me, if desired.


danigm (approved):
LGTM. Do we want this new `opencloud-server` also in the main branch?

Issue #1350, opened 2025-06-02
buildroot check now matching a script that was "undetected" for 10+ years


darix:
https://build.opensuse.org/projects/home:darix:apps/packages/ruby-common/files/gem_install.sh?expand=1

is that an intended real match or some false positive from the check?

ana:
This comes from https://build.opensuse.org/request/show/1280116

it could be a regression from:
```
* Fixup fallback regexp for buildroot detection
```

darix:
probably fallout from removing the anchoring in the fallback regexp.

danigm:
The buildroot detection regex now match this script, when we update to support rpm 4.20 here:
https://github.com/rpm-software-management/rpmlint/commit/2a5b7a4de4d7c0723f5b76064e8adf6f7f2977f8

The actual regex can be found here:
https://github.com/rpm-software-management/rpmlint/blob/main/rpmlint/checks/BuildRootAndDateCheck.py#L20-L25

```
    def __init__(self, config, output):
        super().__init__(config, output, r'.*')
        self.looksliketime = re.compile('(2[0-3]|[01]?[0-9]):([0-5]?[0-9]):([0-5]?[0-9])')
        self.istoday = re.compile(time.strftime('%b %e %Y'))
        self.prepare_regex(rpm.expandMacro('%{?buildroot}') or '/.{1,40}/BUILDROOT/')

    def prepare_regex(self, buildroot):
        for m in ('name', 'version', 'release', 'NAME', 'VERSION', 'RELEASE'):
            buildroot = buildroot.replace('%%{%s}' % (m), r'[\w\!-\.]{1,20}')
        self.lookslikebuildroot = re.compile(buildroot)
```

smithfarm:
It looks like @dirkmueller authored the change? Maybe he could weigh in.

dirkmueller:
but with rpm 4.20 it is just `/BUILDROOT` iirc. to me it looks like a valid match.

darix:
How about this? that should actually you know match what rpm 4.20 will use so we should only get files that match the current buildroot.

```
diff --git a/rpmlint/checks/BuildRootAndDateCheck.py b/rpmlint/checks/BuildRootAndDateCheck.py
index bf56fd06..486d3128 100644
--- a/rpmlint/checks/BuildRootAndDateCheck.py
+++ b/rpmlint/checks/BuildRootAndDateCheck.py
[...]

PR #1353, opened 2025-06-03
BuildRootAndDateCheck: More specific buildroot check

danigm:
This patch makes the BUILDROOT regex more specific for rpm >= 4.20. In this case the expandMacro('%{?buildroot}') will return and empty string and the fallback match any path with "/BUILDROOT". The change convert the cheking path to "/.../rpmbuild/...-...-build/BUILDROOT/"

Fix https://github.com/rpm-software-management/rpmlint/issues/1350


dirkmueller:
@danigm still needs a flake8 fix

PR #1355, opened 2025-06-05
[opensuse] document review bug for systemd sysctl coredump file 

mgerstner:
I forgot to add this when I made the change to the digest previously.


FilippoBonazziSUSE (approved):
Corresponding changes introduced in #1351

PR #1357, opened 2025-06-06
Consider /usr/etc in checks

mgerstner:
Since this is not strictly limited to security whitelisting checks it might also be interesting for @danigm to have a look.


FilippoBonazziSUSE:
Looks good to me, concerning that we've had a gap like this for so long. Leaving for Daniel to review

danigm (approved):
LGTM

mgerstner:
Thanks for reviewing, merging now!

PR #1359, opened 2025-06-18
dbus-services: whitelist libgpiod (bsc#1244702)

wfrisch:
https://bugzilla.suse.com/show_bug.cgi?id=1244702

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -1463,3 +1463,13 @@ hash     = "27fe3098931a301fcbe871e319f8fd6e50138447dec1080935151c41b29703c9"
 path     = "/usr/share/dbus-1/system-services/org.kde.kio.admin.service"
 digester = "shell"
 hash     = "ce9ed24db3c3654551bada756eb9aac5a58358364b41827b8154d0f73d6dc506"
+
+[[FileDigestGroup]]
+package  = "libgpiod"
+note     = "GPIO character device management daemon"
+bug      = "bsc#1244702"
+type     = "dbus"
+[[FileDigestGroup.digests]]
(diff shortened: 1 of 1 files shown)


configs/openSUSE/dbus-services.toml
```diff
@@ -1463,3 +1463,13 @@ hash     = "27fe3098931a301fcbe871e319f8fd6e50138447dec1080935151c41b29703c9"
 path     = "/usr/share/dbus-1/system-services/org.kde.kio.admin.service"
 digester = "shell"
 hash     = "ce9ed24db3c3654551bada756eb9aac5a58358364b41827b8154d0f73d6dc506"
+
+[[FileDigestGroup]]
+package  = "libgpiod"
```
DimStar77, line 1468: is 'package' the 'source package name' or the binary package name containing the file? If the latter, then there is an error, as this is packaged in libgpiod-manager

```
[   79s] libgpiod-manager.x86_64: E: dbus-file-unauthorized (Badness: 10000) /etc/dbus-1/system.d/io.gpiod1.conf (sha256 file digest default filter:c8033257b782186356e9ebf636a58d39c86de1b0a55d1c127b107e698d9cde85 shell filter:1acfa8f50efc9a37dd8e5fc14434b9c38cc5446fe991bc753a1239a36e475fb5 xml filter:4a620749a1b62fb229cd564e2ddafdd59e0302702c57ad7ac76ec5f56429f48b)
```

FilippoBonazziSUSE: Prepared fix in https://github.com/rpm-software-management/rpmlint/pull/1361 , only waiting for Wolfgang to review. Sorry that we missed this both in writing and reviewing the PR, it's uncharacteristic for us. Perhaps there's still something we can automate in the process

PR #1361, opened 2025-06-23
Fix libgpiod whitelisting package name (bsc#1244702)

FilippoBonazziSUSE:


wfrisch (approved):
Thanks for fixing my mistake.
Sorry for the trouble.

PR #1363, opened 2025-07-03
Be more forgiving before raising invalid spec file

dirkmueller:
there are cases where the name is "foo-variant" and there is only a "foo.spec", e.g. when the spec file is usign a flavor build in the Open Build Service. I think it is fine to not raise an error for those cases.


Conan-Kudo:
I would rather this extra check be handled as separate policy. Not everyone wants to permit that, and OBS flavor build quirks aren't a good reason to allow it everywhere unconditionally.

PR #1372, opened 2025-08-04
AlternativesCheck: detect update-alternatives on Fedora

xdelaruelle:
Update the update-alternatives-requirement-missing check to also detect "update-alternatives" on Fedora.

On such systems, update-alternatives is installed under /usr/bin. Thus the requirement may also be expressed as /usr/bin/update-alternatives or %{_bindir}/update-alternatives.


danigm (approved):
LGTM

Issue #1374, opened 2025-08-08
post-without-tmpfile-creation warning is not valid anymore in openSUSE


danigm:
%tmpfiles macro nowadays is a noop since it is done via triggers in the systemd package. This warning is shown in every package even if they are using `%tmpfiles_create %{_tmpfilesdir}/%{pkgname}.conf` in `%post`.

For example the package [matrix-synapse](https://build.opensuse.org/package/show/openSUSE:Factory/matrix-synapse), uses the macro in `%post` but it gets the warning in the rpmlint output.

We should disable this check for openSUSE, and maybe consider to remove it completely if it's not relevant for other distributions.

The change in macros was done in the revision 102: https://build.opensuse.org/package/rdiff/Base:System/systemd-rpm-macros?linkrev=base&rev=102

danigm:
As far as I can tell, fedora is not using the `systemd-tmpfiles --creeate` in their `POSTIN` so maybe it's safe to remove the `_check_post_tmpfile` method completely from `TmpFilesCheck`

FilippoBonazziSUSE:
> Sat Jun 24 08:40:00 UTC 2023 - Franck Bui <fbui@suse.com>
>  4+
>  5+- Turn %tmpfiles_create/%sysusers_create into NOPs
>  6+
>  7+  The 2 following macros have also been converted into NOPs since we moved to
>  8+  file triggers. Some packages might have assumed that their effects were
>  9+  effective as soon as the macros return. However such assumption on tmpfiles
> 10+  can't work on transactional systems anyways where changes must take place on
> 11+  reboot.
> 12+  
> 13+  When a system user/group needs to be created in %%pre, so proper ownership are
> 14+  used when package's files are installed, "sysusers_create_package()" should be
> 15+  used.

Maybe this should have been updated in the packaging guidelines?

https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Creating_files_and_subdirectories_in_/var/run_and_/run

@fbuihuu do you have more context on this?

danigm:
So I think we can safely merge the PR https://github.com/rpm-software-management/rpmlint/pull/1375, and drop the check completely as it's not needed anymore.

FilippoBonazziSUSE:
Nice, thanks for merging. So what is the current guideline now to report on the [wiki](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Creating_files_and_subdirectories_in_/var/run_and_/run)?

[...]

PR #1375, opened 2025-08-08
Remove post-without-tmpfile-creation warning

danigm:
This warning is not needed anymore. %tmpfiles macro nowadays is a noop since it is done via triggers in the systemd package.

Fix https://github.com/rpm-software-management/rpmlint/issues/1374

### DIFF
--- a/rpmlint/checks/TmpFilesCheck.py
+++ b/rpmlint/checks/TmpFilesCheck.py
@@ -30,7 +30,6 @@ class TmpFilesCheck(AbstractCheck):
                 continue
 
             self._check_pre_tmpfile(fname, pkg)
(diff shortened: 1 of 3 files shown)


FilippoBonazziSUSE:
@danigm , can we get this backported on `opensuse`? So we can address reports such as https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/5GIGESLINVGKVCQ5IWKGIVHKHD35UIID/

danigm:
> @danigm , can we get this backported on `opensuse`? So we can address reports such as https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/5GIGESLINVGKVCQ5IWKGIVHKHD35UIID/

done

PR #1383, opened 2025-08-20
Simplify `AbstractPkg._gather_aux` to resolve `consider-using-enumerate`

vil02:


rpmlint/pkg.py
```diff
@@ -409,18 +409,18 @@ def _calc_magic(self, pkgfile):
...
 
         if versions:
-            for loop in range(len(versions)):
-                name = byte_to_string(names[loop])
-                evr = stringToVersion(byte_to_string(versions[loop]))
-                if prereq is not None and flags[loop] & PREREQ_FLAG:
-                    prereq.append((name, flags[loop] & (~PREREQ_FLAG), evr))
+            names = header[nametag]
+            flags = header[flagstag]
+            for version, name_bytes, flag in zip(versions, names, flags):
```
vil02, line 417: Usually I would use:
```suggestion
            for version, name_bytes, flag in zip(versions, names, flags, struct=True):
```
This is available starting from python 3.10 (cf. [`PEP618`](https://peps.python.org/pep-0618/)) and this project claims to work for any python starting from 3.8:
https://github.com/rpm-software-management/rpmlint/blob/7b8902506fa1dc1c4357376d7825e2b4d5417688/README.md?plain=1#L33

danigm (approved):
LGTM

PR #1384, opened 2025-08-22
[opensuse] cleanup no longer used whitelisting entries

mgerstner:


wfrisch (approved):
Thanks for cleaning this up. LGTM.

PR #1388, opened 2025-09-02
scoring: raise badness for logrotate-user-writable-log-dir (bsc#1245961)

mgerstner:
Since this finding is security related it should fail the build by default.

There is no difference between the strict and non-strict scoring here, as is the case with whitelisting related diagnostics. This finding can be fixed by packagers on their own and no help from the security team is needed. Thus it should fail the build right away, not only in strict build configurations.


FilippoBonazziSUSE (approved):
LGTM

PR #1392, opened 2025-09-05
Synced /var/spool/mail entry with Factory filesystem package.

sudomibo:
It appears that the list in `FilesCheck.py` is no longer in sync with filesystem package of openSUSE Factory.

In particular, error `standard-dir-owned-by-package` is emitted on path `/var/spool/mail` that is not provided by the filesystem package (see https://build.opensuse.org/projects/openSUSE:Factory/packages/filesystem/files/directory.list?expand=1), and is instead owned by optional packages like `postfix`, `sendmail`, `exim`, etc. (see https://bugzilla.opensuse.org/show_bug.cgi?id=1179574).

New packages that provide `smtp_daemon` would be expected to own that path.  If I see correctly, this is different topic from the topic of allowlisting the typical 1777 permissions that goes through a separate security review process and has a different check in `rpmlint`.

Therefore, I suggest the attached change. Thank you in advance for your consideration.

### DIFF
--- a/rpmlint/checks/FilesCheck.py
+++ b/rpmlint/checks/FilesCheck.py
@@ -144,7 +144,6 @@ STANDARD_DIRS = (
     '/var/opt',
     '/var/preserve',
     '/var/spool',
-    '/var/spool/mail',
     '/var/tmp',
 )
 


mgerstner:
I'm not completely sure about this change. Other MTAs like postfix _do_ currently package /var/spool/mail and just live with the warning. I'm not entirely sure about the change to the filesystem package, maybe it was not properly completed?

`/var/spool/mail` kind of _is_ a standard directory, however. Maybe there is a conflict with the permissions package, because currently all MTAs invoke `permctl` to apply the `1777` permissions on the directory.

Before we change more things we should try to fully understand the current situation, maybe also talk to the people who made this change in the filesystem package. Ludwig Nussel, who initiated it, is no longer in the company, though.

sudomibo:
As far as I can [see](https://build.opensuse.org/projects/openSUSE:Factory/packages/postfix/files/postfix-rpmlintrc?expand=1), packages like `postfix` use `rpmlintrc` to ignore the mentioned error on OBS. In case that would be a standard way to go with packages that aim to provide `smtp_daemon`, this PR can be rejected.

mgerstner:
We discussed it in the team and nobody has any objections. As the configuration file says it should be kept in sync with the filesystem package and this is what we do with this PR#. I wouldn't want to spread this change also onto the main branch, however, since we don't know what the requirements for other distributions are.

Issue #1393, opened 2025-09-29
Permit repeating --rpmlintrc or at least permit passing it a fifo


qha:
I use rpmlint in what seems to be a somewhat atypical workflow.

Different jobs in the pipeline for the same commit builds packages for different %{dist}s.

As i want the warnings about unused filters but don't want to repeat myself i have split my rpmlintrc into one for warnings that occur from all dists and one per %{dist}.

Ideally i'd like to be able to provide one --rpmlintrc with each file and not have just the last one get ingested.

Failing that i'd like to at least be able to pass a fifo like with `--rpmlintrc <(cat .rpmlintrc .rpmlintrc$(rpm --eval '%dist'))`.

I'd be happy to work on a pull request for this if anyone is prepared to tell me which option/s is/are likely to get merged.

danigm:
I can see that we can extend the `rpmlintrc` arg to accept more than one parameters [here](https://github.com/rpm-software-management/rpmlint/blob/main/rpmlint/cli.py#L78):

```
    parser.add_argument('-r', '--rpmlintrc', '--file', nargs="*", type=_is_file_path, help='load up specified rpmlintrc file')
```

And then handles that correctly in the [lint.py](https://github.com/rpm-software-management/rpmlint/blob/main/rpmlint/lint.py#L40-L41)

PR #1395, opened 2025-10-02
Permit repeating --rpmlintrc

qha:
Closes: #1393


qha:
It looks like maybe flake8 is unhappy with my changes but i don't see why, could someone explain to me how to get at the check output?

Conan-Kudo:
We typically do rebase merges, so I would ask that you clean up your commits first.

danigm:
mageia CI build is failing because:
> This system is not registered with an entitlement server. You can use subscription-manager to register.

So it's not related to this change.

PR #1396, opened 2025-10-22
Add checks for atomic update compatibility

mtravitzky:
This checks for packages installing into locations which are illegal for compatibility with atomic updates. E.g. a package which installs into /var/ would get corrupted after a rollback from a newer snapshot.

### DIFF
--- a/configs/openSUSE/opensuse.toml
+++ b/configs/openSUSE/opensuse.toml
@@ -6,6 +6,10 @@ UseVarLockSubsys = false
 UseVersionInChangelog = false
 BadnessThreshold = 999
 
+# Set to true to issue a warning for ghost entries outside snapshots
+# when checking for atomic update compatibility
+AtomicCheckGhosts = false
+
 # Enabled checks for the rpmlint to be run (besides the default set)
 Checks = [
     "BashismsCheck",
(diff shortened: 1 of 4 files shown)


mtravitzky:
Re-ran flake8 locally, should be fine now

danigm:
@mtravitzky the check looks good, but maybe we should tweak the default configuration, at least for now. It's probably not wise to set the badness to 10000 immediately, it could produce a lot of broken packages in Factory once this change lands on the rpm package.

Vogtinator:
> I'm not sure this check is valid, since it's completely valid to install things into /var, even in transactional update mode.

No, `/var` is not available inside the transactional environment. It's not part of the snapshotted part of the system.

> Otherwise it's probably impossible to set up directory hierarchies and initialize data trees for persistent data required for applications and services.

Programs which need a specific hierarchy in `/var` can request creation through `tmpfiles.d` or create it on first start.

(For now there's still the "create-dirs-from-rpmdb" service which looks at the RPM database for directory creation, but it'll be removed in the future)

danigm (changes requested):
Looks good, just fix the comment to match the config new value

danigm (approved):

PR #1402, opened 2025-11-03
sysctl-whitelist: Support old aaa_base whitelisting

FilippoBonazziSUSE:
Support the old aaa_base sysctl whitelisting until the changes in SR#1314569 can be merged without breaking other packages in Factory.

This commit can then be simply reverted.


jsegitz (approved):
Thanks for handling this

PR #1403, opened 2025-11-06
Format and modify test/spec/bogus-date.spec

wanggai-kylin:
The description information has been changed to explain the purpose of this test case file.

Based on #1328

### DIFF
--- a/test/spec/bogus-date.spec
+++ b/test/spec/bogus-date.spec
@@ -1,20 +1,17 @@
 ### Based on macro-in-changelog.spec, I added the date information in changlog, as follows：
-Name:           macro-in
+Name:           bogus-date
 Version:        0
 Release:        0
-Summary:        macro-in
+Summary:        bogus-date
 License:        GPL-2.0-only
 Group:          Undefined
 URL:            http://rpmlint.zarb.org/#%{name}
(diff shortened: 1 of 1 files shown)


danigm:
> > > Yes, my intention is also to make test cases look more standardized and easier to read.
> > 
> > 
> > Sounds good, I've a pending PR with some cleaning for test cases: #1309. My plan was to remove as much binaries as I can, but I didn't find the time to complete this.
> 
> I'd like to learn more about this issue before making any changes. Could you please confirm if you're open to community contributions for this one?

This PR is part of the big issue of removing binary files from the tests: https://github.com/rpm-software-management/rpmlint/issues/1105

The idea is to replace the usage of `test/binary/*.rpm` with mock packages defined with `get_tested_mock_package`, you can see some of these mocks here: https://github.com/rpm-software-management/rpmlint/tree/main/test/mockdata

With this change it's easier to understand the test cases and even extend or modify without the need of generating new rpm files or inspect.

But the transition is difficult some times, because we need to make sure that during the transition we don't loose any important check, and because we don't have the sources of the binary rpm, you will need to unpack and inspect the rpm to be able to mock the tested functionality.

Depends on the test, we need to mock files, rpm metadata, pre/post scripts, etc.

PR #1405, opened 2025-11-11
Add (open)SUSE product check

adrianschroeter:
Currently just validating that the provided cpeid by rpm provides matches the one registered by the installed .prod file.

### DIFF
--- a/rpmlint/checks/ProductCheck.py
+++ b/rpmlint/checks/ProductCheck.py
@@ -1,0 +1,76 @@
+from xml.dom.minidom import parse
+from urllib.parse import unquote
+
+from rpmlint.checks.AbstractCheck import AbstractFilesCheck
+
+
+class ProductCheck(AbstractFilesCheck):
+    """
+    Validate that product files are correct. currently only cpeid.
+    """
+
+    def __init__(self, config, output):
+        super().__init__(config, output, r'/etc/products.d/.*\.prod$')
+
+    def check_file(self, pkg, filename):
+        cpeid_provider_found = None
+        cpeid_xml_found = None
+        for provide in pkg.provides:
+            if provide.name == 'product-cpeid()' and len(provide.version) > 1:
+                if cpeid_provider_found:
+                    self.output.add_info('E', pkg, 'product-cpeid-multiple-provider', 'multiple product-cpeid() provider, this is not specified yet', filename)
+                    return
+                cpeid_provider_found = unquote(provide.version[1])
+
+        if not cpeid_provider_found:
+            self.output.add_info('E', pkg, 'product-cpeid-no-provider', 'no product-cpeid() provider', filename)
+            return
+
+        lf = pkg.dir_name() + filename
+
+        try:
+            xml = parse(lf)
+        except xml.parsers.expat.ExpatError:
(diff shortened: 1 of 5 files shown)


rpmlint/checks/ProductCheck.py
```diff
@@ -0,0 +1,46 @@
...
+            for cpeid in xml.getElementsByTagName('cpeid'):
+                cpeid_xml_found = cpeid.firstChild.data
+
+            if not cpeid_xml_found:
+                self.output.add_info('E', pkg, 'product-parsing-exception', 'no cpeid defined in prod file', lf)
+
+            if cpeid_xml_found != cpeid_provider_found:
+                self.output.add_info('E', pkg, 'product-parsing-exception', 'cpeid defined different in prod file to rpm provides', lf)
+
+        except Exception as e:
```
dirkmueller: never create a "general exception" handling across many, can we just have it where it is actually necessary?

rpmlint/checks/ProductCheck.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def check_file(self, pkg, filename):
+        cpeid_provider_found = None
+        cpeid_xml_found = None
+        for provide in pkg.provides:
+            if provide.name == 'product-cpeid()' and len(provide.version) > 1:
+                  cpeid_provider_found = unquote(provide.version[1])
+
+        if not cpeid_provider_found:
+            self.output.add_info('E', pkg, 'product-parsing-exception', 'no product-cpeid() provider', filename)
```
dirkmueller: this shouldn't be "product-parsing-exception". each error should have its own unique identifier. the format is "check-name", "additional information".

rpmlint/checks/ProductCheck.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, config, output):
+        super().__init__(config, output, r'/etc/products.d/.*\.prod$')
+
+    def check_file(self, pkg, filename):
+        cpeid_provider_found = None
+        cpeid_xml_found = None
+        for provide in pkg.provides:
+            if provide.name == 'product-cpeid()' and len(provide.version) > 1:
+                  cpeid_provider_found = unquote(provide.version[1])
```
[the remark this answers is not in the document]
adrianschroeter: it is at least not clearly specified ... error out in that case for now ...

rpmlint/checks/ProductCheck.py
adrianschroeter: done

dirkmueller (approved):

dirkmueller (changes requested):
see comment on line 33

dirkmueller:
> I still question the validity of this check in rpmlint. Things like "product checks" should go with product creation tools, not rpmlint. I really think the concept of this check in rpmlint is a seriously bad idea.

I don't know. I mean, from all the things that are looking like a bad idea these days, I think this one is still on the okay scale of things. in openSUSE, products are just regular rpms, they just ship some files in some magic locations. having rpmlint check those files in those locations seem like a very sensible thing to be doing, because thats closest to the source of where the problem is coming from.

PR #1409, opened 2025-11-20
[opensuse] opensuse.toml: block filters for sudoers-, sysctl- whitelisting errors

mgerstner:
When these new restrictions have been added in commit 171c7ad1b7c925 and commit e9d5a19fe90b, we've overlooked that the diagnostics are not in the BlockedFilters list. This means packagers can currently reduce the badness of these warnings via their *.rpmlintrc files.

I checked the current list of rpmlint diagnostics on rpmlint.opensuse.org and luckily nobody seems to make use of this loophole at the moment.

Add the missing diagnostics to the list to sync everything now.


Conan-Kudo:
That is not okay at all. Filters need to be able to be disabled especially because OBS will fail package builds on it. For people building their own packages outside of openSUSE, it is not reasonable for this misfeature to exist.

Please remove it.

Conan-Kudo:
I regret that I didn't catch this when it was first added.

FilippoBonazziSUSE:
As mentioned, these measures apply to openSUSE Factory and are a fundamental part of our whitelisting process there. Third parties building software on OBS in their home or devel projects are not impacted by our whitelisting restrictions.

We of course do not mean to break any other workflow that is not subject to our whitelisting restrictions. If there are some real world issues which are escaping us, please do bring them up.

mgerstner:
> They need to be documented then, because if it's only in OBS that's not good enough. I would prefer this not be in the openSUSE default policy and be controlled in openSUSE:Factory in the OBS project space. It's a horrible default.

We have documented this in the openSUSE wiki [here](https://en.opensuse.org/openSUSE:Package_security_guidelines#Rpmlint_whitelisting_errors_in_home_and_devel_Projects).
 
> > You still don't seem to understand what we are doing here.
> 
> I understand just fine. But it's not okay for a tool that's used everywhere to be restrictive by default in ways that are unknown to control.

There are a couple of things wrong about this. Firstly, this specfic configuration is not used everywhere, it is tailored towards the needs of openSUSE and SUSE distributions only. Secondly the high badness values for the blocked filters diagnostics only appear in our "strict" configuration set, which is only active in production OBS projects.

I can assure you that our SUSE rpmlint maintainer and we people from the security team know quite well what we are doing here and there have not been any complaints about this setup for years, as in fact, we have improved the situation compared to what we had in rpmlint 1 before.

In this context I would like to ask you not to intrude this way in business that concerns the opensuse branch only. We are always open to discussions, but your activity in this PR# here has been somewhat rude.

PR #1413, opened 2025-12-12
permissions-whitelist: add migrated shadow-pw-mgmt (bsc#1254844)

mgerstner:
These are moved from the central permissions files to drop-in configuration files, to avoid conflicts with alternative password handling utilities (account-utils).

The package for is currently found in OBS in home:kukuk:no_new_privs/shadow


FilippoBonazziSUSE (approved):
changes lgtm

PR #1421, opened 2026-01-08
configs/openSUSE: whitelist Foomuuri D-Bus and sysctl files (bsc#1254385)

mgerstner:
The package for this is found in OBS in home:UbiquitousPhoton/Foomuuri at the moment.


FilippoBonazziSUSE (approved):
Hashes, package and file names match, bsc#s are correct, ...

PR #1426, opened 2026-01-13
[opensuse] polkit-rules-whitelist: add systemd-259 empower.rules (bsc#1255368)

mgerstner:
This is currently only found in an OBS devel project in `home:fbui:systemd:v259/systemd`, but systemd 259 has already been released, so we will need this sooner or later.


wfrisch (approved):
LGTM. Hash matches.

PR #1428, opened 2026-01-19
[opensuse] dbus-services: replace nodigests entries with digests entries (bsc#12…

mgerstner:
…15247)

When the digest-based whitelistings have first been introduced in rpmlint2 we have not been sure how well this new feature would work out in practice, which is why we used the "nodigests" approach for existing whitelistings and also for some newly introduced ones.

By now we have collected ample experience with digest-couplings and know that this restrictions works out well and changes are not overly frequent.

To improve the monitoring of D-Bus services in the distribution we are now converting all whitelistings to digest-based. The newly introduced digests have been automatically generated via a script and match the current version of the respective files as they are found in Factory.

I also inspected the files in question manually and looked out for any obviously problematic content, but couldn't find any.

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -34,10 +34,14 @@ digester = "xml"
 package = "upower"
 type = "dbus"
 note = "legacy: not audited"
-nodigests = [
-    "/usr/share/dbus-1/system-services/org.freedesktop.UPower.service",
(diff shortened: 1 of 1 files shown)


wfrisch (approved):
Wow, that's a lot of changes. Good work!
I only skimmed the paths and digester types. I assume it doesn't make sense to either check all the hashes manually or write my own script to verify them, since yours were already machine-generated.
LGTM.

PR #1432, opened 2026-01-20
[opensuse] InputPlumber D-Bus and Polkit whitelistings

mgerstner:
The Polkit whitelisting digest is based on a patched version of the file, I calculated the digest manually. See [Bugzilla comment](https://bugzilla.suse.com/show_bug.cgi?id=1249149#c35).

The devel package is found in OBS in `hardware/inputplumber`.


FilippoBonazziSUSE (approved):
Package name, filenames, bug numbers, hashes all match.

PR #1444, opened 2026-01-31
Fix TagsCheck for multiple InvalidRequires

### LINKED ISSUE #1443: [TagsCheck] InvalidRequires are only checked against first listed requirement and not any others
In `TagsCheck`, in `_check_multiple_dependencies` where it's checking the listed dependencies in the RPM against the invalid_requires ([around line 322](https://github.com/rpm-software-management/rpmlint/blob/main/rpmlint/checks/TagsCheck.py#L322)), the logic looks roughly like this:
```python
for dep in deps:
    for req in self.invalid_requires:
        if req == dep:
            add the error
```

This is all fine, except for the fact that `self.invalid_requires` is not an _iterable_, it's an _iterator_, produced by the [`map()` function on line 39](https://github.com/rpm-software-management/rpmlint/blob/main/rpmlint/checks/TagsCheck.py#L39). This means that on the first iteration of `for dep in deps`, the iterator gets consumed, and on subsequent loops the iterator is empty and so the check effectively is happening against an empty list. This means that the list of InvalidRequires is only checked against the first requirement of the RPM, and not any of the others.

ikmckenz:
Fixes #1443, this also adds a unit test to cover the case using an existing RPM from the test folder with multiple requirements.

### DIFF
--- a/rpmlint/checks/TagsCheck.py
+++ b/rpmlint/checks/TagsCheck.py
@@ -36,7 +36,7 @@ class TagsCheck(AbstractCheck):
         super().__init__(config, output)
         self.valid_groups = config.configuration['ValidGroups']
         self.valid_licenses = config.configuration['ValidLicenses']
-        self.invalid_requires = map(re.compile, config.configuration['InvalidRequires'])
+        self.invalid_requires = list(map(re.compile, config.configuration['InvalidRequires']))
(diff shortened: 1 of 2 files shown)


danigm (approved):
LGTM

PR #1455, opened 2026-02-10
[opensuse] SUIDPermissionsCheck: support permissions package coupling

mgerstner:
The next major version of permctl will support a new package coupling syntax in its profiles. The SUIDPermissionsCheck needs to be able to cope with this. This change extends the parser accordingly and also the PermissionsEntry data structure, which now carries an optional list of packages the entry relates to.

This change also includes a larger refactoring of the check, trying to make it easier to understand and to document it better.

### DIFF
--- a/rpmlint/checks/SUIDPermissionsCheck.py
+++ b/rpmlint/checks/SUIDPermissionsCheck.py
@@ -1,5 +1,4 @@
 import os
-import re
 import stat
 
 import rpm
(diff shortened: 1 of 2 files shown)


rpmlint/checks/SUIDPermissionsCheck.py
```diff
@@ -108,85 +101,105 @@ def _paths_to(*file_names):
...
-                    # FileDigestCheck and FileDigestLocation.
-                    bn = f'{dropin_dir}/' + f[len(prefix):].split('.')[0]
-                    if bn not in permfiles:
-                        permfiles.add(bn)
-
-        for f in permfiles:
+                if not f.startswith(prefix):
+                    continue
+                elif f in pkg.ghost_files:
+                    continue
```
FilippoBonazziSUSE:
```python
if not f.startswith(prefix) or f in pkg.ghost_files:
    continue
```
?

FilippoBonazziSUSE (approved):
Looks fine. The existing code could be cleaner and more idiomatic but that's out of scope for this.

PR #1457, opened 2026-02-11
GDM 50 beta D-Bus & Polkit Whitelistings

mgerstner:
These versions of the files can be found in OBS in GNOME:Next/gdm. I removed one no longer needed compatibility entry for GDM but needed to add new compatibility entries, since GDM 50 will only reach Factory in a couple of weeks.

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -71,17 +71,6 @@ hash = "00d25f898cf4c78cf45d68000ee00be27b0aa15b7ebe22c4ba86bc1b5f33b898"
 package  = "gdm"
 type     = "dbus"
 note     = "D-Bus interface for managing GDM sessions"
-bugs     = ["bsc#1204052", "bsc#1218922", "bsc#1230466"]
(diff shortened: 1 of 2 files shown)


FilippoBonazziSUSE:
The cleanup commit is not exactly 1:1 (it's not reverting 325d97126b8cdcff7e739be3a08615bc90ef578b, while it looks like it should), but it leaves the file in the correct state. Evidently the "temporary entry" was not removed and was repurposed as a fresh entry with a new digest at some point in the past.

FilippoBonazziSUSE (approved):

mgerstner:
> The cleanup commit is not exactly 1:1 (it's not reverting [325d971](https://github.com/rpm-software-management/rpmlint/commit/325d97126b8cdcff7e739be3a08615bc90ef578b), while it looks like it should), but it leaves the file in the correct state. Evidently the "temporary entry" was not removed and was repurposed as a fresh entry with a new digest at some point in the past.

Yeah, I believe the "temporary entry" comment was placed in front of the wrong block.

PR #1485, opened 2026-03-26
TagsCheck: Reorder is_devel and is_source parameters

danigm:
Fix https://github.com/rpm-software-management/rpmlint/issues/1483


rpmlint/checks/TagsCheck.py
vil02:
```suggestion
           is_devel: The param to check if a package name ends with *-devel
           is_source: Variable to check if a package is of source type
```

vil02 (approved):

PR #1489, opened 2026-04-06
AlternativesCheck: Fix .conf files regex

danigm:
The previous regex matches any path in /usr/share/libalternatives/ that ends with "conf" so it was crashing when trying to parse an rpm file that contains a binary "ldaptor-ldap2dhcpconf" with a libalternatives configuration.

This patch fixes this issue making the regex more specific so it doesn't catch any folder, just ".conf" files.

Cherry-pick: 27ee58ace
Fix https://bugzilla.suse.com/show_bug.cgi?id=1261308


FilippoBonazziSUSE (approved):
Thanks

PR #1492, opened 2026-04-08
dbus-services: systemd v260 (bsc#1259318)

wfrisch:
I also merged the duplicate sections for `systemd-container`.
As far as I can see, this was an artifact from the rpmlint1 import.

### DIFF
--- a/configs/openSUSE/dbus-services.toml
+++ b/configs/openSUSE/dbus-services.toml
@@ -619,7 +619,7 @@ hash = "c8784129137498425fb70ddd5d2efe24723b522794ea942ab9d0dd463bd65d58"
 package = "systemd-container"
 type = "dbus"
 note = "imported from rpmlint1 DBUSServices.WhiteList"
-bugs = ["bsc#828207", "bsc#1192033", "bsc#1257388", "bsc#1255368"]
+bugs = ["bsc#828207", "bsc#1192033", "bsc#1257388", "bsc#1255368", "bsc#1259318"]
 [[FileDigestGroup.digests]]
 path = "/usr/share/dbus-1/system-services/org.freedesktop.machine1.service"
 digester = "shell"
@@ -633,7 +633,7 @@ hash = "ba1ed5dc253c52e7664fdd5a8e92016ab4d3ebacd444f9f51fcac7ea0a8e3f9b"
 package = "systemd-container"
 type = "dbus"
 note = "imported from rpmlint1 DBUSServices.WhiteList"
-bug = "bsc#964935"
+bugs = ["bsc#964935", "bsc#1259318"]
 [[FileDigestGroup.digests]]
 path = "/usr/share/dbus-1/system-services/org.freedesktop.import1.service"
 digester = "shell"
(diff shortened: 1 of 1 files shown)


configs/openSUSE/dbus-services.toml
```diff
@@ -619,7 +619,7 @@ hash = "c8784129137498425fb70ddd5d2efe24723b522794ea942ab9d0dd463bd65d58"
 package = "systemd-container"
 type = "dbus"
 note = "imported from rpmlint1 DBUSServices.WhiteList"
-bugs = ["bsc#828207", "bsc#1192033", "bsc#1257388", "bsc#1255368"]
+bugs = ["bsc#828207", "bsc#964935", "bsc#1192033", "bsc#1257388", "bsc#1259318"]
```
mgerstner: I'm not completely sure about this merging of two D-Bus services into a single entry. They are now coupled to each other and if one is present the other needs to be present as well for the whitelisting to work.

At the moment this should do, since they're both from the same package. It could be problematic in the future, however.

In the interest of systemd v260 moving forward I'm merging this now anway.

configs/openSUSE/dbus-services.toml
```diff
@@ -619,7 +619,7 @@ hash = "c8784129137498425fb70ddd5d2efe24723b522794ea942ab9d0dd463bd65d58"
 package = "systemd-container"
 type = "dbus"
 note = "imported from rpmlint1 DBUSServices.WhiteList"
-bugs = ["bsc#828207", "bsc#1192033", "bsc#1257388", "bsc#1255368"]
```
mgerstner, line 622: There's a blocker anway, this bsc#1255368 is dropped by the change, I don't think that is correct.

wfrisch: Thanks. Fixed now:
https://github.com/rpm-software-management/rpmlint/compare/opensuse..d9133a059d67c66a4178e8e037506a860941f49c

configs/openSUSE/dbus-services.toml
wfrisch: You're right. I unmerged the services.
https://github.com/rpm-software-management/rpmlint/compare/opensuse..d9133a059d67c66a4178e8e037506a860941f49c

wfrisch:
GitHub's "Compare" diff looks a bit messy due to the rebase.
Better: https://github.com/rpm-software-management/rpmlint/compare/opensuse..d9133a059d67c66a4178e8e037506a860941f49c

configs/openSUSE/dbus-services.toml
```diff
@@ -619,7 +619,7 @@ hash = "c8784129137498425fb70ddd5d2efe24723b522794ea942ab9d0dd463bd65d58"
...
+bugs = ["bsc#828207", "bsc#1192033", "bsc#1257388", "bsc#1255368", "bsc#1259318"]
```
FilippoBonazziSUSE, line 622: I think with the changes you've accidentally dropped the hash update for `org.freedesktop.machine1.conf`. You're mentioning the new bug here but you're not changing the hash value.

wfrisch: Good catch. That was a bit unfortunate because the changes are actually new in v260 (this audit), but also backported to v259.3 for which we whitelisted them already. I thought it makes sense to include the v260 bug nevertheless because this is the version that actually introduced them.

PR #1494, opened 2026-04-09
Fix: Add sysusers.d to usrlibbinaryexception

### LINKED ISSUE #1493: only-non-binary-in-usr-lib reported for systemd-sysusers file
I ran in to a problem resembling: https://github.com/rpm-software-management/rpmlint/issues/1035

I include a textfile (systemd sysusers configuration) as %{_sysusersdir}/%{name}.conf which triggers "W: only-non-binary-in-usr-lib" from rpmlint.

Would it make sense to extend UsrLibBinaryException with "|sysusers\\\.d" too?

qha:
Closes: #1493


danigm (approved):
LGTM

PR #1503, opened 2026-04-13
Fix ci slfo

danigm:
This should fix the CI in the opensuse-slfo-main branch. Cherry-pick of:

* 4855f8c7aaa8f9f94f78d9434510a1d708a434ce
* bfd7440dbfca4a2f9d8e9791ac42ff105ed0d60e


danigm:
Instead of cherry-picking the fixes, it will be better to update the CI to use the corresponding container instead of tumbleweed.

PR #1506, opened 2026-04-16
[opensuse] PermissionsParser: fix parse error when spaces occur between package names

mgerstner:
The `:package:` coupling syntax allows spaces between comma-separated package names. This is currently not considered by the PermissionsParser, leading to whitelisting errors e.g. in the mininews packages, because an empty package name `''` is parsed out of the permissions profile.

Fix this by carefully removing comments and `:package` prefix, the splitting and stripping package names.

### DIFF
--- a/rpmlint/permissions.py
+++ b/rpmlint/permissions.py
@@ -171,7 +171,11 @@ class PermissionsParser:
             for entry in context.active_entries:
                 entry.caps = caps
         elif line.startswith(':package:'):
-            parts = line.split()
-            self._active_packages = parts[1].split(',')
+            # get rid of any trailing comment
+            line = line.split('#')[0]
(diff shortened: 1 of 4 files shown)


FilippoBonazziSUSE (approved):
Looks fine. Do we have tests for this feature?

mgerstner:
> Looks fine. Do we have tests for this feature?

I fixed & added test coverage for this in `permctl` already, but forgot about the rpmlint parsing logic, which led to this bug report.

I can check if I can add some test coverage here as well.

danigm (approved):

mgerstner:
I added commits to add test coverage. Please have a look at that as well.

FilippoBonazziSUSE (approved):

PR #1508, opened 2026-04-21
[opensuse] Refactoring & Reorg of FileDigestCheck

mgerstner:
This is preparation for changes I plan in the FileDigestCheck to support new whitelisting restrictions and features. The current code is hard to follow in some spots and also handles the TOML configuration in a peculiar way not matching the original idea of file digest groups.

These commits introduce refactoring and restructuring of the checker. This also required adjustment of some test cases to match the new behaviour.

### DIFF
--- a/rpmlint/checks/FileDigestCheck.py
+++ b/rpmlint/checks/FileDigestCheck.py
@@ -21,7 +21,7 @@ DEFAULT_DIGEST_ALG = 'sha256'
 # target files depending on their file type. Currently we have two common
 # special cases:
 #
-# - shell like files (configuration files, shell scripts, Python
+# - shell-like files (configuration files, shell scripts, Python
 # scripts, Perl scripts etc) that can contain empty lines or comments
 # introduced by '#'.
 # - XML files for things like D-Bus configuration files or polkit policies.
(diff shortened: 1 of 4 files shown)


rpmlint/checks/FileDigestCheck.py
```diff
@@ -124,101 +124,125 @@ def parse_content(self):
...
+        # {
+        #     "FollowSymlinks": True,
+        #     "Locations: ["/path/one", "/path/two"],
+        #     "NamePatterns: ["*.so"],
+        #     "type": "pam"
+        # }
+        self.checks = []
+        # maps from a check type name like "pam" to its configuration dict as
+        # found in self.checks
+        self.known_check_types: dict[str, dict] = {}
```
FilippoBonazziSUSE, line 142: Why do you need a list and dict of the same `config` objects? Can't you make do with one dict and use its `values()` when you need a list?

rpmlint/checks/FileDigestCheck.py
```diff
@@ -124,101 +124,125 @@ def parse_content(self):
...
+
+        for name, config in self.config.configuration['FileDigestLocation'].items():
```
FilippoBonazziSUSE, line 144: Why are you calling this `name` when it's to all appearances the check type (`pam`, `cron`, `dbus`, ... unless I'm mistaken)

rpmlint/checks/FileDigestCheck.py
```diff
@@ -124,101 +124,125 @@ def parse_content(self):
...
         for exc in self.ghost_file_exceptions:
-            self._verify_package_keys_in_dict('GhostFilesExceptions', exc)
+            self._sanity_check_package_keys('GhostFilesExceptions', exc)
         for exc in self.symlink_exceptions:
-            self._verify_package_keys_in_dict('SymlinkExceptions', exc)
+            self._sanity_check_package_keys('SymlinkExceptions', exc)
 
-    def _get_digester(self, entry):
-        name = entry.get('digester', 'default')
+    def _get_digester(self, digest_info):
```
FilippoBonazziSUSE, line 170: Some type annotations would not hurt here, trying to figure out what this is supposed to be is more work

rpmlint/checks/FileDigestCheck.py
mgerstner: This was a late addition that I ran into. I wanted to use the `config` dicts everywhere, but then there is the problem that a dict cannot be used as a key in a dict, necessitating the use of strings for looking up checks instead. This is where `known_check_types` was added. You are right that the `values()` of a dict could be used for list access as well. But it would make the more common and frequent case of iterating over all checks more verbose again. Also it's not really a big issue, the `config` dictionaries are not deep-copied or anything.

rpmlint/checks/FileDigestCheck.py
mgerstner: Because `type` is a reserved word in Python, which would require us to use `_type` instead, which is ugly. Also it's just a string, so it's kind of the type name. If you have a good idea how to solve this more elegantly, I'm open to suggestions.

rpmlint/checks/FileDigestCheck.py
mgerstner: I can add a commit on top with more annotations. I'm unsure where to draw the line. The types alone don't give a full picture, mostly we're only dealing with `dict` and `str`, this does only convey a bit of meaning, not much semantics.

rpmlint/checks/FileDigestCheck.py
```diff
@@ -373,127 +429,271 @@ def _check_paths_match(self, rpm_path, whitelist_path):
...
+            if group['type'] == check['type'] and self._matches_pkg(group, pkg):
+                ret.append(group)
+        return ret
 
-        Ensures that all files in secured paths have to be whitelisted in an
-        digest whitelisting belonging to group_type.
+    def _check_for_unauthorized(self, pkg, check, restricted_paths):
+        """Detect and complain about files with missing whitelistings.
+
+        Returns the list of paths with are basically whitelisted (but digests
```
FilippoBonazziSUSE:
```suggestion
        Returns the list of paths which are basically whitelisted (but digests
```

rpmlint/checks/FileDigestCheck.py
FilippoBonazziSUSE: ok, it was just confusing at first

mgerstner:
I rebased a bit (the typo, moved some changes that didn't belong there from the additional test case commit into the refactoring commit). I added a commit on top doing some more fine tuning of the code, among it is the harmonization of the `check name` and additional annotation in function parameters.

FilippoBonazziSUSE (approved):

PR #1512, opened 2026-05-11
[opensuse] permissions-whitelist: drop no longer used texlive-filesystem entry

mgerstner:
This is now using systemd-tmpfiles instead.


wfrisch (approved):
LGTM

wfrisch:
https://build.opensuse.org/requests/1352462

PR #1515, opened 2026-05-12
[opensuse] permissions-whitelist: merge postfix and postfix-bdb sections

mgerstner:
The -bdb flavor was overlooked in the previous adjustment. These sections are equivalent, so we can merge them using multi-package assignment syntax.


wfrisch (approved):
LGTM

PR #1518, opened 2026-05-20
checks: Add new RubyCheck

danigm:
This patch adds a new check for ruby files. The goal is to detect wrong shebang lines that points to the interpreter that can use libalternatives, so each script should point to the real interpreter.


Conan-Kudo:
Why aren't you dealing with shebangs with a buildroot policy script at rpmbuild time? This seems to put more burden on the packagers rather than just automating the problem away.

PR #1528, opened 2026-06-09
configs/openSUSE/users-groups.toml: Replace group 'singularity' by 'apptainer'

e4t:


danigm (approved):
LGTM, singularity doesn't exists anymore in Factory and the apptainer is using the apptainer group, so we should update the config: https://build.opensuse.org/package/show/network:cluster/apptainer

PR #1533, opened 2026-06-29
sysctl-whitelist: adjusted 50-coredump.conf digest for systemd - temporary revert (bsc#1267504)

sudomibo:


wfrisch:
LGTM.
As discussed, we will reinstate the change when the systemd maintainers are ready

wfrisch (approved):

PR #1534, opened 2026-07-06
[opensuse] filedigestcheck: use custom socket unit parser to support multiple keys

mgerstner:
Currently we use Python's ConfigParser in non-strict mode to deal with systemd's INI style configuration syntax. systemd also supports multiple values for the same keys and they are often accumulative. ConfigParser only considers the last occurrence of a key and discards all previous occurrences. This means the whitelisting digest calculation is incomplete.

To fix this come up with a simple custom parser for systemd-style INI files. This parser collects values in a list which allows the digester and other users of the API to insect all occurrences in a file.

Also extend the unit test to verify multi-value keys are properly detected by the digester.

This was reported to me by @dirkmueller based on an automatic AI scan of OBS submissions.

### DIFF
--- a/rpmlint/filedigestcheck.py
+++ b/rpmlint/filedigestcheck.py
@@ -1,5 +1,3 @@
-import configparser
-
 # This module contains helper types used by the FileDigestCheck to implement
 # configuration-specific logic.
 
@@ -13,24 +11,67 @@ import configparser
 
 
 def _open_socket_config(path):
-    """Opens a systemd .socket unit in `path` as an INI configuration file and
-    returns the ConfigParser instance resulting from it. On error None is
-    returned."""
-    # systemd services files are not 100 % compatible with regular INI
-    # files, thus we make python's configparser a bit more relaxed.
-    config = configparser.ConfigParser(
-        interpolation=None,
-        strict=False
-    )
+    """Opens a systemd .socket unit in `path` and returns a dictionary
+    containing parsed data. On error None is returned."""
 
-    # by default option keys are converted to lower-case, we don't want that
-    config.optionxform = lambda opt: opt
+    # Sadly we cannot use Python's configparser for this purpose, because the
+    # systemd semantics are too far apart from regular INI files. In
+    # particular we need to be able to process multiple keys of the same name
+    # in a given section, while configparser discards old values when running
+    # in non-strict mode.
 
-    read = config.read(path)
-    if len(read) == 1:
-        return config
+    # see systemd.syntax(7) man page for the details of this syntax
+    #
+    # NOTE: we do not worry about quoting of strings and escaping here, for
+    # the purposes of the whitelisting this should be good enough (we only
+    # lookup simple key/value pairs, calculating a digest over quotes/escapes
+    # should work just as well).
 
-    return None
+    try:
+        ret = {}
+        section = None
+        multiline = ''
+
+        with open(path) as fl:
+            for line in fl:
+                line = line.strip()
+                if not line or line.startswith('#') or line.startswith(';'):
+                    # NOTE: comments are ignored even in multi-line continuation
+                    continue
+                elif line.endswith('\\'):
+                    # backslash is replaced by a space
+                    multiline += line[:-1] + ' '
+                    continue
+
+                line = multiline + line
+                multiline = ''
+
+                if len(line) > 2 and line.startswith('[') and line.endswith(']'):
+                    section = ret.setdefault(line[1:-1], {})
+                    continue
+
+                if section is None:
+                    # this would be a directive before a section appeared,
+                    # which is invalid
+                    return None
+                elif '=' not in line:
+                    # all non-comment, non-section, non-continuation lines
+                    # must contain an equal sign
+                    return None
+
+                # all non-comment, non-section, non-continuation lines must
+                # contain an equal sign
+                key, value = line.split('=', 1)
+                # extra whitespace surrounding the equal sign is ignored
+                key = key.rstrip()
+                value = value.lstrip()
+
+                entries = section.setdefault(key, [])
+                entries.append(value)
+
+        return ret
+    except OSError:
+        return None
 
 
 class VarlinkServiceCheck:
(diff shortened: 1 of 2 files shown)


rpmlint/filedigestcheck.py
```diff
@@ -12,25 +10,62 @@
...
+        section = None
+        multiline = ''
+
+        with open(path) as fl:
+            for line in fl:
+                line = line.strip()
+                if not line or line.startswith('#') or line.startswith(';'):
+                    # NOTE: comments are ignored even in multi-line continuation
+                    continue
+                elif line.endswith('\\'):
```
danigm, line 41: nitpick: We don't need the `elif` just `if`


```suggestion
                if line.endswith('\\'):
```

rpmlint/filedigestcheck.py
```diff
@@ -12,25 +10,62 @@
...
+    # in non-strict mode.
+
+    # see systemd.syntax(7) man page for the details of this syntax
+    #
+    # NOTE: we do not worry about quoting of strings and escaping here, for
+    # the purposes of the whitelisting this should be good enough (we only
+    # lookup simple key/value pairs, calculating a digest over quotes/escapes
+    # should work just as well).
+
+    try:
```
danigm, line 30: I don't like this try-exept wrap around the whole function content. I think it's better to move the try-except code outside, to the caller, so this function doesn't hide all exception but the caller "decides" to ignore.

rpmlint/filedigestcheck.py
```diff
@@ -12,25 +10,62 @@
...
-    if len(read) == 1:
-        return config
+                entries = section.setdefault(key, [])
+                entries.append(value)
 
-    return None
+        return ret
+    except FileNotFoundError:
+        return None
+    except Exception:
```
danigm: As `FileNotFoundError` is a subclass of `Exception` and the handling is the same, it doesn't make sense to catch both, it will be enough to catch the more generic.

In any case, I will try to avoid the plain `Exception` because it will hide any bug in the parsing, not just broken files. At least, if we want to do not crash and catch everything we should add some logging information.

It's usual to have broken files? the only problematic code that I can see is the `key, value = line.split('=', 1)`, that will fail if `=` is not present in the line, but it's something easy to detect with:

```
if not '=' in line:
    # Invalid line, ignoring
    continue
```

I can't see any other `Exception` happening in this code.

danigm (changes requested):

FilippoBonazziSUSE (approved):
Ok from security feature POV, but do see the comments by Dani

rpmlint/filedigestcheck.py
```diff
@@ -12,25 +10,62 @@
...
-    """Opens a systemd .socket unit in `path` as an INI configuration file and
-    returns the ConfigParser instance resulting from it. On error None is
-    returned."""
-    # systemd services files are not 100 % compatible with regular INI
-    # files, thus we make python's configparser a bit more relaxed.
-    config = configparser.ConfigParser(
-        interpolation=None,
-        strict=False
-    )
+def _open_socket_config(path) -> dict[str, list]:
```
wfrisch: `dict[str, list]` doesn't describe the actual nested shape and doesn't include the possible `None` return value.

Suggestion:
```
def _open_socket_config(path) -> dict[str, dict[str, list[str]]] | None:
```

Consider either removing the type annotation or adjust it.

rpmlint/filedigestcheck.py
```diff
@@ -12,25 +10,62 @@
...
+                key, value = line.split('=', 1)
+                # extra whitespace surrounding the equal sign is ignored
+                key = key.rstrip()
+                value = value.lstrip()
 
-    read = config.read(path)
-    if len(read) == 1:
-        return config
+                entries = section.setdefault(key, [])
+                entries.append(value)
```
wfrisch, line 70: This can be optionally collapsed into one line:
```
section.setdefault(key, []).append(value)
```

mgerstner: Feels too compressed to me

rpmlint/filedigestcheck.py
mgerstner: But what's the gain? If we decide not to continue in the `if` branch at a later time then this could bite us so I rather want to make it explicit that only one of the branches is supposed to execute.

rpmlint/filedigestcheck.py
mgerstner: We are seeing all kind of weird files in packages and propagating the exception could mean that the whole rpmlint run breaks. The hiding of exceptions is also a bit due to the way the FileDigestCheck works. But let me see if I can improve a bit on it.

rpmlint/filedigestcheck.py
mgerstner: Good point, I thought I was missing something there, but got over it. The full specification doesn't seem too helpful, so I'll rather drop it.

rpmlint/filedigestcheck.py
mgerstner: Looking at it more closely again I think there is no big improvement by propagating the exceptions outside. This is a private method for this module and we thus know all the callers. `None` return on error is clearly documented. We don't make a difference between parsing errors / file not found, because we consider all errors to be whitelisting violations.

There are additional things that can go wrong in the parsing (e.g.: No `[section]` present but `key = value` lines).

IMHO the real problem here is that we do not have the rpmlint log / diagnostic infrastructure in these classes and thus we are not able to output information about what the actual parsing error might have been.

rpmlint/filedigestcheck.py
danigm: No gain, it's okay to keep the `elif`, it's just that looks weird to me to have the else when in there's a continue inside the `if`, but good for me to keep the `elif` in this case.

rpmlint/filedigestcheck.py
danigm: oh yes, you are right, it could be nice to pass the `self.output` to the checker class, but that's out of the scope of this PR, so I think it's okay, it doesn't worth to block this change to do a refactoring.

I don't like the try-except wrap for the whole function, because it's like an anti-pattern, but in this case, it could be okay, the block inside the try-except is not too big.

In any case, I still think that we can avoid all parsing exceptions with something like:

```python
                # all non-comment, non-section, non-continuation lines must
                # contain an equal sign
                if section is None or '=' not in line:
                    return None
```

And we can just catch all possible exceptions from `open` that should be `OSError` instead of `FileNotFoundError`.

mgerstner: Yes I think that is a good compromise. I also checked during runtime what exactly happens when invalid unit files appear. We will get a diagnostic like this:

```
test.x86_64: E: varlink-file-unauthorized (Badness: 10) /usr/lib/systemd/system/strange.socket (sha256 file digest default filter:3fa3803e230d2ee9a605e516d7ffa7b92313907f71689ab92d0266c845553213 shell filter:463e5041089c447e9d022d10bfcc308ecc971e2f8ff3750ade0bf5d3cb75135e xml filter:<failed-to-calculate> systemd-socket filter:<failed-to-calculate>)
```

So a whitelisting violation will occur (the content is unclear, so this is what we want) and the digest for `systemd-socket filter` will be `failed-to-calculate`.

We do it the same way for the XML digester already. It means that the cause of the error is not user visible, but we rarely run into such situations and a security engineer will then diagnose the situation when packagers approach us about the whitelisting error so it should be managable.

I adjusted the code as you suggested, splitting the error conditions in two, however, for the purpose of clearer documentation.

danigm (approved):
LGTM

PR #1540, opened 2026-07-13
[opensuse] SUIDPermissionsCheck: enforce whitelistings for capabilities (bsc#126…

mgerstner:
…8674)

For a while now RPM natively supports packaging file-based capabilities. Thus the diagnostic permissions-fscaps is no longer valid. Even worse, packagers can bypass our whitelisting policy, because permissions-fscaps does not add badness.

For this reason drop permissions-fscaps and use
permissions-file-setuid-bit in case any capabilities are encountered in a file. Capabilities are considered technically similar to the classic setuid/setgid bits, thus reusing the setuid-bit diagnostic is possible in this context.

Regarding the whitelisting enforcement the same logic as with setuid/setgid bits applies: we're looking up the permissions.secure profile and compare what we see packaged in RPM against what is configured in the permissions profile. If there's a mismatch then we complain.

### DIFF
--- a/rpmlint/descriptions/SUIDPermissionsCheck.toml
+++ b/rpmlint/descriptions/SUIDPermissionsCheck.toml
@@ -16,8 +16,8 @@ please use the %attr macro to set the correct ownership.
 permissions-file-setuid-bit="""Packaging binaries with setuid/setgid bits or capabilities set requires a
 review and whitelisting by the SUSE security team. #REVIEW_NEEDED_TEXT#"""
 permissions-directory-setuid-bit="""#REVIEW_NEEDED_TEXT#"""
-permissions-fscaps="""Packaging files with capabilities is now allowed. The
-permissions package manages capability assigned via the %set_permissions macro
+permissions-fscaps="""Packaging files with capabilities is not allowed. The
+permissions package manages capabilities via the %set_permissions macro
 during %post. Please remove the capabilities from the file."""
 permissions-missing-postin="""
 Please add an appropriate %post section
--- a/rpmlint/descriptions/SUIDPermissionsCheck.toml
+++ b/rpmlint/descriptions/SUIDPermissionsCheck.toml
@@ -8,17 +8,17 @@ permissions-file-as-dir="""
 the entry in the permissions file refers to a directory but the package actually contains a file. Please contact security@suse.de to remove the slash. Please refer to #AUDIT_BUG_URL# for more information.
 """
 permissions-incorrect="""
-please use the %attr macro to set the correct permissions.
+please use the %attr/%caps macro to set the correct permissions.
 """
 permissions-incorrect-owner="""
 please use the %attr macro to set the correct ownership.
 """
-permissions-file-setuid-bit="""Packaging setuid/setgid binaries requires a
+permissions-file-setuid-bit="""Packaging binaries with setuid/setgid bits or capabilities set requires a
 review and whitelisting by the SUSE security team. #REVIEW_NEEDED_TEXT#"""
 permissions-directory-setuid-bit="""#REVIEW_NEEDED_TEXT#"""
-permissions-fscaps="""
-Packaging file capabilities is currently not supported. Please use normal permissions instead. You may contact the security team to request an entry that sets capabilities in /usr/share/permissions/permissions instead.
-"""
+permissions-fscaps="""Packaging files with capabilities is now allowed. The
+permissions package manages capability assigned via the %set_permissions macro
+during %post. Please remove the capabilities from the file."""
 permissions-missing-postin="""
 Please add an appropriate %post section
 """
--- a/configs/openSUSE/scoring-strict.override.toml
+++ b/configs/openSUSE/scoring-strict.override.toml
@@ -21,6 +21,7 @@ permissions-file-ghost = 10000
 permissions-file-unauthorized = 10000
 permissions-file-setuid-bit = 10000
 permissions-file-symlink = 10000
+permissions-fscaps = 10000
 # this one is problematic, because e.g. /etc/cron.* is packaged by multiple
 # packages and only one should be responsible for caring about %postin
 permissions-missing-postin = 100
--- a/configs/openSUSE/scoring-strict.override.toml
+++ b/configs/openSUSE/scoring-strict.override.toml
@@ -22,9 +22,7 @@ permissions-file-unauthorized = 10000
 permissions-file-setuid-bit = 10000
 permissions-file-symlink = 10000
 permissions-fscaps = 10000
-# this one is problematic, because e.g. /etc/cron.* is packaged by multiple
-# packages and only one should be responsible for caring about %postin
-permissions-missing-postin = 100
+permissions-missing-postin = 10000
 permissions-parse-error = 10000
 polkit-file-digest-mismatch = 10000
 polkit-file-ghost = 10000
--- a/configs/openSUSE/scoring.toml
+++ b/configs/openSUSE/scoring.toml
@@ -69,6 +69,7 @@ permissions-file-ghost = 10
 permissions-file-unauthorized = 10
 permissions-file-setuid-bit = 10
 permissions-file-symlink = 10
+permissions-fscaps = 10
 permissions-parse-error = 10
 pam-unauthorized-module = 10
 world-writable-unauthorized-file = 10
--- a/rpmlint/checks/SUIDPermissionsCheck.py
+++ b/rpmlint/checks/SUIDPermissionsCheck.py
@@ -35,30 +35,46 @@ class SUIDPermissionsCheck(AbstractCheck):
         parser = PermissionsParser(self.var_handler, path)
         self.perms.update(parser.entries)
 
-    def _complain_restricted_mode(self, pkg, path, mode):
-        msg = f'{path} is packaged with setuid/setgid bits (0{stat.S_IMODE(mode):o})'
-        diag = 'permissions-directory-setuid-bit' if stat.S_ISDIR(mode) else 'permissions-file-setuid-bit'
-        self.output.add_info('E', pkg, diag, msg)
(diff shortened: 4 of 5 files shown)


wfrisch (changes requested):
I dug a bit into the fs caps format supported by RPM and found something
previously unknown to me.

[RPM filecaps definition](https://rpm.org/docs/4.20.x/manual/tags.html).

> string array, `cap_to_text(3)` textual representation of file capabilities.

From `cap_to_text(3)`:

> A textual representation of capability sets consists of one or more
> whitespace-separated clauses. [...]
> Each clause consists of a list of comma-separated capability names.

For example `cap_chown=i cap_net_raw=ep` is a valid RPM filecaps entry.

Currently we only parse single clauses. There are conceivable scenarios
involving multiple clauses, even though I suspect virtually all real-life cases
will involve just one clause. Nevertheless we should try to aim for 100%
compatibility with the RPM spec.

Added unit tests and clause parsing:
https://github.com/mgerstner/rpmlint/compare/permissions-fscaps...wfrisch:rpmlint:permissions-fscaps-r2?expand=1

wfrisch:
P.S. multiple clauses would trigger an exception otherwise:
```
FAILED test/test_suid_permissions.py::test_permissions_caps_multigroup - ValueError: too many values to unpack (expected 2)
```

Rare examples for such constructs:
* [cyrus-imapd](https://github.com/cyrusimap/cyrus-imapd/blob/140633944d1b6b58e3a707f99ec27c5aea7fc31d/lib/util.c#L700): `cap_net_bind_service=p cap_setuid=ep`
* [old iputils.spec](https://android.googlesource.com/platform/external/iputils/+/1397b19c1447a157c9d4e54fd820a80f6fe4e360%5E1..1397b19c1447a157c9d4e54fd820a80f6fe4e360/): `cap_net_raw=ep cap_net_admin=ep`
* [libcap.go](https://git.kernel.org/pub/scm/libs/libcap/libcap.git/tree/cap/text.go#n170)
* [btop](https://github.com/aristocratos/btop/blob/9527231dfbcbe697f8eb80849e267377194e6ca8/Makefile#L398): `cap_perfmon=+ep cap_dac_read_search=+ep`

mgerstner:
> I dug a bit into the fs caps format supported by RPM and found something previously unknown to me.

Yes, I was aware of potential additional complexities regarding the syntax. I wanted to avoid the explosion in complexity that supporting exotic syntax results in. I'm not even sure we support this in `permissions` currently. It would be preferrable to use libcap directly for this purpose, but this is again difficult in Python. Also libcap is not a gem regarding its API as well.

The comparison of the packaged and whitelisted capabilities is currently more kind of an informational feature, a mismatch does actually _not_ cause badness at the moment (the same is true for a mismatch of the UNIX file mode). I decided to keep it that way under the notion that:

- something is whitelisted already, so it's not directly a violation.
- we do have three different permissions profiles but compare everything against the `.secure` profile in this check, which is kind of arbitrary.
- the `permissions` package is supposed to manage the permissions in the end anyway, so whatever is packaged is not expected to exist on the filesystem except for a few moments in time during installation until `%post` is executed to call `permctl`. This race condition was discussed a while ago, it also exists for setuid binaries, for example, that are not supposed to be world-accessible. We decided to live with this weakness.

When we consider an evil packager, which is not really the threat model we have, I believe, then I currently see the following scenario as the only problematic one:

- the package contains an escalation vector e.g. `cap_sys_admin` on a binary instead of say `cap_net_bind_service` as we expect it to be in the `permissions` profile.
- the package does not invoke `permctl` in its `%post` section, causing the wrong capabilities to persist

We do check for the invocation of `permctl` in `%post` and then trigger `permissions-missing-postin` which currently has the following strict badness override:

```
# this one is problematic, because e.g. /etc/cron.* is packaged by multiple
# packages and only one should be responsible for caring about %postin
permissions-missing-postin = 100
```

This diagnostic is currently only triggered for the `netcfg` package, because there is an entry for `/etc/exports`. It is actually a bug that the %post is missing there.

So we could consider raising the badness here to `1000` to prevent this scenario from happening. If we really want to have the fully blown capability string parsing in this checker, I don't know, however.

mgerstner:
> P.S. multiple clauses would trigger an exception otherwise:

Right, this should not happen for sure. Maybe we can just add some more robustness to avoid this from happening and see how long this will work out for us. Actually packager's don't really need to add capabilities to the RPM, because we are managing them in `permissions`. Contrary to the classic UNIX file mode, which _must_ be present. In case of an unsupported capability assignment in the package then we can still ask the packager simply to drop the caps from the package and purely rely on `permctl` to manage them.

Otherwise having all this additional complexity just to keep two different mechanisms in sync doesn't seem worthwhile to me.

mgerstner:
> For example `cap_chown=i cap_net_raw=ep` is a valid RPM filecaps entry.

I just tried assigning this capability string to a file in a test RPM package and the result when building the package is as follows:

```
error: unpacking of archive failed on file /usr/bin/capbin;6a560447: cpio: cap_set_file failed - No data available
error: rpmlint-integration-test-0.1.0-0.x86_64: install failed
failed to install rpms, aborting build
```

mgerstner:
> Otherwise having all this additional complexity just to keep two different mechanisms in sync doesn't seem worthwhile to me.

Maybe, to make things even cleaner and clearer, simply drop all the capability parsing and comparison code and reject any packaged capability right away. Maybe that would be the KISS approach leading to the least confusion.

wfrisch:
> > I dug a bit into the fs caps format supported by RPM and found something previously unknown to me.
> 
> Yes, I was aware of potential additional complexities regarding the syntax. I wanted to avoid the explosion in complexity that supporting exotic syntax results in. I'm not even sure we support this in `permissions` currently. It would be preferrable to use libcap directly for this purpose, but this is again difficult in Python. Also libcap is not a gem regarding its API as well.

I don't think this leads to an **explosion** in complexity. It's roughly the same amount of code, and not particularly complex.

However I agree, it might not be worth spending too much energy on. Red Hat also doesn't seem to use this syntax anymore, even though they used to, e.g. in `iputils.spec`.

> > For example `cap_chown=i cap_net_raw=ep` is a valid RPM filecaps entry.
> 
> I just tried assigning this capability string to a file in a test RPM package and the result when building the package is as follows:
> 
> ```
> error: unpacking of archive failed on file /usr/bin/capbin;6a560447: cpio: cap_set_file failed - No data available
> error: rpmlint-integration-test-0.1.0-0.x86_64: install failed
> failed to install rpms, aborting build
> ```

True. It took me a while to realize the reason for that. Multiple clauses share one bit field for the capability sets, that is, it's not possible to have one clause with e(ffective) and the other one without. I was able to build an RPM using this syntax:
```
%caps(cap_chown=ei  cap_net_raw=ep) /usr/bin/fscaps-binary
```
https://paste.opensuse.org/pastes/0e68b01a38b2

It actually works:
```
$ getcap /usr/bin/fscaps-binary 
/usr/bin/fscaps-binary cap_chown=ei cap_net_raw+ep
```

> Maybe, to make things even cleaner and clearer, simply drop all the capability parsing and comparison code and reject any packaged capability right away. Maybe that would be the KISS approach leading to the least confusion.

This might be the most pragmatic solution to be honest. +1 from me.

mgerstner:
> I don't think this leads to an **explosion** in complexity. It's roughly the same amount of code, and not particularly complex.

I'm seeing this more in context of the overall complexity of the whitelisting checkers and `permctl` which amount to a large amount of code and complexity by now, and adding more to that pile if it's not strictly necessary gives me a headache.
 
> This might be the most pragmatic solution to be honest. +1 from me.

All right then I'll go down that route. We only have about 5 packages actually including capabilities at the moment in Factory, they need to drop them after we merge this change. I'll update the PR# to reflect this, tomorrow.

mgerstner:
@wfrisch I just pushed a new version of the change. Basically file-based capabilities are always rejected now, but in two different modes:

- if there is no whitelisting for the capability then `permissions-file-setuid-bit` is raised, asking for a review.
- if there is a whitelisting then the old `permissions-fscaps` is repurposed, asking to remove the capabilities. This diagnostic now also raises badness.

On top of this I now also raised badness for `permissions-missing-postin`, since the situation in Factory in this point changed and we can enforce this again.

I would have liked to add a unit test to verify the enforcement of `permissions-fscaps`, we do not have a possibility to fake the permissions profiles, however. `rpmlint` looks them up in `/usr/share/permissions`, which we don't have control over.

I do have a dedicated integration test package in OBS, however, which will cover this aspect to detect regressions.

rpmlint/descriptions/SUIDPermissionsCheck.toml
```diff
@@ -8,17 +8,17 @@ permissions-file-as-dir="""
...
 please use the %attr macro to set the correct ownership.
 """
-permissions-file-setuid-bit="""Packaging setuid/setgid binaries requires a
+permissions-file-setuid-bit="""Packaging binaries with setuid/setgid bits or capabilities set requires a
 review and whitelisting by the SUSE security team. #REVIEW_NEEDED_TEXT#"""
 permissions-directory-setuid-bit="""#REVIEW_NEEDED_TEXT#"""
-permissions-fscaps="""
-Packaging file capabilities is currently not supported. Please use normal permissions instead. You may contact the security team to request an entry that sets capabilities in /usr/share/permissions/permissions instead.
-"""
+permissions-fscaps="""Packaging files with capabilities is now allowed. The
```
wfrisch: * Packaging files with capabilities is **not** allowed.
* The permissions package manages **capabilities** via the %set_permissions macro.

wfrisch (approved):
LGTM except for 2 typos.
I particularly like that this simplified solution solves the race condition problem.

rpmlint/descriptions/SUIDPermissionsCheck.toml
mgerstner: Good catch. I fixed it.

PR #1549, opened 2026-07-21
[opensuse] dbus-services: update keepalived (bsc#1271920)

wfrisch:
This change should be inconsequential with modern D-Bus brokers, but it doesn't cause any harm either.


configs/openSUSE/dbus-services.toml
FilippoBonazziSUSE: Add a bug reference here? I know it's a minor change but still

wfrisch: Thanks! I missed that. Updated.

PR #1550, opened 2026-07-27
ErlangCheck: Handle beams compiled with the deterministic option

bmwiedemann:
Beams compiled with the Erlang compiler's `'deterministic'` option (e.g. via `ERL_COMPILER_OPTIONS=deterministic` for reproducible builds) omit the 'source' and 'options' entries from the `CInf` chunk, as documented in compile(3).
`ErlangCheck` accessed both unconditionally, so the `KeyError` was caught by the blanket except and every such beam was reported as pybeam-failed (an error with badness), aborting package builds that compile deterministically.

Skip the source check when there is no source entry, and use the presence of the `Dbgi` chunk as the debug-info signal when the options entry is absent. The deterministic option does not strip debug info, so the check stays meaningful.

### DIFF
--- a/rpmlint/checks/ErlangCheck.py
+++ b/rpmlint/checks/ErlangCheck.py
@@ -20,14 +20,24 @@ class ErlangCheck(AbstractFilesCheck):
                 self.output.add_info('W', pkg, 'beam-compile-info-missed', filename)
                 return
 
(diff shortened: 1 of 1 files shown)


danigm (approved):
LGTM

PR #1554, opened 2026-07-31
[opensuse] add whitelistings for Varlink services which already received a review

mgerstner:
The Varlink whitelisting restriction is now in place in Factory, but only with low badness. This PR# adds those packages to the new whitelist that already received an AUDIT in the recent past.


sudomibo (approved):
LGTM

PR #1559, opened 2026-08-06
[opensuse] whitelist gnome-initial-setup polkit rules digest (bsc#1273685)

mgerstner:
This also drops an outdated entry and keeps the current entry, because GNOME:Next is not yet in Factory (scheduled in September).

### DIFF
--- a/configs/openSUSE/polkit-rules-whitelist.toml
+++ b/configs/openSUSE/polkit-rules-whitelist.toml
@@ -99,6 +99,16 @@ path = "/usr/share/polkit-1/rules.d/20-gnome-initial-setup.rules"
 hash = "6241e147d107fbc2f786b7149c26abb75185cf9db5d921401e310f78f4e8bced"
 
 [[FileDigestGroup]]
+package  = "gnome-initial-setup"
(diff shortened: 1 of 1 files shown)


FilippoBonazziSUSE (approved):

FilippoBonazziSUSE:
@danigm CI seems proper broken on `ruff` at least. If this was not already on your radar, can you look into this at some point? Thanks

danigm:
> @danigm CI seems proper broken on `ruff` at least. If this was not already on your radar, can you look into this at some point? Thanks

yes, latest ruff changed a lot of things. [I've disabled it in main branch](https://github.com/rpm-software-management/rpmlint/commit/1f8ce51dbdfcd48bbf3e80d1561b4c52bbe1ff68), so this will be fixed once I merge the master branch into opensuse, maybe next week?

I think that it doesn't worth to reformat a lot of code because the ruff policy changes. So disable for now is the easiest change to do. At some point we can explore a way of configuring to match previous policy. In any case, we have flake8 in the CI, so it's not a major problem to just remove ruff from the CI chain.

PR #1560, opened 2026-08-07
[opensuse] FileDigestCheck: varlink: don't process files in sub-directories

mgerstner:
Currently in the systemd rpmlint log we see the diagnostics of this type:

```
systemd.x86_64: E: varlink-file-symlink (Badness: 10) /usr/lib/systemd/system/sockets.target.wants/systemd-ask-password.socket
```

This is not intended, since this is internal dependency management done by systemd. We only care for direct descendants for the restricted locations.

For this reason introduce a new "Recursive" setting for file digest checks. It defaults to `true` in which case everything will be as before. If it is set to `false` then files in sub-directories of the restricted locations will no longer be considered.

Set this new property to `true` for the VarlinkServiceCheck to avoid these bogus diagnostics. Also add a new test cast to assert that this actually works.

### DIFF
--- a/configs/openSUSE/security.toml
+++ b/configs/openSUSE/security.toml
@@ -35,6 +35,8 @@ NamePatterns = [
     "*.socket",
 ]
 ContentCheck = "VarlinkServiceCheck"
+# don't deal with sub-directories like "socket.target.wants"
+Recursive = false
 
 [FileDigestLocation.polkit]
 FollowSymlinks = false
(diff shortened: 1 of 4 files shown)


mgerstner:
I've run the unit tests locally and they all work. Given that the CI pipeline is utterly broken.

FilippoBonazziSUSE (approved):
Change looks fine. Two points in the commit message:

```diff
-Set this new property to `true` for the VarlinkServiceCheck ...
+Set this new property to `false` for the VarlinkServiceCheck ...
```

this should be `false`, right?

```diff
-Also add a new test cast ...
+Also add a new test case ...
```

mgerstner:
Right, I fixed the errors in the commit message. Should be good to go now.

wfrisch (approved):
The property is now correctly described with its new `false` default and a typo has been fixed.
LGTM

wfrisch:
Bypassing CI due to new unrelated Python Ruff errors.

PR #1564, opened 2026-08-12
add fallback due to ldd behavior change in glibc 2.44

spotaws:
glibc 2.44 changed the behavior of 'ldd -u' when a shared library has IFUNC-related issues. Instead of producing the expected 'Unused direct dependencies:' output, it now emits a 'Relink' message and exits non-zero without any useful dependency information.

Thus, I added a fallback mechanism: when 'ldd -u' fails without producing the expected output format, run 'ldd -r' and detect NEEDED libraries that are marked as 'not found'. These libraries cannot contribute any resolved symbols and are effectively unused from the linker's perspective, matching the previous behavior of 'ldd -u' for unresolvable dependencies.

I'm not entirely sure if this is the right fix, or if this is a glibc regression, but I at least understand this (and I don't really
understand glibc internals).

### DIFF
--- a/rpmlint/lddparser.py
+++ b/rpmlint/lddparser.py
@@ -37,6 +37,7 @@ class LddParser:
 
     unused_regex = re.compile(r'^\s+(?P<lib>\S+)')
     undef_regex = re.compile(r'^undefined symbol:\s+(?P<symbol>[^, ]+)')
+    notfound_regex = re.compile(r'^\s+(?P<lib>\S+)\s+=>\s+not found')
 
     def __init__(self, pkgfile_path, path, is_installed_pkg):
         self.pkgfile_path = pkgfile_path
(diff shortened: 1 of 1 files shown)


danigm:
I think that this is a problem in glibc, we have a [related bug in opensuse](https://bugzilla.opensuse.org/show_bug.cgi?id=1273686), and looks like glibc upstream fixes the issue that will be included in 2.45 and backported to 2.44 release: https://sourceware.org/bugzilla/show_bug.cgi?id=34428

danigm:
The commit 2f2e9bae4093e1c3ba61250e340d3c3b99788f68 in glibc fixes the issue and this check continue working with existing lld, so I think that this patch is not needed.

spotaws:
Late reply, but yes, I agree, that is the correct fix.

PR #1565, opened 2026-08-14
[opensuse] whitelist existing Varlink socket units

mgerstner:
Now that we introduced a new Varlink socket unit whitelisting restriction it is time to populate the whitelist with the currently existing units as found in Factory. I reviewed them over the course of the last couple of days and all looks well.

This is the last step before turning up the badness on the new `varlink-file-unauthorized` diagnostic.

### DIFF
--- a/configs/openSUSE/varlink-whitelist.toml
+++ b/configs/openSUSE/varlink-whitelist.toml
@@ -127,3 +127,13 @@ type     = "varlink"
 path     = "/usr/lib/systemd/system/systemd-resolved-varlink.socket"
 digester = "systemd-socket"
 hash     = "924a769f997001348c10369a0af748932058c8f7e88ad5daf1a7e84abda53f77"
+
+[[FileDigestGroup]]
+package  = "systemd"
+note     = "service which helps to obtain a password from an interactive user"
+bugs     = ["bsc#1261919", "bsc#1257943"]
+type     = "varlink"
+[[FileDigestGroup.digests]]
+path     = "/usr/lib/systemd/system/systemd-ask-password.socket"
+digester = "systemd-socket"
+hash     = "65796969f76971117b1f8405906e7ed54d95230faa45465714ab71ccf158e5aa"
--- a/configs/openSUSE/varlink-whitelist.toml
+++ b/configs/openSUSE/varlink-whitelist.toml
@@ -177,3 +177,13 @@ type     = "varlink"
 path     = "/usr/lib/systemd/system/systemd-sysext.socket"
 digester = "systemd-socket"
 hash     = "695f5a22b67ede89005bbf39065e94d1c17fff449cd925ecc969da654bba5e26"
+
+[[FileDigestGroup]]
+package  = "systemd"
+note     = "access to /etc/passwd, /etc/group data"
+bug      = "bsc#1261919"
+type     = "varlink"
+[[FileDigestGroup.digests]]
+path     = "/usr/lib/systemd/system/systemd-userdbd.socket"
+digester = "systemd-socket"
+hash     = "f18c943e7ee65fea7d34d282111230867b1d309d5bcbf5e0bb00982567c6bb82"
--- a/configs/openSUSE/varlink-whitelist.toml
+++ b/configs/openSUSE/varlink-whitelist.toml
@@ -167,3 +167,13 @@ type     = "varlink"
 path     = "/usr/lib/systemd/system/systemd-logind-varlink.socket"
 digester = "systemd-socket"
 hash     = "7e62a81cb560d9dbb4bf27a8c485ec4a2b2ce8d85a9cfddf3c1061795732ae7d"
+
+[[FileDigestGroup]]
+package  = "systemd"
+note     = "management of system extension / configuration images"
+bugs     = ["bsc#1261919", "bsc#1259318"]
+type     = "varlink"
+[[FileDigestGroup.digests]]
+path     = "/usr/lib/systemd/system/systemd-sysext.socket"
+digester = "systemd-socket"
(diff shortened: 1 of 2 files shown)


configs/openSUSE/varlink-whitelist.toml
```diff
@@ -45,3 +45,239 @@ type     = "varlink"
...
+bug      = "bsc#1261919"
+type     = "varlink"
+[[FileDigestGroup.digests]]
+path     = "/usr/lib/systemd/system/systemd-resolved-varlink.socket"
+digester = "systemd-socket"
+hash     = "924a769f997001348c10369a0af748932058c8f7e88ad5daf1a7e84abda53f77"
+[[FileDigestGroup.digests]]
+path     = "/usr/lib/systemd/system/systemd-resolved-varlink.socket"
+digester = "systemd-socket"
+hash     = "924a769f997001348c10369a0af748932058c8f7e88ad5daf1a7e84abda53f77"
```
wfrisch: This is appears to be an accidental duplicate of the file above.
I suppose it's intended to be `systemd-resolved-monitor.socket`
(with a different hash of course).

configs/openSUSE/varlink-whitelist.toml
```diff
@@ -45,3 +45,239 @@ type     = "varlink"
...
+bugs     = ["bsc#964935", "bsc#1259318", "bsc#1261919"]
+type     = "varlink"
+[[FileDigestGroup.digests]]
+path     = "/usr/lib/systemd/system/systemd-importd.socket"
+digester = "systemd-socket"
+hash     = "fa1c9b80e6ba5b08f09abebcc7eabb00f93ef5c86767e2989785f7c6264e24b5"
+
+[[FileDigestGroup]]
+package  = "systemd-container"
+note     = "main daemon for controllings VMs/containers, analogoues to its D-Bus twin"
```
wfrisch: controlling

analogous

wfrisch (changes requested):
`systemd-networkd-varlink-metrics.socket` appears to be missing.

According to [comment 7](https://bugzilla.suse.com/show_bug.cgi?id=1261919#c7) in the audit bug, it's intended to be whitelisted.

wfrisch:
varlink should be added to the whitelist syntax test too:
```
diff --git a/test/test_whitelist_syntax.py b/test/test_whitelist_syntax.py
index 92fdefc1..14aa5966 100644
--- a/test/test_whitelist_syntax.py
+++ b/test/test_whitelist_syntax.py
@@ -50,7 +50,8 @@ def test_file_digest_whitelists():
         ('dbus', 'dbus-services'),
         ('pam', 'pam-modules'),
         ('permissions', 'permissions-whitelist'),
-        ('polkit', 'polkit-rules-whitelist')
+        ('polkit', 'polkit-rules-whitelist'),
+        ('varlink', 'varlink-whitelist')
     ]
 
     for _type, whitelist in WHITELISTS:
```

configs/openSUSE/varlink-whitelist.toml
mgerstner: It's even weirder:

- for some reason there are two consecutive build logs for some of the systemd components, which is why a duplicate badness diagnostic appears for systemd-resolved-varlink.socket. The `buildlog2whitelist` script picked both up.
- I actually expected two entries, one for `systemd-resolved-monitor.socket` as you suggested. That socket is not covered by the current check, however, because it contains: `FileDescriptorName=varlink-monitor`. I'm not completely sure why they use a different name here.

I simply dropped the duplicate now.

configs/openSUSE/varlink-whitelist.toml
mgerstner: fixed

mgerstner:
All findings should be addressed now.

configs/openSUSE/varlink-whitelist.toml
wfrisch: Just to be sure, did you intentionally drop a large number of entries?

```
-[[FileDigestGroup.digests]]
-path     = "/usr/lib/systemd/system/systemd-resolved-varlink.socket"
-digester = "systemd-socket"
-hash     = "924a769f997001348c10369a0af748932058c8f7e88ad5daf1a7e84abda53f77"
-
-[[FileDigestGroup]]
-package  = "systemd"
-note     = "service which helps to obtain a password from an interactive user"
-bugs     = ["bsc#1261919", "bsc#1257943"]
-type     = "varlink"
-[[FileDigestGroup.digests]]
-path     = "/usr/lib/systemd/system/systemd-ask-password.socket"
-digester = "systemd-socket"
-hash     = "65796969f76971117b1f8405906e7ed54d95230faa45465714ab71ccf158e5aa"
```
[...]

https://github.com/rpm-software-management/rpmlint/compare/d8899e5e455a2b42f9d44d07f291fc15212e089f..555a12e101fe47b640fbc8ce6363ec978a0f9858

mgerstner: Sorry I seem to have messed up the Git rebase completely. I restored the state and applied the fixes again.

mgerstner:
> `systemd-networkd-varlink-metrics.socket` appears to be missing.
> 
> According to [comment 7](https://bugzilla.suse.com/show_bug.cgi?id=1261919#c7) in the audit bug, it's intended to be whitelisted.

Good point. This socket unit also has a non-default file descriptor name:

```
FileDescriptorName=varlink-metrics
```

I'm not completely sure why they deviate in these few cases. We could consider adjusting the Varlink check to trigger if the `FileDescriptorName` contains `varlink` as a sub-string. I'd leave that for a follow-up change, however.

wfrisch:
LGTM now but the commit adding varlink to `test/test_whitelist_syntax.py` has vanished
Not that important but it would be nice to have

mgerstner:
> vanis

I cherry-picked and pushed it.

PR #1569, opened 2026-08-18
[opensuse] pam-modules: whitelist pam_ssh_agent (bsc#1274633)

mgerstner:


configs/openSUSE/pam-modules.toml
```diff
@@ -577,3 +577,12 @@ nodigests = [
...
     "glob:*/security/pam_unix_ng.so"
 ]
+
+[[FileDigestGroup]]
+package  = "pam-ssh-agent"
+note     = "authentication via SSH public keys"
+bug      = "bsc#1274633"
+type     = "pam"
+nodigests = [
+    "global:*/security/pam_ssh_agent.so"
```
wfrisch: This should be `glob:*/security/pam_ssh_agent.so`

wfrisch (changes requested):
global/glob typo. LGTM otherwise.

configs/openSUSE/pam-modules.toml
mgerstner: oh my, only now I realize that the typo is in the `buildlog2whitelist` script, this bug happened a couple of times before, and I never found the culprit

PR #1575, opened 2026-08-31
configs: Support eln release extension

jelly:
Now that rpmlint runs on eln builds they will always fail with `incoherent-version-in-changelog` as `eln` is not understood as release extension.


sgallagher (approved):
LGTM 👍

danigm (approved):

